PR #1, opened 2026-05-18
Migrate AFD into external vLLM plugin

jiangkuaixue123:
## Summary

This PR migrates the original in-tree AFD implementation into an out-of-tree vLLM external plugin targeting vLLM `v0.19.1`, without modifying the target vLLM source tree.

The plugin now provides the packaging, registration, runtime class paths, connector layer, model integration, compatibility shims, docs, CI, and tests needed to run Attention-FFN Disaggregation through native vLLM entrypoints plus explicit plugin-owned classes.

## What Changed

- Added the `vllm-afd-plugin` package skeleton with a `vllm.general_plugins` entry point, idempotent `register_afd()`, optional vLLM runtime dependency handling, and CPU-safe imports.
- Added plugin-owned `AFDConfig` parsing from `additional_config["afd"]`, validation helpers, and class-path smoke coverage for Attention/FFN worker and runner classes.
- Implemented Attention-side runtime adapters around vLLM v1 `GPUWorker` / `GPUModelRunner`, including AFD metadata construction and `ForwardContext.additional_kwargs["afd_metadata"]` integration.
- Implemented FFN-side runtime adapters, including `AFDFFNWorker`, `GPUFFNModelRunner`, empty KV-cache handling, connector-driven FFN daemon execution, and fail-fast handling for unexpected scheduler execution.
- Migrated the P2P connector path with topology helpers, distributed process-group support, DP/AFD metadata exchange, hidden-state send/recv, and FFN-to-Attention output return.
- Added DBO / ubatching support for AFD metadata propagation, per-stage DP metadata, empty-slice handling, and two-way ubatch replay behavior.
- Added `FULL_DECODE_ONLY` CUDA graph support for AFD, including Attention warmup/capture/replay metadata sends, FFN graph-keyed capture/replay, and guarded policy validation for unsupported graph modes.
- Added plugin-owned DeepSeekV2 integration and forward-context helpers so model code can consume AFD metadata without patching `vllm.forward_context`.
- Added narrowly scoped compatibility patches for vLLM `v0.19.1` config validation and EngineCore FFN daemon lifecycle support.
- Added migration/design documentation plus CPU-safe and GPU-gated E2E test coverage.
- Added CPU-only GitHub Actions CI for ruff linting, ruff format checks, issue-template YAML parsing, and unit tests that do not require GPU or an importable vLLM runtime.

## Current Support

Supported today:

- vLLM target: `v0.19.1`.
- Startup model: native `vllm serve` with plugin registration and explicit worker/model class paths.
- Connector: `p2pconnector`.
- Topologies covered by tests/docs: `1A1F` and `2A2F` with native vLLM DP layout.
- CUDA graph mode: `FULL_DECODE_ONLY` only.
- DBO + CUDA graph: supported for `num_ubatches=2`.
- Model coverage: DeepSeekV2 plugin-owned wrapper for AFD E2E validation.

Known limitations / deferred hardening:

- Role-based weight pruning is not implemented yet; Attention and FFN sides still load full DeepSeekV2 weights.
- Unsupported CUDA graph modes fail fast (`PIECEWISE`, `FULL`, `FULL_AND_PIECEWISE`, and other non-eager configurations).
- DBO graph with `num_ubatches != 2` fails fast.
- TP/PP, ratio > 1 topologies, EP collective behavior, non-DeepSeekV2 models, graph output equivalence checks, strict graph-miss handling, and graph memory planning still need follow-up validation/hardening.

## Validation

Added CPU-safe tests for:

- package import and plugin registration
- AFD config parsing and validation
- runtime class-path resolution
- config-validation and EngineCore compatibility patches
- Attention/FFN runtime behavior
- connector metadata and P2P connector helpers
- CUDA graph policy, graph keys, warmup/capture/replay flags, and ubatch metadata cloning
- model forward-context metadata access

Added CPU-only CI coverage for:

- `uv sync --locked --group dev`
- `uv run ruff check .`
- `uv run ruff format --check .`
- GitHub issue-template YAML parsing
- `uv run pytest -q tests/unit -m "not gpu and not vllm_runtime"`

Added opt-in GPU E2E coverage for:

- DeepSeekV2 eager `1A1F`
- DeepSeekV2 eager `2A2F`
- DeepSeekV2 `FULL_DECODE_ONLY` CUDA graph `1A1F`
- DeepSeekV2 `FULL_DECODE_ONLY` CUDA graph `2A2F`
- DeepSeekV2 `FULL_DECODE_ONLY` CUDA graph `2A2F` with two-way DBO ubatch replay

Manual L20X validation documented in `docs/PHASE6_CUDA_GRAPH_IMPLEMENTATION.md` includes a successful `2A2F + FULL_DECODE_ONLY + DecodeBenchConnector + DBO` run with capture size `64`, concurrency `128`, and `128/128` completion requests returning `200`.


hsliuustc0106:
COMPILE DONE?

jiangkuaixue123:
Non-DBO compilation done; DBO-based compilation remains to be fixed.

afd_plugin/connectors/p2p.py
```diff
@@ -0,0 +1,570 @@
...
+            device=device,
+        )
+
+        for dst in self.dst_list:
+            torch.distributed.send(size_tensor, dst=dst, group=self.p2p_pg)
+            torch.distributed.send(object_tensor, dst=dst, group=self.p2p_pg)
+
+    def recv_dp_metadata_list(
+        self,
+        timeout_ms: int | None = None,
```
hsliuustc0106, line 229: **P2: `timeout_ms` silently discarded — FFN daemon can hang indefinitely.**

The `recv_dp_metadata_list(timeout_ms=100)` call in `ffn_worker.py:141` looks like it has a timeout, but the first line here does `del timeout_ms`. The `torch.distributed.recv` on line 243 is a blocking call with no timeout mechanism. If the Attention side process crashes or never sends, the FFN daemon thread blocks forever — the `while not event.is_set()` check never gets a chance to re-evaluate.

**Fix:** Either implement a real timeout (e.g., poll-based recv or watchdog thread), or remove the misleading `timeout_ms` parameter from the API since it's not implemented.

afd_plugin/compat/patches/config_validation.py
```diff
@@ -0,0 +1,155 @@
...
+        import vllm
+
+        version_value = getattr(vllm, "__version__", None)
+    except Exception:
+        version_value = None
+    if version_value is None:
+        return True
+    version_text = str(version_value)
+    if "dev" in version_text:
+        return True
```
hsliuustc0106, line 155: **P2: Dev version heuristic is too permissive.**

`return True` for any version containing "dev" means a `0.20.0.dev0` install passes validation against `0.19.1` APIs. When the vLLM API changes in a dev build, the monkey-patched methods will fail with inscrutable errors deep in vLLM internals rather than a clear version-mismatch message.

**Fix:** Parse the dev version prefix and compare major.minor against `TARGET_VLLM_VERSION`:
```python
if "dev" in version_text:
    release_part = version_text.split("dev")[0].rstrip(".")
    if release_part:
        return release_part.startswith(TARGET_VLLM_VERSION)
    return True
```

afd_plugin/v1/worker/ffn_worker.py
```diff
@@ -0,0 +1,185 @@
...
+        if not connector.is_initialized:
+            self.model_runner.initialize_afd_connector()
+
+        self._ffn_shutdown_event = threading.Event()
+        self._ffn_loop_error = None
+
+        def ffn_worker_loop() -> None:
+            try:
+                self._run_ffn_server_loop()
+            except BaseException as exc:
```
hsliuustc0106: **P2: `BaseException` catch in daemon thread can suppress graceful shutdown.**

Catching `BaseException` includes `KeyboardInterrupt` and `SystemExit`. The exception is stored in `self._ffn_loop_error` and re-raised, but since this runs in a `daemon=True` thread, the re-raise may not reach the main thread. Combined with `daemon=True`, the thread is silently killed on process exit without NCCL cleanup (`connector.close()`). The `shutdown()` tries `thread.join(timeout=5)`, but if the thread is blocked in `torch.distributed.recv`, the join times out and cleanup never runs.

**Fix:** Catch `Exception` instead of `BaseException` (let `KeyboardInterrupt`/`SystemExit` propagate naturally). Consider a non-daemon thread or a signal handler that calls `stop_ffn_server_loop()` on SIGTERM/SIGINT.

afd_plugin/compat/patches/engine_core.py
```diff
@@ -0,0 +1,347 @@
...
+        dp_engine_core_proc_cls.run_busy_loop = (
+            patched_dp_engine_core_proc_run_busy_loop
+        )
+
+    setattr(core_module, _PATCH_ATTR, state)
+    logger.debug("AFD EngineCore patch applied")
+
+
+class _AFDFFNKVCacheConfig:
+    kv_cache_groups: list[Any] = []
```
hsliuustc0106, line 166: **P3: `kv_cache_groups` is a class attribute, shared across all instances.**

`kv_cache_groups: list[Any] = []` is declared at class scope — all instances of `_AFDFFNKVCacheConfig` share the same list. Currently only constructed once, so it works, but if this is ever instantiated more than once (process restart, fork, or future reuse), mutations to one instance silently affect others.

**Fix:** Move to an instance attribute:
```python
class _AFDFFNKVCacheConfig:
    def __init__(self):
        self.kv_cache_groups: list[Any] = []
```

hsliuustc0106 (commented):
Thanks for the PR. A few findings from review:

**1 CRITICAL, 3 INFORMATIONAL**

The most important one is the silently discarded `timeout_ms` in `recv_dp_metadata_list` — the FFN daemon loop can hang forever if the Attention side crashes or a network partition occurs.

Overall the code quality is solid: clean dataclasses, CPU-safe imports, narrow compat patches, and good test coverage.

pyproject.toml
```diff
@@ -0,0 +1,93 @@
...
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Topic :: Scientific/Engineering :: Artificial Intelligence",
+]
+dependencies = []
+
+# Keep vLLM optional so macOS and CPU-only development can run import/config tests.
+# Install the runtime extra only in environments with a compatible vLLM wheel.
+[project.optional-dependencies]
+vllm = ["vllm>=0.19.1"]
```
hsliuustc0106: fix version

afd_plugin/runtime/__init__.py
```diff
@@ -0,0 +1,17 @@
+# SPDX-License-Identifier: Apache-2.0
```
hsliuustc0106: runtime -> worker

afd_plugin/model_executor/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: Apache-2.0
```
hsliuustc0106, line 1: model_executor

afd_plugin/distributed/process_group.py
```diff
@@ -0,0 +1,99 @@
+# SPDX-License-Identifier: Apache-2.0
```
hsliuustc0106: combine

afd_plugin/config.py
```diff
@@ -0,0 +1,282 @@
...
+    if "extra_config" in normalized and not isinstance(
+        normalized["extra_config"],
+        dict,
+    ):
+        raise TypeError("extra_config must be a dict")
+
+    return normalized
+
+
+def AFDConfig_from_mapping(
```
hsliuustc0106: afd_config_xxx

tests/unit/config/test_config.py
```diff
@@ -0,0 +1,106 @@
+from __future__ import annotations
```
hsliuustc0106, line 1: ut/e2e

ut/
modelxxx
coneccxx

hsliuustc0106: e2e/ds-v2

hsliuustc0106 (approved):

Issue #36, opened 2026-06-29
[RFC]: Support HunyuanImage3 with vLLM-Omni recipe on GPU


specture724:
### Motivation

Current AFD-Plugin only runs on vLLM with DeepSeekV3.2 tested. New models need to be adapted. The goal is to adapt HunyuanImage3 model into current framwork, and make it work with vLLM-Omni. 
HunyuanImage3 runs in vLLM-Omni as two stage-level workers — autoregressive (AR) and diffusion-transformer (DiT) — each containing Attention + MoE blocks. Existing stage-level serving does not separate Attention from MoE inside a stage. Whether intra-stage AFD is worthwhile is still open, and it likely differs between the two stages:
- AR is autoregressive decode with a growing KV cache, closer to the existing MoE-LLM AFD regime. It is expected to be the lower-risk track, but still requires profiling and integration validation. 
- DiT is a 50-step denoising loop over fixed-length latent/image tokens with no KV-cache growth. Our current DiT profiling suggests a different regime: attention is full self-attention and is not dominated by KV-cache growth, so the classic “offload memory-bound attention” rationale does not directly apply.

### Proposed change

- AR track
 - [ ] Model Adaption: adapt HunyuanImage3 model to AFD Plugin, with pruned weight loading, TP/EP/DP parallelism, Integration with vLLM-Omni, DBO, Cuda graph support.
 - [ ] Benchmark Building: Build real-world workload patter benchmark for latency, throughput, GPU utilization, memory, and exposed communication after overlap as the go/no-go metric
 - [ ] Performance Profiling: Attention vs MoE time, KV behavior, find AF payload sweet spot.
- DiT track
 - [ ] TODO

### Plugin boundary

Plugin-owned: 
- HunyuanImage3 model wrapper.
- (TODO: DiT classes to be injected to vLLM-Omni)

### Risks and alternatives

AFD Plugin currently works with vLLM v0.19.0. Upstream vLLM-Omni now supports vLLM 0.23.0. We are trying to update vLLM version for compatibility.

### CC list

@jiangkuaixue123 @hsliuustc0106 @yujuancao07 @jiaran-king

hsliuustc0106:
I think we can go with AR stage first

jiangkuaixue123:
Closing this for now. We can reopen it later if needed.

PR #56, opened 2026-07-03
[Feat]: HunyuanImage3 AR stage recipe with vLLM-Omni

specture724:
## Purpose

The basic implimentation of HunyuanImage3 AR stage with vLLM-Omni compatible plugin. 
## Issue

- Related issue(s): #36 

## Test Plan

Recipe like DeepSeek V2 Lite is implimented in `recipe/gpu/hunyuan_image3`. Only prefill-decode colocation tests in both eager mode and cuda graph is done because I don't have so much GPUs to run TP2 1P1A1F test.

---


recipe/gpu/hunyuan_image3/prefill_decode_colocation/run.sh
```diff
@@ -0,0 +1,97 @@
...
+# than maintain four near-identical YAMLs, both configs are rendered from the
+# single template below, parameterised by run mode (eager/graph) and role.
+#
+# Usage: ./run.sh [eager|graph]   (default: graph)
+#   eager - enforce_eager, no cudagraph capture (debugging)
+#   graph - FULL_DECODE_ONLY cudagraph capture (performance)
+set -euo pipefail
+
+MODE="${1:-graph}"
+MODEL_PATH="${MODEL_PATH:-/path/model_weights/DeepSeek-V2-Lite}"
```
hsliuustc0106, line 15: This default points to DeepSeek-V2-Lite, so running the Hunyuan recipe without MODEL_PATH launches the wrong checkpoint. Please default to HunyuanImage-3 or fail with a clear message.

afd_plugin/model_executor/models/hunyuan_v1.py
```diff
@@ -0,0 +1,245 @@
...
+            ref_tensor=hidden_states,
+            ubatch_idx=stage_idx,
+        )
+        return hidden_states, residual
+
+    def compute_ffn_output(
+        self,
+        hidden_states: torch.Tensor,
+        layer_idx: int,
+        **kwargs: Any,
```
hsliuustc0106, line 201: Please avoid the `**kwargs: Any` pass-through on this core model/FFN boundary. Spell out the optional inputs this runner supports, or remove them, so the AFD Hunyuan contract is clear and typed.

afd_plugin/compat/patches/omni_server.py
```diff
@@ -0,0 +1,182 @@
...
+    except ImportError:
+        logger.debug("AFD omni-server patch skipped: vllm-omni is unavailable")
+        return
+
+    if hasattr(_server_mod, _PATCH_ATTR):
+        return
+
+    original_run_server_worker = _server_mod.omni_run_server_worker
+
+    async def patched_omni_run_server_worker(
```
hsliuustc0106, line 53: Please replace the `Any`/`**kwargs` patch boundary with concrete vllm-omni server argument types or a small protocol. This entry point decides FFN daemon behavior, so its supported inputs should be explicit.

hsliuustc0106 (commented):
Maintainability follow-up from the project review rules.

Issue #88, opened 2026-07-09
Refactor: Connector metadata interfaces and ownership


jiangkuaixue123:
Parent RFC: #86

## Scope

Refactor connector metadata naming, ownership, and lifecycle hooks as one coherent interface cleanup.

## Goals

- Clarify the lifecycle currently represented by `create_recv_metadata()`, `configure_metadata()`, and `update_metadata()`.
- Introduce clearer lifecycle vocabulary, for example:
  - `build_recv_transfer_metadata()`
  - `attach_transfer_state()`
  - `complete_recv_transfer_metadata()`
- Clarify the ownership boundary between:
  - `AFDMetadata`
  - `AFDConnectorMetadata`
  - `AFDConnectorData`
- Consider conceptual aliases such as:
  - `AFDForwardContextMetadata`
  - `AFDConnectorTransferMetadata`
  - `AFDConnectorTransferState`
- Keep migration compatibility while updating internal docs and call sites.

## Non-goals

- Do not change backend communication behavior in this refactor.
- Do not remove old class or method names until call sites and tests have migrated.

## Acceptance Criteria

- Connector metadata lifecycle phases are documented.
- The preferred naming model is agreed.
- Backward-compatible aliases or wrappers are proposed or implemented.
- Tests cover wrapper compatibility and key metadata invariants.

specture724:
```mermaid
classDiagram
    class AFDForwardContextMetadata { 
        +tokens_start_loc: list[int]
        +request_start_locs: list[int]
        +tokens_lens: list[int]
        +tokens_unpadded_lens: list[int]
        +num_stages: int
        +stage_idx: int
        +connector: AFDConnectorBase
        +transaction_id: str | None
        +clone()
    }
    
    class AFDConnectorBase {
    	+control_plane: AFDControlPlane | None
    	+ffn_step_trigger: enum
	+is_initialized: bool
	+init_afd_connector()
	+send_ffn_output()
	+send_attn_output()
	+recv_ffn_output()
	+recv_attn_output()
		
    }
    class AFDControlPlane {
    	+ingest_control_payload()
    	+publish_control_payload()
        +poll_control_payload()
    }
    class AFDDPMetadata { 
        +num_tokens_across_dp_cpu: Tensor
        +max_tokens_across_dp_cpu: Tensor | None = None
        +local_sizes: list[int] | None = None
        +sp_local_sizes()
        +get_chunk_sizes_across_dp_rank()
[...]

PR #128, opened 2026-07-15
Refactor/split control plane

### LINKED ISSUE #107: Refactor: Decoupling control plain methods from `AFDConnectorBase`
Parent RFC: #88 

### Goals
-------
Change the control plane from a mandatory abstraction to an optional capability.

```python
class AFDControlPlane(Protocol):
    """Only connectors that require DP token count negotiation implement this.
    GPU P2P / CAMP2P implement it;
    AFDAsyncConnector does NOT implement it (it relies on CAM op's built-in counting)."""
    def update_state_from_dp_metadata(self, payload: AFDControlPayload) -> None: ...  
    def send_dp_metadata_list(self, payload: AFDControlPayload) -> None: ... 
    def recv_dp_metadata_list(self, timeout_ms: int | None = None) -> AFDControlPayload: ... 
```

`AFDConnectorBase` no longer marks these three methods as `@abstractmethod`; instead:

```python
class AFDConnectorBase(ABC):
    control_plane: AFDControlPlane | None
    ffn_step_trigger: Trigger = Trigger.DP_METADATA   # Trigger.DP_METADATA | Trigger.CONNECTOR
```

On the runner side, use `if connector.control_plane:` to decide whether to go through the control-plane handshake, rather than unconditionally calling an abstract method that is a no-op in an async context. This eliminates the interface lie of "abstract methods being implemented as empty stubs."

specture724:
## Purpose

Make DP metadata control plain a pluggable component for connector. 

## Issue

- Related issue(s): #107 
- Closing keyword, only if fully resolved: Closes #107 

## Scope

- In scope: control plain
- Out of scope: other

## Test Result

All smoke tests (GPU, NPU, NPU async) passed


afd_plugin/connectors/base.py
```diff
@@ -20,6 +21,11 @@
     from vllm.config import VllmConfig
 
 
+class Trigger(enum.Enum):
```
jiangkuaixue123: Add doc string.  Explain the meaning of DP_METADATA and CONNECTOR

afd_plugin/v1/worker/attention_model_runner.py
```diff
@@ -132,8 +132,8 @@ def _send_dp_metadata(
             is_graph_capturing=is_graph_capturing,
             is_warmup=is_warmup,
         )
-        self.afd_connector.update_state_from_dp_metadata(payload)
-        self.afd_connector.send_dp_metadata_list(payload)
+        self.afd_connector.control_plane.update_state_from_dp_metadata(payload)
+        self.afd_connector.control_plane.send_dp_metadata_list(payload)
```
jiangkuaixue123, line 143: Image缩进看上去有点奇怪

afd_plugin/connectors/base.py
```diff
@@ -210,9 +225,9 @@ def send_ffn_output(
         """
         raise NotImplementedError
 
-    # ==============================
-    # DP metadata control plane
-    # ==============================
+
+class AFDControlPlane(ABC):
```
hsliuustc0106, line 243: These methods were previously part of the public `AFDConnectorBase` contract. Removing them makes existing callers fail with `AttributeError`, while the parent RFC (#88) requires migration compatibility. Please keep non-abstract forwarding methods on the connector, with defined behavior when `control_plane` is `None`, and cover the compatibility path with tests.

afd_plugin/connectors/base.py
```diff
@@ -20,6 +21,20 @@
     from vllm.config import VllmConfig
 
 
+class Trigger(enum.Enum):
```
hsliuustc0106: `Trigger.CONNECTOR == "connector"` is false, and the worker compares with `is`. A connector registered through the public factory that still uses the previous string value falls into the DP-metadata branch and dereferences a `None` control plane. Please preserve the legacy values with a Python-3.10-compatible string enum plus equality or coercion, and add a compatibility test.

afd_plugin/v1/worker/attention_model_runner.py
```diff
@@ -132,8 +132,8 @@ def _send_dp_metadata(
             is_graph_capturing=is_graph_capturing,
             is_warmup=is_warmup,
         )
-        self.afd_connector.update_state_from_dp_metadata(payload)
-        self.afd_connector.send_dp_metadata_list(payload)
+        self.afd_connector.control_plane.update_state_from_dp_metadata(payload)
```
hsliuustc0106: `control_plane` is optional in the new base contract, but the GPU Attention and FFN paths dereference it unconditionally. A connector-driven GPU connector therefore crashes on its first step. Please implement the `Trigger.CONNECTOR` branch here and in `ffn_worker.py`, or reject no-control-plane GPU connectors during initialization and document that restriction.

docs/gpu/NCCL_P2P_CONNECTOR_USER_GUIDE.md
```diff
@@ -39,6 +39,8 @@ Tensor shapes vary per step (token counts per DP rank), so before data transfers
 
 The pairing is 1:1 by index: Attention role rank `r` sends the control payload to FFN rank `r`. The payload carries the whole per-rank token-count vector plus graph-capture/warmup flags, so one sender per FFN rank is enough — each FFN rank derives every Attention peer's token count from that vector, including peers that are not in the control-plane group.
 
+In code, the control plane is a pluggable module rather than part of the connector interface. `P2pNcclAFDConnector` creates a `P2pNcclAFDControlPlane` instance — an implementation of the `AFDControlPlane` contract from `afd_plugin/connectors/base.py` — and exposes it as `connector.control_plane`. The Attention-side runner calls `control_plane.update_state_from_dp_metadata(...)` and `control_plane.send_dp_metadata_list(...)` before each step; the FFN worker loop blocks on `control_plane.recv_dp_metadata_list()`. The connector sets `ffn_step_trigger = Trigger.DP_METADATA`, meaning each FFN step is triggered by the arrival of a DP metadata payload (connectors without a DP metadata control plane leave `control_plane` as `None` and use `Trigger.CONNECTOR` to drive FFN steps from the connector itself).
```
hsliuustc0106: Please update the existing GPU/NPU runtime design docs in the same change. `docs/npu/NPU_ATTENTION_RUNTIME_DESIGN.md` still tells readers to call `connector.update_state_from_dp_metadata(...)` and `connector.send_dp_metadata_list(...)`, and `async_cam.py` still references the removed `uses_dp_metadata_control_plane` attribute.

afd_plugin/v1/worker/npu/attention_model_runner.py
```diff
@@ -1235,9 +1235,7 @@ def _install_afd_metadata_on_forward_context(
         if forward_context.additional_kwargs is None:
```
hsliuustc0106, line 1246: @jiangkuaixue123 we should have changed all of the ascend folder names into npu

jiangkuaixue123: ok

afd_plugin/connectors/base.py
specture724: This PR aims to refactor / redesign the connector architecture and this repo hasn't been deployed to anywhere yet. I removed all the calling to `AFDConnectorBase.update_state_from_dp_metadata`, `AFDConnectorBase.send_dp_metadata_list` and `AFDConnectorBase.recv_dp_metadata_list`. So no redundency code needs to be remained

afd_plugin/connectors/base.py
specture724: I removed all the string value control plane indicator in the three connectors, so there will not be connectors registered through the public factory that still uses the previous string value

afd_plugin/v1/worker/attention_model_runner.py
specture724: I'll add

docs/gpu/NCCL_P2P_CONNECTOR_USER_GUIDE.md
specture724: I'll update

afd_plugin/connectors/gpu/p2p.py
```diff
@@ -714,6 +601,172 @@ def _current_ubatch_idx() -> int:
...
+
+        Args:
+            payload: DP metadata control-plane payload with per-stage token
+                counts and graph-capturing/warmup flags.
+        """
+        connector = self.connector
+        connector.dp_metadata_list = payload.dp_metadata_list
+        connector.is_graph_capturing = payload.is_graph_capturing
+        connector.is_warmup = payload.is_warmup
+        connector._tensor_metadata_list = {}
```
jiangkuaixue123: 那是不是不能用私有变量 我看有下划线开头

afd_plugin/v1/worker/ffn_model_runner.py
```diff
@@ -183,6 +183,40 @@ def _ffn_forward(
                     self.connector.send_ffn_output(rank_ffn_output, metadata)
         return rank_ffn_output
 
+    def execute_connector_driven_step(self) -> None:
```
jiangkuaixue123: gpu侧的这个路径当前还没有 是不是直接raise  下面的_ffn_forward_connector_driven也先不要了

afd_plugin/v1/worker/ffn_worker.py
```diff
@@ -123,7 +123,13 @@ def _run_ffn_server_loop(self) -> None:
             torch.cuda.set_device(self.device)
 
         while not event.is_set():
-            payload = self.model_runner.connector.recv_dp_metadata_list()
+            if self.model_runner.connector.control_plane is None:
```
jiangkuaixue123, line 126: 或者这儿先raise not implenment

PR #144, opened 2026-07-22
Preserve native model registrations for non-AFD workers

### LINKED ISSUE #141: [Bug]: Installing afd-plugin overrides model implementations for non-AFD paths
## Summary

Installing `afd-plugin` currently overrides model implementations globally, including models that are running through a non-AFD path. Merely having the plugin installed should not change the model code used by standard vLLM workloads.

AFD-specific model implementations or patches must only be activated when the AFD execution path is explicitly selected. Non-AFD workloads must continue to execute the original upstream vLLM model implementation.

## Current environment

```text
Python: Not environment-specific
OS: Not environment-specific
afd-plugin commit: Current main branch
vLLM version/checkout: v0.19.1
CUDA/GPU: Not hardware-specific
Install command: Install afd-plugin in the same environment as vLLM
```

## Reproduction

1. Install `afd-plugin` in a vLLM environment.
2. Start a normal vLLM workload without selecting an AFD worker, runner, connector, or `additional_config["afd"]`.
3. Load a model for which afd-plugin registers or patches an AFD-specific implementation.
4. Observe that the plugin-provided model implementation replaces or modifies the upstream implementation even though the workload is not using AFD.

## Expected behavior

- Installing `afd-plugin` alone must not alter standard vLLM model execution.
- AFD-specific model registration and patch behavior must be gated by explicit AFD activation.
- When the AFD path is not active, model construction and forward execution must use the original upstream vLLM code unchanged.
- AFD and non-AFD workloads should be able to coexist in the same installed environment without affecting each other.

## Actual behavior and logs

```text
The afd-plugin model override is applied globally after installation/plugin loading.
As a result, non-AFD model paths execute the plugin-provided or patched model code
instead of the original upstream vLLM implementation.
```

## AFD configuration

```yaml
# No AFD configuration is provided.
# The workload is intentionally using the standard non-AFD vLLM path.
```

## Suggested validation

- Add a regression test that installs/loads the plugin, starts a non-AFD model path, and verifies that the upstream model class and behavior remain unchanged.
- Add a complementary test confirming that the AFD-specific implementation is selected only when AFD is explicitly enabled.

## Before submitting

- [x] This is described against the repository's pinned vLLM v0.19.1 target.

jiaran-king:
## Summary

- preserve vLLM's native DeepSeek architecture registrations when the plugin loads
- register AFD model implementations under explicit `AFD*` aliases
- switch only AFD GPU/NPU workers to a copied model configuration using those aliases
- update the model-integration documentation and add regression coverage

## Root cause

`register_afd()` registered AFD wrappers under the checkpoint's native architecture names. Because vLLM loads general plugins automatically, installing the plugin globally replaced the native model mappings even when no AFD worker or `additional_config["afd"]` was selected.

## Impact

Standard vLLM workloads now retain the upstream model implementation when afd-plugin is installed. Explicit AFD workers continue to select the role-aware AFD implementation without mutating the original model configuration.

## Validation

- `ruff check .`
- `ruff format --check .`
- `tests/unit`: 248 passed, 29 skipped
- `git diff --check`
- DeepSeek-V2-Lite standard vLLM GPU smoke without AFD configuration: native model load and completion succeeded
- DeepSeek-V2-Lite AFD 1A1F eager GPU smoke: completion succeeded
- DeepSeek-V2-Lite AFD 1A1F `FULL_DECODE_ONLY` CUDA Graph smoke: graph capture and completion succeeded

NPU hardware validation was not run; NPU unit cases were included in the unit suite where the environment supported collection.

Fixes #141

### DIFF
--- a/afd_plugin/__init__.py
+++ b/afd_plugin/__init__.py
@@ -128,7 +128,7 @@ def register_afd() -> None:
     from vllm.model_executor.models import ModelRegistry
 
     for model_arch, model_cls in _DEEPSEEK_MODEL_REGISTRATIONS.items():
-        ModelRegistry.register_model(model_arch, model_cls)
+        ModelRegistry.register_model(f"AFD{model_arch}", model_cls)
 
     _registered = True
 
--- a/afd_plugin/model_executor/models/model_utils.py
+++ b/afd_plugin/model_executor/models/model_utils.py
@@ -1,0 +1,25 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
+"""Utilities for AFD model configuration."""
+
+from __future__ import annotations
+
+from copy import copy
+from typing import TYPE_CHECKING
+
+from afd_plugin import _DEEPSEEK_MODEL_REGISTRATIONS
+
+if TYPE_CHECKING:
+    from vllm.config import ModelConfig
+
+
+def get_afd_model_config(model_config: ModelConfig) -> ModelConfig:
+    """Return a model config that resolves to an AFD model implementation."""
+
+    for model_arch in model_config.hf_config.architectures:
+        if model_arch in _DEEPSEEK_MODEL_REGISTRATIONS:
+            afd_model_config = copy(model_config)
+            afd_model_config.hf_config = copy(model_config.hf_config)
+            afd_model_config.hf_config.architectures = [f"AFD{model_arch}"]
+            return afd_model_config
+    return model_config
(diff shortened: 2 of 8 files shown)


afd_plugin/model_executor/models/__init__.py
```diff
@@ -2,13 +2,35 @@
...
     AsyncMoeUbatchMetadata,
     get_afd_metadata_from_forward_context,
     get_async_moe_ubatch_metadata_from_forward_context,
 )
 
+if TYPE_CHECKING:
+    from vllm.config import ModelConfig
+
+
+def _get_afd_model_config(model_config: ModelConfig) -> ModelConfig:
```
jiangkuaixue123: 这个换一个合适的文件 并且不要用_开头 这个表示私有函数尽在本文件中使用

jiaran-king: 看了一下现有的文件，暂时没有合适的承载位置， 新增了一个很小的 models/utils.py。

PR #152, opened 2026-07-24
feat(npu): token-balanced async MoE microbatch split for DP+TP/SP

### LINKED ISSUE #149: [Feature]: Support token-based CAMAsync microbatch splitting for non-PCP DP+TP/SP topologies

`CAMAsyncAFDConnector` currently supports AFD-managed asynchronous MoE ubatching by splitting prefill work into exactly two microbatches at request boundaries. This policy is suitable for the existing PCP-oriented path, but it does not cover CAMAsync deployments where Attention is not using PCP and the topology is instead DP+TP or DP+SP.

In non-PCP DP+TP/SP deployments, request lengths can differ significantly. Splitting only by request count can therefore produce heavily imbalanced microbatches, reduce Attention/FFN overlap, and leave one pipeline stage much larger than the other.

Add a token-count-based microbatch split policy for CAMAsync in non-PCP DP+TP/SP topologies so the two stages carry approximately balanced token workloads.

## Proposed change

Extend the plugin-owned CAMAsync async-MoE ubatch planner to support token-based splitting:

- Detect the non-PCP CAMAsync topology and select a token-count-based split for DP+TP and DP+SP execution.
- Split the flattened prefill token workload into exactly two microbatches with token counts as balanced as possible.
- Preserve token order and rebuild all stage-local request/token offsets, sequence lengths, routing metadata, and Attention metadata required by TP/SP execution.
- Ensure every input token belongs to exactly one microbatch, with no loss or duplication.
- Keep CAM dispatch/combine payloads and pending state paired by microbatch/stage.
- Preserve the existing PCP request-boundary behavior and avoid changing native vLLM DBO semantics.
- Expose the behavior through the existing `connector_extra_config` namespace, either by adding a token split value such as `async_moe_split="token"` or by resolving the policy automatically from the validated topology. The final configuration contract should fail early for unsupported combinations.

Initial scope is CAMAsync prefill with `async_moe_ubatching=true` and exactly two microbatches.

Preferred extension point: the plugin-owned CAMAsync async-MoE ubatch planner and stage metadata construction used by `AFDNPUAttentionModelRunner` and the CAMAsync DeepSeek forward path.

Compat shim needed: only if the pinned vLLM v0.19.1 / vLLM-Ascend v0.19.1rc1 metadata types require version-specific slicing or reconstruction for DP/SP fields.

Monkey patch needed: none expected. The implementation should not modify the vLLM or vLLM-Ascend source tree.

## Alternatives considered

- Continue using request-boundary splitting for all topologies. This is simple but can be badly imbalanced when request token lengths are skewed.
- Use native vLLM DBO. CAMAsync currently rejects native DBO, and async MoE ubatching has different scheduling and connector semantics.
- Enable PCP solely to reuse the existing policy. This unnecessarily constrains deployments that require DP+TP/SP.

## Validation plan

- CPU-safe unit tests for balanced and skewed token distributions, including odd token counts, a single long request, multiple short requests, and padded batches.
- Assert complete token coverage, stable ordering, no overlap, and correct reconstruction of stage-local token/request offsets and sequence metadata.
- Validate topology selection: PCP keeps the current request-boundary policy; non-PCP DP+TP and DP+SP select or accept the token policy; unsupported combinations fail early.
- NPU E2E coverage for CAMAsync prefill with DP+TP and DP+SP, `async_moe_ubatching=true`, and two microbatches.
- Compare output accuracy with the non-ubatched CAMAsync baseline and verify dispatch/combine stage ordering.
- Add a skewed-length workload check showing that the two microbatch token counts are approximately balanced.
- Confirm no regression for the existing PCP CAMAsync recipe.

## Acceptance criteria

- CAMAsync async-MoE ubatching works in non-PCP DP+TP and DP+SP topologies.
- The two microbatches are split according to token count rather than request count and are approximately balanced.
- TP/SP metadata and CAM routing state remain correct for both stages.
- PCP behavior remains unchanged.
- The implementation stays within plugin-owned scheduling/metadata code and supports the pinned vLLM/vLLM-Ascend baseline without source-tree changes.

## Before submitting

- [x] I identified this as plugin-owned CAMAsync scheduling and metadata work, with a versioned compat helper only if required by pinned upstream metadata contracts.

ShwStone:
CAMAsyncAFDConnector previously split async MoE prefill work into two microbatches only at request boundaries. This works for PCP deployments but produces heavily imbalanced stages in non-PCP DP+TP/SP topologies where request lengths differ.

Add a token-count-based split policy:

- Introduce `async_moe_split` connector_extra_config with values `"request"` (default, preserves PCP behavior) and `"token"`.
- On non-PCP DP+TP/SP topologies, split the padded prefill token workload into two TP-aligned stages of approximately equal size.
- Rebuild stage-local request/token offsets, sequence lengths, and Ascend attention metadata for each stage.
- Add SP-local ubatch slices (`sp_local_ubatch_slices`) so model-side stage slicing works when sequence parallelism shards tokens across TP ranks.
- Move stage-input slicing into `ubatch_sp.py` and keep the forward orchestration in `deepseek_v2_async_cam_forward.py` minimal.

Startup validation fail-fast:
- Reject `async_moe_split="token"` on unsupported topologies.
- Warn when running `"request"` on a topology that supports `"token"`.

Tests cover request-boundary vs token-balanced selection, TP-size alignment, padded/odd token counts, SP-local token mapping, and topology validation.

## Purpose

Implement the token-balanced async MoE microbatch split for non-PCP DP+TP/SP topologies, as described in #149. The goal is to avoid heavily imbalanced pipeline stages when request lengths are skewed, while keeping the existing PCP request-boundary behavior unchanged.

## Issue

- Related issue(s): #149
- Closing keyword, only if fully resolved: Closes #149

## Scope

- In scope:
  - CAMAsync async MoE ubatch planning for prefill with exactly two microbatches.
  - Token-balanced split on non-PCP DP+TP/SP topologies.
  - TP-aligned split points and padded batch handling.
  - Stage-local attention metadata reconstruction.
  - SP-local ubatch slices for sequence-parallel tensor slicing.
  - Startup validation and user-facing `async_moe_split` config contract.
  - CPU-safe unit tests for planner and SP-local slicing.
- Out of scope:
  - Native vLLM DBO behavior changes.
  - Decode context parallel support.
  - More than two microbatches.
  - GPU-specific implementations.

## Implementation Notes

- The split policy is selected in plugin-owned `afd_plugin.v1.worker.npu.ubatch_utils.create_async_moe_ubatch_slices`, which returns both the slices and the applied policy string (`request_boundary` or `token_balanced_tp`).
- Topology capability is shared between runtime planning and startup validation via `enable_token_balanced_async_moe_split`.
- SP-local slice mapping is isolated in `afd_plugin.model_executor.models.npu.ubatch_sp.build_async_moe_stage_inputs`; the forward orchestration only consumes the resulting stage tensors and slice metadata.
- No changes are made to the vLLM or vLLM-Ascend source checkout; all extension points are plugin-owned classes/functions.
- All new imports are CPU-safe; NPU-specific modules are stubbed in unit tests.

## Test Plan

- CPU-only unit tests:
  - `tests/unit/v1/worker/test_npu_runtime.py` — ubatch planning, SP-local slice mapping, validation warnings/errors.
  - `tests/unit/model_executor/models/test_forward_context.py` — forward orchestration structure assertions.
- GPU/NPU-gated E2E (planned, requires hardware):
  - CAMAsync prefill with DP+TP and DP+SP, `async_moe_ubatching=true`, two microbatches.
  - Output accuracy comparison against non-ubatched CAMAsync baseline.
  - Skewed-length workload balance check.
  - Existing PCP CAMAsync recipe regression check.

## Test Result

- CPU unit tests pass locally: `py_compile`, `ruff check`, `ruff format`.
- Full pytest suite was not executed; NPU E2E validation is pending hardware access.

## Docs Impact

- Files updated: inline code comments and docstrings in modified modules.
- No standalone documentation changes; the feature is controlled through `connector_extra_config.async_moe_split`.

---

<details>

- [x] Scope is bounded.

</details>

### DIFF
--- a/afd_plugin/model_executor/models/forward_context.py
+++ b/afd_plugin/model_executor/models/forward_context.py
@@ -18,7 +18,44 @@ from afd_plugin.connectors import AFDForwardContextMetadata
 ASYNC_MOE_UBATCH_METADATA_KEY: Final[str] = "afd_async_moe_ubatch_metadata"
 
 
-class AsyncMoeUbatchMetadata(TypedDict):
+class _AsyncMoeUbatchMetadataOptional(TypedDict, total=False):
+    """Optional SP-local fields for ``AsyncMoeUbatchMetadata``.
+
+    A separate ``total=False`` base class is used instead of
+    ``typing.NotRequired`` because the plugin supports Python 3.10, where
+    ``NotRequired`` is only available via the external ``typing_extensions``
+    package. Keeping the base class avoids adding a new runtime dependency.
+    """
+
+    # SP-local stage layout: when SP shards the full batch across TP ranks,
+    # transpose it into equal per-stage rank shards before attention and
+    # restore the original layout after the async MoE pipeline.
+    use_sp_stage_resharding: bool
+    sp_local_stage_slices: UBatchSlices
+    stage_actual_token_counts: list[int]
+    sp_local_stage_actual_token_counts: list[int]
+
+
+class AsyncMoeUbatchMetadata(_AsyncMoeUbatchMetadataOptional):
+    """Async MoE ubatch sidecar metadata carried by the forward context.
+
+    The required fields (``attn_metadata``, ``ubatch_slices``) are always
+    populated by the attention model runner before the model forward starts.
+    The optional layout fields describe padded stage inputs and their real
+    token coverage:
+
+    - ``use_sp_stage_resharding``: set to ``True`` by the attention model
+      runner when Ascend sequence parallelism is active and the applied split
+      is TP-aligned.
+    - ``sp_local_stage_slices``: set by
+      ``build_async_moe_stage_inputs`` during model forward; its lengths
+      describe the equal per-rank stage shards, not ranges into the original
+      full-batch local shard.
+    - ``stage_actual_token_counts`` and
+      ``sp_local_stage_actual_token_counts`` distinguish real token rows from
+      DP/SP padding globally and on the current TP rank.
+    """
+
     attn_metadata: object
     ubatch_slices: UBatchSlices
 
--- a/afd_plugin/model_executor/models/npu/ubatch_sp.py
+++ b/afd_plugin/model_executor/models/npu/ubatch_sp.py
@@ -1,0 +1,406 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
+"""Sequence-parallel stage layout helpers for async MoE ubatching.
+
+Ascend sequence parallelism stores the full-batch tensor as one contiguous
+rank-major shard per TP rank. Async MoE stages instead need a contiguous
+global stage reconstructed by the attention operator's TP all-gather. The
+helpers below explicitly transpose between those two layouts:
+
+* full-batch rank shards -> global full batch -> per-stage rank shards;
+* per-stage rank shards -> global stages -> original full-batch rank shards.
+
+Slicing each full-batch rank shard independently cannot implement this
+transpose and silently pairs hidden states with another stage's positions.
+"""
+
+from __future__ import annotations
+
+from typing import overload
+
+import torch
+from vllm.distributed import tensor_model_parallel_all_gather
+from vllm.distributed.parallel_state import get_tp_group
+from vllm.v1.worker.ubatch_utils import UBatchSlice, UBatchSlices
+
+
+def build_async_moe_stage_inputs(
+    *,
+    hidden_states: torch.Tensor,
+    residual: torch.Tensor | None,
+    positions: torch.Tensor,
+    llama_4_scaling: torch.Tensor | None,
+    ubatch_slices: UBatchSlices,
+    use_sp_stage_resharding: bool,
+) -> tuple[
+    list[torch.Tensor],
+    list[torch.Tensor | None],
+    list[torch.Tensor],
+    list[torch.Tensor | None],
+    UBatchSlices,
+]:
+    """Build stage inputs in the layout expected by Ascend attention."""
+    if not use_sp_stage_resharding:
+        return _build_non_sp_stage_inputs(
+            hidden_states=hidden_states,
+            residual=residual,
+            positions=positions,
+            llama_4_scaling=llama_4_scaling,
+            ubatch_slices=ubatch_slices,
+        )
+
+    tp_group = get_tp_group()
+    tp_rank = int(tp_group.rank_in_group)
+    tp_size = int(tp_group.world_size)
+    global_num_tokens = sum(
+        int(ubatch_slice.num_tokens) for ubatch_slice in ubatch_slices
+    )
+    if tp_size <= 1 or int(hidden_states.shape[0]) == global_num_tokens:
+        return _build_non_sp_stage_inputs(
+            hidden_states=hidden_states,
+            residual=residual,
+            positions=positions,
+            llama_4_scaling=llama_4_scaling,
+            ubatch_slices=ubatch_slices,
+        )
+    if global_num_tokens % tp_size != 0:
+        raise ValueError(
+            "Async MoE SP full-batch token count must be divisible by TP size; "
+            f"got num_tokens={global_num_tokens}, tp_size={tp_size}",
+        )
+    expected_local_tokens = global_num_tokens // tp_size
+    if int(hidden_states.shape[0]) != expected_local_tokens:
+        raise ValueError(
+            "Async MoE SP hidden-state layout mismatch: expected "
+            f"{expected_local_tokens} local rows for {global_num_tokens} "
+            f"global tokens at TP={tp_size}, got {int(hidden_states.shape[0])}",
+        )
+
+    # This transpose is paid once before and once after the complete MoE
+    # pipeline, not once per layer. A future optimization can replace the
+    # temporary global tensors with variable-split TP all-to-all after vLLM
+    # exposes a stable backend-neutral API for it.
+    global_hidden_states = _gather_sp_sequence_tensor(
+        hidden_states,
+        token_dim=0,
+        expected_global_tokens=global_num_tokens,
+    )
+    global_residual = (
+        None
+        if residual is None
+        else _gather_sp_sequence_tensor(
+            residual,
+            token_dim=0,
+            expected_global_tokens=global_num_tokens,
+        )
+    )
+    global_positions, positions_token_dim = _to_global_sequence_tensor(
+        positions,
+        local_num_tokens=expected_local_tokens,
+        global_num_tokens=global_num_tokens,
+    )
+    if llama_4_scaling is None:
+        global_llama_4_scaling = None
+        scaling_token_dim = None
+    else:
+        global_llama_4_scaling, scaling_token_dim = _to_global_sequence_tensor(
+            llama_4_scaling,
+            local_num_tokens=expected_local_tokens,
+            global_num_tokens=global_num_tokens,
+        )
+
+    stage_hidden_states: list[torch.Tensor] = []
+    stage_residual: list[torch.Tensor | None] = []
+    stage_positions: list[torch.Tensor] = []
+    stage_llama_4_scaling: list[torch.Tensor | None] = []
+    sp_local_stage_slices: UBatchSlices = []
+    local_stage_start = 0
+
+    for ubatch_slice in ubatch_slices:
+        stage_tokens = int(ubatch_slice.num_tokens)
+        if stage_tokens % tp_size != 0:
+            raise ValueError(
+                "Async MoE SP stage token count must be divisible by TP size; "
+                f"got token_slice={ubatch_slice.token_slice}, tp_size={tp_size}",
+            )
+        local_stage_tokens = stage_tokens // tp_size
+        global_stage_start = int(ubatch_slice.token_slice.start)
+        local_global_start = global_stage_start + tp_rank * local_stage_tokens
+        local_global_stop = local_global_start + local_stage_tokens
+        local_global_slice = slice(local_global_start, local_global_stop)
+
+        stage_hidden_states.append(global_hidden_states[local_global_slice])
+        stage_residual.append(
+            _slice_optional_first_dim(global_residual, local_global_slice),
+        )
+        stage_positions.append(
+            _slice_sequence_dim(
+                global_positions,
+                positions_token_dim,
+                local_global_slice,
+            ),
+        )
+        stage_llama_4_scaling.append(
+            None
+            if global_llama_4_scaling is None or scaling_token_dim is None
+            else _slice_sequence_dim(
+                global_llama_4_scaling,
+                scaling_token_dim,
+                local_global_slice,
+            )
+        )
+
+        local_stage_stop = local_stage_start + local_stage_tokens
+        sp_local_stage_slices.append(
+            UBatchSlice(
+                ubatch_slice.request_slice,
+                slice(local_stage_start, local_stage_stop),
+            ),
+        )
+        local_stage_start = local_stage_stop
+
+    return (
+        stage_hidden_states,
+        stage_residual,
+        stage_positions,
+        stage_llama_4_scaling,
+        sp_local_stage_slices,
+    )
+
+
+def restore_async_moe_stage_outputs(
+    stage_outputs: list[torch.Tensor],
+    ubatch_slices: UBatchSlices,
+    *,
+    use_sp_stage_resharding: bool,
+) -> torch.Tensor:
+    """Restore stage outputs to the model's original full-batch SP layout."""
+    if not use_sp_stage_resharding:
+        return torch.cat(stage_outputs, dim=0)
+
+    tp_group = get_tp_group()
+    tp_rank = int(tp_group.rank_in_group)
+    tp_size = int(tp_group.world_size)
+    global_num_tokens = sum(
+        int(ubatch_slice.num_tokens) for ubatch_slice in ubatch_slices
+    )
+    if len(stage_outputs) != len(ubatch_slices):
+        raise ValueError(
+            "Async MoE stage output count does not match ubatch metadata; "
+            f"got {len(stage_outputs)} outputs and {len(ubatch_slices)} slices",
+        )
+
+    global_stage_outputs: list[torch.Tensor] = []
+    for stage_output, ubatch_slice in zip(
+        stage_outputs,
+        ubatch_slices,
+        strict=True,
+    ):
+        global_stage_outputs.append(
+            _gather_sp_sequence_tensor(
+                stage_output,
+                token_dim=0,
+                expected_global_tokens=int(ubatch_slice.num_tokens),
+            ),
+        )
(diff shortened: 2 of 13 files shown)


jiangkuaixue123:
Test plan: Let’s first use DSV2-Lite to validate accuracy with TP/SP u-batch splitting.

afd_plugin/model_executor/models/npu/ubatch_sp.py
```diff
@@ -0,0 +1,261 @@
...
+    padded_tokens = sp_local_token_count(stage_tokens, tp_size) * tp_size
+    if padded_tokens > stage_tokens:
+        stage_tensor = _pad_sequence_tensor(
+            stage_tensor,
+            padded_tokens - stage_tokens,
+        )
+    tokens_per_rank = padded_tokens // tp_size
+    local_start = tp_rank * tokens_per_rank
+    local_stop = local_start + tokens_per_rank
+    return _slice_sequence_tensor(stage_tensor, slice(local_start, local_stop))
```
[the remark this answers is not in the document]
ShwStone: this is a bug

jiangkuaixue123:
Please fix DCO and CI.

jiangkuaixue123:
**Scan:**
| Category | Result |
|---|---|
| Tests / verification | 3 finding(s) below |
| Docs / comments | 2 finding(s) below |
| Correctness | no finding reported |

**Validated:**
- [sweep] Checked ASYNC_MOE_REQUEST_SPLIT consumers via grep: feature_validation.py (updated), async_cam.py (constant unchanged, docstring updated). No stale consumers.
- [sweep] Checked create_request_boundary_ubatch_slices callers via grep: attention_model_runner.py now calls create_async_moe_ubatch_slices; create_request_boundary_ubatch_slices still exported and used as fallback. Tests updated.
- [sweep] Checked AsyncMoeUbatchMetadata consumers via grep: forward_context.py (definition extended with optional fields), deepseek_v2_async_cam_forward.py (reads new sp_local_ubatch_slices via .get with fallback), deepseek_v2.py (reads via get_async_moe_ubatch_metadata_from_forward_context). Backward-compatible.
- [sweep] Checked removed _slice_optional_first_dim, _slice_positions, _slice_llama_4_scaling: all were private to deepseek_v2_async_cam_forward.py. No external consumers found — safe removal.
- [sweep] Checked async_moe_split default: ASYNC_MOE_REQUEST_SPLIT = "request" unchanged. New ASYNC_MOE_TOKEN_SPLIT = "token" constant added. AFDAsyncExtraInfo.async_moe_split default remains "request".
- [sweep] Checked docs/npu/CAM_ASYNC_CONNECTOR_USER_GUIDE.md:154 — says 'request-boundary splitting only', stale after PR. Line in 'Current limitations' also says 'two request-boundary stages'. Must document "token".
- [sweep] Checked docs/design/module/connector_contracts.md:94 — says 'optional request-boundary pipeline', stale. docs/design_zh/module/connector_contracts.md:77 same.
- [sweep] Checked recipe/npu/CAMAsyncAFDConnector/deepseek_v3_2/README.md:72 — describes 'request-level splitting' for this recipe, still accurate but doesn't acknowledge "token" option.

`afd_plugin/v1/worker/npu/attention_model_runner.py:234` [major] — The diff changes `_build_attention_metadata_with_async_moe_ubatches` to call `create_async_moe_ubatch_slices` and store `use_sp_local_ubatch_slices` in the metadata dict. There are no unit tests for this path. Add a CPU unit test that instantiates `AFDNPUAttentionModelRunner` with `async_moe_ubatching=True`, `split='token'`, and a suitable batch; verify the metadata contains `use_sp_local_ubatch_slices` and the correct `ubatch_slices`. (evidence: grep pattern '_build_attention_metadata_with_async_moe_ubatches|async_moe_ubatching' in tests/ returned zero matches; read_file of tests/unit/v1/worker/test_npu_runtime.py confirmed no test exercises this method.)

`afd_plugin/model_executor/models/npu/ubatch_sp.py:190` [major] — `_slice_sequence_tensor_for_sp_stage` has branches for `llama_4_scaling` and 2D positions that are not exercised by `test_ubatch_sp_stage_inputs_use_sp_local_ranges` (which uses 1D positions and `llama_4_scaling=None`). Add test cases with 2D `positions` and nonzero `llama_4_scaling` to cover the `local_token_dim==1` and `global_token_dim==1` paths. (evidence: read_file of test_ubatch_sp_stage_inputs_use_sp_local_ranges shows only 1D positions and llama_4_scaling=None; read_file of _slice_sequence_tensor_for_sp_stage shows branches at lines 198, 200, 204, 207, 209 that are not exercised.)

`docs/npu/CAM_ASYNC_CONNECTOR_USER_GUIDE.md:154` [minor] — The diff adds `async_moe_split="token"` support, but `docs/npu/CAM_ASYNC_CONNECTOR_USER_GUIDE.md:154` still states 'The current async connector supports request-boundary splitting only.' Update this line, the 'AFD-managed asynchronous MoE ubatching' paragraph, and the 'Current limitations' bullet to document the token split, its topology preconditions (non-PCP, TP>1, DP+TP/SP), and use cases; similarly update `recipe/npu/CAMAsyncAFDConnector/deepseek_v3_2/README.md:72` and the design docs. (evidence: grep for `async_moe_split` across docs/ found 5 locations restricting the field to request-boundary only; diff adds `ASYNC_MOE_TOKEN_SPLIT = "token"` in async_cam.py:68 and `"token"` is accepted in feature_validation.py:141-155.)

`afd_plugin/model_executor/models/npu/ubatch_sp.py:198` [minor] — `_slice_sequence_tensor_for_sp_stage` handles `local_token_dim` 0 and 1 but returns the tensor unchanged when the token dimension is >1 (and not detected by `global_token_dim`). The sibling `_slice_positions` uses `positions[..., token_slice]` for higher dims. If a sequence tensor with tokens on a higher axis reaches this path, the input may be passed through incorrectly. Add an ellipsis fallback or assert that the dimensionality is ≤2, and document the assumption. (evidence: Read `ubatch_sp.py` lines 185-227 (`_slice_sequence_tensor_for_sp_stage`): only `local_token_dim == 0` → `tensor[tensor_token_slice]` and `local_token_dim == 1` → `tensor[:, tensor_token_slice]`. Compare with `_slice_positions` at line 240: `positions[..., token_slice]` for dim>1.)

`afd_plugin/model_executor/models/npu/ubatch_sp.py:131` [minor] — In `_build_async_moe_stage_inputs_with_slices`, `stage_hidden_states` uses bare indexing `hidden_states[slice]` while `stage_residual` uses `_slice_and_pad_first_dim`. Both slices are exact (non‑SP path), so the asymmetry is functionally correct but confusing—and the failure modes differ: out‑of‑bounds on hidden_states would crash while residual would silently pad. Use the same slicing helper (e.g., `_slice_and_pad_first_dim`) for both, or add a comment explaining the intent. (evidence: Read `ubatch_sp.py` lines 125-140: `stage_hidden_states = [hidden_states[sp_local_ubatch_slice.token_slice] ...]` vs `stage_residual = [_slice_and_pad_first_dim(residual, sp_local_ubatch_slice.token_slice) ...]`. The old code in deepseek_v2_async_cam_forward.py used `_slice_optional_first_dim` (just a bare slice) for both.)

**Verdict:** REQUEST CHANGES

afd_plugin/model_executor/models/forward_context.py
```diff
@@ -18,7 +18,15 @@
 ASYNC_MOE_UBATCH_METADATA_KEY: Final[str] = "afd_async_moe_ubatch_metadata"
 
 
-class AsyncMoeUbatchMetadata(TypedDict):
+class _AsyncMoeUbatchMetadataOptional(TypedDict, total=False):
```
jiangkuaixue123, line 21: Could you add a class docstring explaining why the SP-local microbatch metadata is optional and when each field is populated and consumed?

afd_plugin/model_executor/models/forward_context.py
```diff
@@ -18,7 +18,15 @@
...
+    # SP-local ubatch slices: when SP shards tokens across TP ranks, slice
+    # model tensors by per-rank local token ranges derived from the global
+    # ubatch slices (an identity mapping when SP is disabled).
+    use_sp_local_ubatch_slices: bool
+    sp_local_ubatch_slices: UBatchSlices
+
+
+class AsyncMoeUbatchMetadata(_AsyncMoeUbatchMetadataOptional):
```
jiangkuaixue123, line 39: Could we define these optional keys directly on `AsyncMoeUbatchMetadata` using `NotRequired[...]` instead of introducing a separate `total=False` base class? That would keep the schema in one place and make the required/optional distinction more explicit.

jiangkuaixue123 (commented):
Two suggestions on the async MoE ubatch metadata type definition.

afd_plugin/v1/worker/npu/attention_model_runner.py
```diff
@@ -248,20 +249,26 @@ def _build_attention_metadata_with_async_moe_ubatches(
         if num_scheduled_tokens_np is None:
             return full_metadata
 
-        ubatch_slices = create_request_boundary_ubatch_slices(
+        ubatch_slices, split_mode = create_async_moe_ubatch_slices(
```
jiangkuaixue123, line 260: Please make sure the changes in this file remain compatible with the PCP path, where microbatches must continue to be split at request boundaries.

jiangkuaixue123 (commented):
One compatibility concern for the updated ubatch split path.

jiangkuaixue123 (commented):
Overall, the structure LGTM. Please address the review comments and validate the accuracy. Once those are complete, this should be ready to merge.

ShwStone:
Pushed a follow-up fix commit to resolve a test collection failure on CPU-only development paths.

**Root cause:** The previous commit added a module-level `from vllm.logger import init_logger` in `afd_plugin/compat/npu/feature_validation.py`. `tests/unit/v1/worker/test_npu_runtime.py` imports `afd_plugin.compat.npu` at module scope, so in CI environments that only have pytest installed (without the vLLM runtime), test collection threw `ModuleNotFoundError: No module named 'vllm'` instead of skipping the NPU runtime tests gracefully.

**Fix:** Replaced `from vllm.logger import init_logger` with the standard library `logging.getLogger(__name__)`; `logger.warning(...)` calls remain unchanged.

afd_plugin/model_executor/models/forward_context.py
ShwStone: Thanks for the suggestion. Using `NotRequired` directly on `AsyncMoeUbatchMetadata` would indeed make the schema more explicit and keep everything in one place.

One constraint: the plugin currently supports Python `>=3.10`, while `typing.NotRequired` was only added in Python 3.11. On Python 3.10 we'd need `typing_extensions.NotRequired`, but the project doesn't currently depend on `typing_extensions`.

Would you prefer:
1. Keep the current `total=False` base class (works on Python 3.10 without new dependencies), or
2. Add `typing_extensions` as a dependency and switch to `NotRequired`?

I'm happy to go with either direction — just let me know your preference.

afd_plugin/v1/worker/npu/attention_model_runner.py
ShwStone: Thanks for flagging this. PCP compatibility is preserved through the split-policy fallback:

- When `async_moe_split="request"` (default), `create_async_moe_ubatch_slices` directly calls `create_request_boundary_ubatch_slices`, which is exactly the original PCP request-boundary logic.
- When `async_moe_split="token"`, startup validation in `feature_validation.py` rejects PCP topologies, so the token-balanced path can never be entered on PCP deployments.
- `use_sp_local_ubatch_slices` is only set to `True` when `split_mode == "token_balanced_tp"`, so PCP paths always use the global ubatch slices.

I've added a docstring on `_build_attention_metadata_with_async_moe_ubatches` to make this contract explicit.

jiangkuaixue123:
Please fix DCO
image

afd_plugin/v1/worker/npu/attention_model_runner.py
jiangkuaixue123: ok 👌

afd_plugin/model_executor/models/forward_context.py
jiangkuaixue123: Good point. Given that we still support Python 3.10 and intentionally keep the core dependency list empty, I’d prefer to keep the current total=False base class for now. Adding typing_extensions solely for NotRequired does not seem necessary. We can simplify this once Python 3.10 support is dropped, or if typing_extensions becomes a direct dependency for other reasons.

afd_plugin/model_executor/models/forward_context.py
ShwStone: Added a class docstring on `AsyncMoeUbatchMetadata` explaining required vs optional fields and when each is populated/consumed.

ShwStone:
Thanks for the thorough scan. All findings have been addressed in the latest push:

**1. [major] `attention_model_runner.py:234` — Missing unit tests for `_build_attention_metadata_with_async_moe_ubatches`**

Added two tests:
- `test_npu_attention_runner_builds_async_moe_ubatch_metadata` (token split)
- `test_npu_attention_runner_builds_async_moe_ubatch_metadata_request_split` (request split)

**2. [major] `ubatch_sp.py:190` — Missing 2D positions / `llama_4_scaling` tests**

Added `test_ubatch_sp_stage_inputs_2d_positions_and_llama4_scaling` covering 2D positions (dim=1) and non-None `llama_4_scaling` (dim=1).

**3. [minor] Docs stale**

Updated `CAM_ASYNC_CONNECTOR_USER_GUIDE.md`, `connector_contracts.md`, and the DeepSeek v3.2 recipe README to document `async_moe_split="token"` and its topology preconditions. Note: `docs/design_zh/module/connector_contracts.md` does not exist in the current tree (checked `origin/main` and all branches), so no change needed there.

**4. [minor] `ubatch_sp.py:198` — Silent pass-through for >1 dim token axis**

`_slice_sequence_tensor_for_sp_stage` now raises `ValueError` when the token dimension is not on axis 0 or 1.

**5. [minor] `ubatch_sp.py:131` — Asymmetric slicing**

Unified `_build_async_moe_stage_inputs_with_slices` to use `_slice_and_pad_first_dim` for both `hidden_states` and `residual`, with `@overload` to keep type inference precise.

DCO sign-off and GPG signature are also included in the latest commit.

ShwStone:
DCO sign-off added. The squashed commit now includes `Signed-off-by` and a GPG signature.

afd_plugin/model_executor/models/forward_context.py
ShwStone: I've added a docstring to explain this

ShwStone:
Closing this PR in favor of a consolidated replacement based on the latest main.

This implementation was developed and debugged while the unrelated model-configuration identity regression reported in #172 was present. That regression materially affected TP2 FlashComm1 behavior and the conclusions drawn while diagnosing this change.

#172 has since been fixed by #174. The replacement rebuilds the feature on the corrected baseline and combines the implementation here with the NPU E2E coverage from #154.

PR #157, opened 2026-07-26
Fix forced load-balancing cycle phases across EP ranks

### LINKED ISSUE #156: [Bug]: Forced load-balancing cycle starts from the same prefix on every EP rank
### Current environment

```text
Python: N/A (static and CPU-testable routing analysis)
OS: Linux
afd-plugin commit: ea7c56a
vLLM version/checkout: Targeted compatibility version v0.19.1
CUDA/GPU: Not available
Ascend NPU: Not available
Install command: uv sync --locked --group dev
```

### Summary

Forced load balancing builds the same deterministic expert cycle on every
source EP rank, and `_get_force_lb_topk_ids()` always returns it from row zero:

```python
return buffer[:batch_tokens, :top_k]
```

This is not the usual case where an incomplete cycle gives a few experts one
extra assignment. Because every source rank repeats the **same prefix**, their
partial-cycle load is perfectly correlated and amplified by the number of
source ranks. Some target ranks can receive substantially more work, while
others can receive none.

This conflicts with the goal in #40 that every EP rank should receive the same
routed-token volume during communication profiling.

### Reproduction and impact

Consider the published EP64 BS16 configuration:

```text
n_routed_experts = 256
ep_size = 64
top_k = 8
force_load_balance_topn_per_rank = 4
batch_tokens = 16 per source rank
```

The imbalance follows directly from the cycle layout:

1. The cycle contains 64 target-rank blocks × 4 expert IDs = 256 IDs.
2. Each source rank emits 16 tokens × top_k 8 = 128 expert assignments.
3. The first 128 IDs cover only the first 128 / 4 = 32 target-rank blocks.
4. All 64 source ranks restart from ID 0, so they all select the same 32 blocks.

Each selected target rank receives four assignments from each source rank:

```text
64 source ranks × 4 assignments = 256 assignments
```

The other 32 target ranks are never reached:

```text
target ranks 0-31:  256 assignments per rank
target ranks 32-63:   0 assignments per rank
```

The same calculation for the other published configurations gives:

```text
48A16F BS42:     target ranks 0-3 receive 384 assignments each;
                 target ranks 4-15 receive 320.
64A16F BS56:     all target ranks receive 448 only because BS56 happens
                 to cover an integer number of cycles.
```

### Possible fix

Arrange the cycle in target-rank blocks and rotate it by one block for each
source EP rank before building the precomputed buffer:

```python
selected_experts_per_rank = (
    config.topn_per_rank
    if config.topn_per_rank > 0
    else config.n_routed_experts // config.ep_size
)
source_rank_offset = config.ep_rank * selected_experts_per_rank
expert_cycle = torch.roll(expert_cycle, shifts=-source_rank_offset)
```

This remains deterministic, requires no additional communication or global
state, and preserves the existing configuration interface.

### CPU regression tests

The fix can be verified without NPU hardware:

1. Build the buffer for every simulated source EP rank.
2. Aggregate the generated expert IDs.
3. Assert equal assignment counts across target EP ranks.
4. Cover the published `(ep_size, batch_tokens)` cases `(64, 16)`, `(16, 42)`,
   and `(16, 56)`.

### Expected behavior

Partial cycles from different source EP ranks should use different phases so
that their aggregate routed-token counts are equal across target EP ranks.

### Actual behavior and logs

```text
There is no runtime exception. All source EP ranks reuse the same cycle prefix.

For EP64 BS16:
- target ranks 0-31 receive 256 assignments per rank
- target ranks 32-63 receive 0 assignments per rank
```

### AFD configuration

```yaml
enable_force_load_balance: true
force_load_balance_topn_per_rank: 4
```

### Before submitting

- [x] I confirmed whether this reproduces against vLLM v0.19.1 or explained why not.

sinpeyw:
## Purpose

Fix forced load balancing so partial expert cycles from different source EP ranks do not all reuse the same target-rank prefix.

Previously, every source EP rank built the same deterministic cycle and consumed it from row zero. When a routing batch covered only part of the cycle, the identical prefixes were correlated across source ranks and could leave some target ranks idle.

## Issue

- Related issue(s): #156
- Closing keyword, only if fully resolved: Closes #156

## Scope

- In scope:
  - Give each source EP rank a distinct target-rank phase.
  - Preserve the existing deterministic cycle order, fixed seed, selected local experts, and configuration interface.
  - Cover both `force_load_balance_topn_per_rank > 0` and the all-expert `topn_per_rank = 0` mode.
  - Add CPU regression coverage for the published routing batch sizes.
- Out of scope:
  - Changes to natural model routing or router weights.
  - Changes to MoE communication collectives.
  - NPU performance benchmarking.

## Implementation Notes

`ForceLoadBalanceConfig` now records the source `ep_rank`. After constructing the existing expert cycle, the implementation shifts every expert ID by:

```text
ep_rank * local_routed_experts
```

and wraps it by `n_routed_experts`.

The offset is an integer number of target-rank expert blocks. It changes only the target-rank phase while preserving each expert's local index. Consequently, every cycle position maps once to every target rank when aggregated across equally sized source EP batches.

For `topn_per_rank = 0`, the existing deterministic `randperm` is retained. For `topn_per_rank > 0`, the same selected local expert IDs remain in use.

## Test Plan

- Run repository Ruff lint and format checks.
- Run the repository CPU-only unit-test command.
- Run `test_force_load_balance.py` with CPU PyTorch available.
- Run pre-commit on the modified files.
- Leave NPU runtime validation to CI or a maintainer with Ascend hardware.

## Test Result

- `uv run --group dev ruff check .` — passed.
- `uv run --group dev ruff format --check .` — passed.
- `uv run --group dev pytest -q tests/unit -m "not gpu and not vllm_runtime"` — passed (2 skipped).
- `tests/unit/compat/patches/test_force_load_balance.py` in a CPU PyTorch environment — 12 passed.
- `uv run --group dev pre-commit run --files afd_plugin/compat/patches/npu/force_load_balance.py tests/unit/compat/patches/test_force_load_balance.py` — passed.
- NPU runtime tests — not run because Ascend hardware is unavailable.

## Docs Impact

- Files updated: none.
- If none, reason: this fixes the existing configuration behavior without changing its interface.

---

<details>

- [x] Scope is bounded.

</details>

### DIFF
--- a/afd_plugin/compat/patches/npu/force_load_balance.py
+++ b/afd_plugin/compat/patches/npu/force_load_balance.py
@@ -45,6 +45,7 @@ class ForceLoadBalanceConfig:
     Args:
         n_routed_experts: Number of routed experts in the MoE layer.
         ep_size: Number of expert-parallel ranks.
+        ep_rank: Source expert-parallel rank that builds the fake routing cycle.
         top_k: Number of routed experts selected for each token.
         topn_per_rank: Number of local experts per EP rank used by the fake
             routing cycle. A value of 0 means all routed experts participate.
@@ -52,6 +53,7 @@ class ForceLoadBalanceConfig:
 
     n_routed_experts: int
     ep_size: int
+    ep_rank: int
     top_k: int
     topn_per_rank: int
 
@@ -67,22 +69,25 @@ def _get_force_lb_config(layer: object) -> ForceLoadBalanceConfig:
     return ForceLoadBalanceConfig(
         n_routed_experts=int(layer.n_routed_experts),
         ep_size=int(layer.ep_size),
+        ep_rank=int(layer.ep_rank),
         top_k=int(layer.top_k),
         topn_per_rank=int(layer.force_load_balance_topn_per_rank),
     )
 
 
 def _validate_force_lb_config(config: ForceLoadBalanceConfig) -> None:
+    assert config.ep_size > 0, "ep_size must be positive"
+    assert 0 <= config.ep_rank < config.ep_size, (
+        "ep_rank must be within the expert-parallel group"
+    )
+    assert config.n_routed_experts % config.ep_size == 0, (
+        "force load balance requires n_routed_experts to be divisible by ep_size"
+    )
+
     if config.topn_per_rank == 0:
         return
 
     assert config.topn_per_rank > 0, "force_load_balance_topn_per_rank must be >= 0"
-    assert config.ep_size > 0, "ep_size must be positive"
-    assert config.n_routed_experts % config.ep_size == 0, (
-        "force_load_balance_topn_per_rank requires n_routed_experts to be"
-        " divisible by ep_size"
-    )
-
     local_routed_experts = config.n_routed_experts // config.ep_size
     assert config.topn_per_rank <= local_routed_experts, (
         "force_load_balance_topn_per_rank exceeds routed experts on each FFN rank"
@@ -96,8 +101,8 @@ def _build_expert_cycle(
     config: ForceLoadBalanceConfig,
     device: torch.device,
 ) -> torch.Tensor:
+    local_routed_experts = config.n_routed_experts // config.ep_size
     if config.topn_per_rank > 0:
-        local_routed_experts = config.n_routed_experts // config.ep_size
         per_rank_cycles = [
             torch.arange(
                 rank * local_routed_experts,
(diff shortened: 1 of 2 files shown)


sinpeyw:
This PR is now ready for review. PR #159 is complementary: it updates the EP64 benchmark configuration, while this PR fixes the underlying shared-prefix issue for both `topn_per_rank = 0` and `topn_per_rank > 0`.

PR #161, opened 2026-07-27
[Feat]: support multistream for CAMP2p

### LINKED ISSUE #68: [Feature]: Enable multi-stream parallelism for CAMP2P A2E/E2A on FFN side

Support enabling multi-stream parallelism for CAMP2P A2E/E2A communication on the FFN side.

When the FFN side is the compute bottleneck, A2E and E2A can overlap with FFN-side computation through multi-stream execution. This should reduce per-layer execution time by improving communication/computation overlap in CAMP2P flows.

## Proposed change

Add support for FFN-side CAMP2P A2E/E2A multi-stream parallel execution.

Expected behavior:
- Allow A2E and E2A on the FFN side to run on separate streams where applicable.
- Overlap A2E/E2A with FFN computation when FFN is the bottleneck.
- Keep the feature configurable so it can be enabled only for supported CAMP2P scenarios.
- Ensure stream synchronization is explicit and safe around tensor ownership, data dependencies, and layer boundaries.

Potential implementation areas may include the FFN model runner/worker path and CAMP2P communication helpers, depending on the current plugin-owned abstraction.

Preferred extension point: plugin-owned FFN runner/worker and CAMP2P communication logic.

Compat shim needed: only if the current plugin interface does not expose enough stream or scheduling control.

Monkey patch needed: avoid modifying the vLLM v0.19.1 source tree. If a patch is unavoidable, keep it isolated, version-guarded, idempotent, and documented.

## Alternatives considered

- Keep A2E/E2A serialized with FFN execution. This is simpler but leaves performance on the table when FFN is compute-bound.
- Apply global communication stream changes. This may be broader than necessary and could affect non-CAMP2P paths, so the first implementation should prefer a scoped FFN-side CAMP2P change.

## Validation plan

- Add or update CPU-safe unit tests for configuration gating and scheduling/stream selection logic where possible.
- Add GPU/NPU-gated validation for CAMP2P FFN-side execution with multi-stream enabled.
- Compare per-layer execution time with and without FFN-side A2E/E2A multi-stream parallelism in an FFN-compute-bottleneck scenario.
- Verify correctness against the existing serialized path, including output consistency and synchronization safety.

## Before submitting

- [x] I identified whether this belongs in plugin-owned code, compat helpers, or compat patches.

yujuancao07:
## Purpose

Add configurable multi-stream execution to the Ascend CAMP2P connector 
so that A2F/F2A communication can overlap with model computation.

Attention-side and FFN-side multi-stream execution can be enabled independently
through `connector_extra_config`. 

## Issue

- Related issue(s): #68 
- Closing keyword, only if fully resolved: Closes #68 

## Scope

This feature applies only to `CAMP2pAFDConnector` on NPU.

- In scope:
  - Attention-side CAMP2P communication stream execution.
  - FFN-side CAMP2P communication stream execution.
  - Independent Attention and FFN configuration switches.
  - Independent Attention and FFN AIV core-count configuration.
  - Eager execution and the existing AFD NPU ubatch/DBO path.
  - Existing ACL graph paths where stream switching can be represented safely by
    the current graph runtime.

- Out of scope:
  - `CAMAsyncAFDConnector`.
  - GPU `P2pNcclAFDConnector`.
 
## Implementation Notes

- The implementation applies only to `CAMP2pAFDConnector`.
- Both switches are located under CAMP2P `connector_extra_config`:

  ```json
  {
    "connector_extra_config": {
      "is_attn_multistream": true,
      "is_ffn_multistream": true,
      "attn_core_num": 8,
      "ffn_core_num": 8
    }
  }

## Test Plan

Focused multi-stream tests:
`pytest -q tests/unit -k multistream -vv`

## Test Result
<img width="2299" height="300" alt="image" src="https://github.com/user-attachments/assets/69efd57e-d200-49b2-afc1-9810ac5747ad" />

### DIFF
--- a/afd_plugin/connectors/npu/camp2p.py
+++ b/afd_plugin/connectors/npu/camp2p.py
@@ -51,6 +51,9 @@ from afd_plugin.connectors.metadata import (
     send_control_payload,
 )
 from afd_plugin.distributed import init_afd_process_group, topology_from_config
+from afd_plugin.v1.worker.npu.multistream import (
+    npu_stream_switch_within_graph,
+)
 
 if TYPE_CHECKING:
     from vllm.config import VllmConfig
@@ -63,6 +66,8 @@ _CAMP2P_EXTRA_CONFIG_FIELDS: Final[frozenset[str]] = frozenset(
         "core_num",
         "attn_core_num",
         "ffn_core_num",
+        "is_attn_multistream",
+        "is_ffn_multistream",
         "compute_gate_on_attention",
         "quant_mode",
     },
@@ -77,6 +82,10 @@ class CAMP2PExtraInfo(ConnectorExtraInfo):
         core_num: Default number of AIV cores used by each AFD role.
         attn_core_num: Optional Attention-role override for ``core_num``.
         ffn_core_num: Optional FFN-role override for ``core_num``.
+        is_attn_multistream: Run Attention-side A2F communication on a
+            dedicated NPU stream.
+        is_ffn_multistream: Run FFN-side F2A communication on a dedicated NPU
+            stream.
         compute_gate_on_attention: Whether Attention computes MoE gate outputs.
         quant_mode: CAM quantization mode; the current runtime supports only 0.
     """
@@ -84,6 +93,8 @@ class CAMP2PExtraInfo(ConnectorExtraInfo):
     core_num: int = 8
     attn_core_num: int | None = None
     ffn_core_num: int | None = None
+    is_attn_multistream: bool = False
+    is_ffn_multistream: bool = False
     compute_gate_on_attention: bool = False
     quant_mode: int = 0
 
@@ -117,6 +128,14 @@ class CAMP2PExtraInfo(ConnectorExtraInfo):
                 raw.get("ffn_core_num"),
                 field_name="ffn_core_num",
             ),
+            is_attn_multistream=coerce_extra_bool(
+                raw.get("is_attn_multistream", False),
+                field_name="is_attn_multistream",
+            ),
+            is_ffn_multistream=coerce_extra_bool(
+                raw.get("is_ffn_multistream", False),
+                field_name="is_ffn_multistream",
+            ),
             compute_gate_on_attention=coerce_extra_bool(
                 raw.get("compute_gate_on_attention", False),
                 field_name="compute_gate_on_attention",
@@ -145,6 +164,8 @@ class CAMP2PExtraInfo(ConnectorExtraInfo):
     def to_mapping(self) -> dict[str, Any]:
         result: dict[str, Any] = {
             "core_num": self.core_num,
+            "is_attn_multistream": self.is_attn_multistream,
+            "is_ffn_multistream": self.is_ffn_multistream,
             "compute_gate_on_attention": self.compute_gate_on_attention,
             "quant_mode": self.quant_mode,
         }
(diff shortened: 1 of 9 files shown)


yujuancao07:
multistream:
image
sigle stream:
image

hsliuustc0106 (commented):
Consolidated review of `ff5e15f1` (static review; lint + cpu-tests 3.10-3.13 + DCO green). The multistream schedule itself is coherent — layer-0 e2a stays on the compute stream, layers 1+ ping-pong via per-stage cached events, the final drain-wait is present, and AFD-off behavior is clean (flags default off, no streams/events created, `nullcontext()` fallback). One P1 and five P2s:

[P1] Gate the new per-step DP barrier on the multistream flags — afd_plugin/v1/worker/npu/attention_model_runner.py:163
The `dist.barrier(group=get_dp_group().cpu_group)` runs on every CAMP2P attention step regardless of `is_attn_multistream`/`is_ffn_multistream`, so existing multistream-OFF deployments take a host-blocking gloo barrier per step: perf-affecting with no A/B evidence and no mention in the body. It also changes failure semantics — a rank that raises before reaching the next barrier leaves peers blocked until process-group timeout, converting a loud local error into a cluster-wide stall. Gate it on the switches (or a dedicated flag) with evidence, or split it into its own PR.

[P2] `is_attn_multistream` silently no-ops without `use_ubatching` — afd_plugin/connectors/npu/camp2p.py:881
The `afd_*` forward-context attributes the ops discover are set only inside `AscendUBatchWrapper`, which is installed only under `use_ubatching`; with the flag on and ubatching off, a2e stays on the compute stream with no error. Validate the dependency in `fail_if_unsupported_npu_afd_features`, and set the attributes at one choke point rather than `getattr` defaults (which also hides upstream renames).

[P2] Declare the multistream × ACL-graph combination — no feature_validation entry at head
Under the standard CAM FULL_DECODE_ONLY config the wrapper installs during capture while the op impls do host-side `comm_event.record/wait` under an active capture; whether torch_npu permits event ops during ACL capture — and whether the stream switch survives replay — is asserted nowhere. Encode the combination as unsupported until proven, or attach device evidence per docs/npu/TESTING.md.

[P2] Pass multistream config into `AscendUBatchWrapper` instead of re-parsing it — afd_plugin/v1/worker/npu/npu_ubatch_wrapper.py:31
The wrapper branches on `"CAMP2pAFDConnector"` and calls `parse_connector_extra_info` itself; the attention runner already owns the resolved `extra_info`. Pass the flag in as a constructor argument — second parse means a second place to fail and drift.

[P2] Document the four new user-facing keys — docs untouched
`is_attn_multistream`/`is_ffn_multistream` and the `CAMP2PExtraInfo` handshake fields are absent from `docs/npu/CAM_P2P_CONNECTOR_USER_GUIDE.md` and `docs/design/module/connector_contracts.md`.

[P2] CI green proves parsing and the helper, not the orchestration — tests/unit/v1/worker/test_npu_runtime.py:220
The FFN dispatch tests — including the good event-ordering trace test — are `importorskip("torch_npu")` and skip in CI; only `test_camp2p_connector.py` parsing and the 32-line helper actually run. The trace test only needs fake events and the fake forward context; running it on CPU via the file's existing fake-module harness would close the gap.

Freshness: this branch is months old — main has since rewritten the surrounding code (attention runner ±561 lines, wrapper, FFN runner, #320/#282/MRV2/ubatching refactors). `merge-tree` shows no textual conflicts, but the semantic assumptions need re-verification on the rebased result, and a perf feature needs its NPU A/B numbers before merge.

PR #174, opened 2026-07-29
Fix pure-text AFD model config identity

### LINKED ISSUE #172: [Bug]: PR #144 breaks pure-text model config identity and causes Async CAM TP2 FlashComm1 RoPE failure
## Current environment

```text
Python: 3.11.14
OS: openEuler 22.03, Linux aarch64
Hardware: Ascend 910C / Atlas A3, 4 NPUs
Model: DeepSeekV2-Lite
afd-plugin: main
vLLM: 0.19.1
vLLM-Ascend: 0.19.1rc1
torch: 2.9.0
torch-npu: 2.9.0
CANN / NNAL: 8.5.1
Execution mode: eager
Connector: CAMAsyncAFDConnector
Topology:
  Attention: DP1 TP2
  FFN: DP2 TP1 EP2
Async MoE ubatching: disabled
```

## Reproduction

Run the existing Async CAM NPU smoke test:

```bash
cd /path/to/afd-plugin

ASCEND_RT_VISIBLE_DEVICES=0,1,2,3 \
AFD_NPU_E2E_MODEL=/path/to/DeepSeek-V2-Lite \
AFD_NPU_ASYNC_CAM_E2E_DEVICES=0,1,2,3 \
AFD_NPU_E2E_VLLM_BIN=vllm \
python -m pytest -vv -s \
  tests/e2e/models/deepseek_v2_lite/test_async_cam_npu.py::test_deepseek_v2_lite_async_cam_attn_dp1tp2_ffn_dp2ep2_smoke
```

The test itself enables `VLLM_ASCEND_ENABLE_FLASHCOMM1=1`. The E2E runner
retains it for the Attention TP2 process and removes it from the FFN TP1
process.

The generated connector configuration includes:

```json
{
  "dynamicQuant": 0,
  "async_moe_ubatching": false,
  "async_moe_num_ubatches": 2,
  "async_moe_split": "request",
  "attn_ranks_per_dp": 2
}
```

## Regression range

The same Async CAM smoke test was run against the commits immediately before
and after PR #144:

| afd-plugin commit | Result |
| --- | --- |
| `220c7626de5195bdcf24271659f0c1519a71f2a3` | Pass |
| `355a1cda4a1f170b0fd5db0029eba8c521151241` | Fail with `npu_interleave_rope` batch-size mismatch |

Commit `355a1cda4a1f170b0fd5db0029eba8c521151241` is:

```text
Merge pull request #144 from vllm-project/codex/fix-issue141-model-registration
Preserve native model registrations for non-AFD workers
```

Its first parent is the passing commit:

```text
220c7626de5195bdcf24271659f0c1519a71f2a3
```

Its second parent is:

```text
1ba33ef9bff816b2f4339ff587a2107a53be343b
```

The failure therefore starts when #144 is merged. This provides a narrow and
reproducible regression boundary and strongly indicates that #144 introduced
the issue.

## Expected behavior

The existing Async CAM smoke test should complete one request successfully.

For a pure-text DeepSeekV2-Lite model, the AFD-specific model configuration
must preserve the original pure-text relationship:

```python
model_config.hf_text_config is model_config.hf_config
```

With FlashComm1 sequence parallelism enabled, MLA should gather the TP-local
query/KV token shards before applying RoPE so that `x`, `cos`, and `sin` have
matching batch dimensions.

## Actual behavior and logs

The Attention TP1 worker fails in the first dense layer during MLA RoPE:

```text
[attention] (Worker_TP1_EP1)
  File "afd_plugin/model_executor/models/deepseek_v2.py", line 497, in forward_with_afd
    return self.forward_with_afd_v2(...)
  File "afd_plugin/model_executor/models/deepseek_v2.py", line 560, in forward_with_afd_v2
    return deepseek_v2_async_cam_forward.run_attention_gate_afd_forward(...)
  File "afd_plugin/model_executor/models/npu/deepseek_v2_async_cam_forward.py", line 63
    hidden_states, residual = layer(...)
  File "vllm/model_executor/models/deepseek_v2.py", line 1005, in forward
    return self.mla_attn(positions, hidden_states, llama_4_scaling)
  File "vllm_ascend/ops/mla.py", line 173, in forward
    torch.ops.vllm.mla_forward(hidden_states, need_gather_q_kv, output, self.prefix)
  File "vllm_ascend/attention/mla_v1.py", line 1542, in mla_preprocess_prefill
    prefill_q_pe = self.rope_single(prefill_q_pe, cos, sin)
  File "vllm_ascend/attention/mla_v1.py", line 1346, in rope_single
    x = torch_npu.npu_interleave_rope(x, cos, sin)

RuntimeError: call aclnnInterleaveRope failed, error code is 561002

EZ9999: cos or sin batchSize must same as x batchSize.
[FUNC:GetShapeAttrsInfo]
[FILE:interleave_rope_tiling.cpp]
[LINE:132]

InterleaveRope do tiling failed.
```

The same failure occurs on the Attention ranks because the FlashComm1
sequence-parallel tensor has a TP-local token dimension while the RoPE metadata
still describes the full token batch.

## Suspected root cause

PR #144 changed AFD model registration so native DeepSeek registrations would
remain available for non-AFD workers. As part of that change, it introduced
`get_afd_model_config()` and began calling it from the AFD Attention and FFN
workers.

The new helper performs two shallow copies:

```python
afd_model_config = copy(model_config)
afd_model_config.hf_config = copy(model_config.hf_config)
afd_model_config.hf_config.architectures = [f"AFD{model_arch}"]
```

For a pure-text DeepSeek model, vLLM initially uses the same object for
`hf_config` and `hf_text_config`:

```python
model_config.hf_text_config is model_config.hf_config
```

After the #144 conversion, only `hf_config` is replaced. `hf_text_config`
continues to reference the original object:

```python
afd_model_config.hf_text_config is not afd_model_config.hf_config
```

This breaks an object-identity invariant used by vLLM/vLLM-Ascend to
distinguish pure-text model configurations from configurations containing a
separate nested text model.

The suspected downstream sequence is:

1. #144 converts the DeepSeek model config to the AFD architecture.
2. The conversion breaks the `hf_text_config is hf_config` identity.
3. vLLM-Ascend treats the pure-text DeepSeekV2-Lite config as a VL/multimodal
   config.
4. The first MLA layer selects `need_gather_q_kv=False`.
5. FlashComm1 has already sequence-sharded the hidden states across TP2, so the
   RoPE input `x` contains only the TP-local token rows.
6. The Attention metadata and its `cos`/`sin` tensors still describe the full
   token batch.
7. `npu_interleave_rope` receives different batch dimensions and fails with
   `cos or sin batchSize must same as x batchSize`.

This explanation is strongly supported by the exact regression boundary and
the code introduced by #144, but the downstream vLLM-Ascend branch should still
be confirmed with runtime instrumentation of:

```text
hf_text_config is hf_config
need_gather_q_kv
x.shape
cos.shape
sin.shape
```

This issue is independent of AFD-managed async MoE ubatching because the
existing reproduction explicitly sets `async_moe_ubatching=false`.

## Proposed fix

Preserve the pure-text object identity while leaving real nested text
configurations unchanged:

```python
afd_model_config = copy(model_config)
afd_hf_config = copy(model_config.hf_config)
afd_hf_config.architectures = [f"AFD{model_arch}"]
afd_model_config.hf_config = afd_hf_config

if model_config.hf_text_config is model_config.hf_config:
    afd_model_config.hf_text_config = afd_hf_config
```

Add unit coverage for both cases:

1. Pure-text configuration: `hf_text_config is hf_config` must remain true
   after conversion.
2. Nested/multimodal configuration: a distinct `hf_text_config` must remain
   distinct and unchanged.

The existing Async CAM NPU smoke test should then provide E2E regression
coverage for the TP2 + FlashComm1 path.

## AFD configuration

```yaml
attention:
  data_parallel_size: 1
  tensor_parallel_size: 2
  afd:
    role: attention
    connector: CAMAsyncAFDConnector
    async: true
    compute_gate_on_attention: true
    num_attention_ranks: 2
    num_ffn_ranks: 2
    connector_extra_config:
      dynamicQuant: 0
      async_moe_ubatching: false
      async_moe_num_ubatches: 2
      async_moe_split: request
      attn_ranks_per_dp: 2

ffn:
  data_parallel_size: 2
  tensor_parallel_size: 1
  enable_expert_parallel: true
  afd:
    role: ffn
    connector: CAMAsyncAFDConnector
    async: true
    compute_gate_on_attention: true
    num_attention_ranks: 2
    num_ffn_ranks: 2
    connector_extra_config:
      dynamicQuant: 0
      async_moe_ubatching: false
      async_moe_num_ubatches: 2
      async_moe_split: request
      attn_ranks_per_dp: 2
```

## Before submitting

- [x] This reproduces with vLLM v0.19.1 and vLLM-Ascend v0.19.1rc1.

ShwStone:
## Purpose

Fix the model configuration regression introduced by #144 and reported in
#172.

`get_afd_model_config()` copied and replaced `hf_config` without updating
`hf_text_config`. For pure-text models, this broke the original identity:

```python
model_config.hf_text_config is model_config.hf_config
```

vLLM-Ascend consequently treated DeepSeekV2-Lite as a model with a distinct
nested text configuration. Under Attention TP2 with FlashComm1/SP, MLA skipped
the required query/KV gather and passed TP-local `x` together with full-batch
RoPE `cos`/`sin`, causing `npu_interleave_rope` to fail.

The regression was confirmed across #144:

- `220c7626de5195bdcf24271659f0c1519a71f2a3`: passes
- `355a1cda4a1f170b0fd5db0029eba8c521151241`: fails

## Issue

- Related issue(s): #172
- Closing keyword, only if fully resolved: Closes #172

## Scope

- In scope:
  - Preserve `hf_config`/`hf_text_config` identity when converting a pure-text
    model config to its AFD architecture.
  - Preserve a genuinely distinct nested `hf_text_config` unchanged.
  - Add unit coverage for both configuration layouts.
- Out of scope:
  - Changes to vLLM or vLLM-Ascend.
  - Changes to FlashComm1/SP scheduling or MLA implementation.
  - Changes to Async CAM communication or MoE ubatching.
  - Changes to multimodal model configuration semantics.

## Implementation Notes

`get_afd_model_config()` now creates one private `hf_config` copy, updates its
architecture, and assigns it to the copied `ModelConfig`.

When the source is a pure-text configuration:

```python
model_config.hf_text_config is model_config.hf_config
```

the copied model config points both fields to the new private `hf_config`.
When the source has a genuinely distinct nested `hf_text_config`, that object
remains distinct and unchanged.

This is a plugin-owned model configuration fix. It does not patch or modify
vLLM/vLLM-Ascend source code.

## Test Plan

Focused unit coverage:

```bash
python -m pytest -vv \
  tests/unit/package/test_package.py \
  -k "afd_model_config"
```

Ascend NPU Async CAM regression coverage:

```bash
ASCEND_RT_VISIBLE_DEVICES=0,1,2,3 \
AFD_NPU_E2E_MODEL=/path/to/DeepSeek-V2-Lite \
AFD_NPU_ASYNC_CAM_E2E_DEVICES=0,1,2,3 \
AFD_NPU_E2E_VLLM_BIN=vllm \
AFD_NPU_E2E_STARTUP_TIMEOUT=900 \
python -m pytest -vv -s \
  tests/e2e/models/deepseek_v2_lite/test_async_cam_npu.py::test_deepseek_v2_lite_async_cam_attn_dp1tp2_ffn_dp2ep2_smoke
```

## Test Result

Both targeted tests passed:

- Focused `afd_model_config` unit tests: **passed**
- Async CAM Attention DP1TP2 + FFN DP2EP2 NPU E2E smoke test: **passed**

The E2E result confirms that the existing main-branch Async CAM test can run
with Attention TP2 and FlashComm1 without the previous RoPE batch-dimension
failure.

## Docs Impact

- Files updated: none
- If none, reason: this restores the original pure-text `ModelConfig` invariant
  and does not introduce a new public configuration or runtime option. The
  regression and reproduction are documented in #172.

---

<details>

- [x] Scope is bounded.

</details>

### DIFF
--- a/afd_plugin/model_executor/models/model_utils.py
+++ b/afd_plugin/model_executor/models/model_utils.py
@@ -19,7 +19,14 @@ def get_afd_model_config(model_config: ModelConfig) -> ModelConfig:
     for model_arch in model_config.hf_config.architectures:
         if model_arch in _DEEPSEEK_MODEL_REGISTRATIONS:
             afd_model_config = copy(model_config)
-            afd_model_config.hf_config = copy(model_config.hf_config)
-            afd_model_config.hf_config.architectures = [f"AFD{model_arch}"]
+            afd_hf_config = copy(model_config.hf_config)
+            afd_hf_config.architectures = [f"AFD{model_arch}"]
+            afd_model_config.hf_config = afd_hf_config
+
+            # Pure-text ModelConfig uses the same object for hf_config and
+            # hf_text_config. Preserve that identity: vLLM Ascend uses it to
+            # distinguish text models from multimodal models.
+            if model_config.hf_text_config is model_config.hf_config:
+                afd_model_config.hf_text_config = afd_hf_config
             return afd_model_config
     return model_config
--- a/afd_plugin/model_executor/models/model_utils.py
+++ b/afd_plugin/model_executor/models/model_utils.py
@@ -4,7 +4,7 @@
 
 from __future__ import annotations
 
-from copy import copy
+from copy import deepcopy
 from typing import TYPE_CHECKING
 
 from afd_plugin import _DEEPSEEK_MODEL_REGISTRATIONS
@@ -18,15 +18,11 @@ def get_afd_model_config(model_config: ModelConfig) -> ModelConfig:
 
     for model_arch in model_config.hf_config.architectures:
         if model_arch in _DEEPSEEK_MODEL_REGISTRATIONS:
-            afd_model_config = copy(model_config)
-            afd_hf_config = copy(model_config.hf_config)
-            afd_hf_config.architectures = [f"AFD{model_arch}"]
-            afd_model_config.hf_config = afd_hf_config
-
-            # Pure-text ModelConfig uses the same object for hf_config and
-            # hf_text_config. Preserve that identity: vLLM Ascend uses it to
-            # distinguish text models from multimodal models.
-            if model_config.hf_text_config is model_config.hf_config:
-                afd_model_config.hf_text_config = afd_hf_config
+            # deepcopy preserves aliasing within the copied object graph, so
+            # the pure-text identity hf_text_config is hf_config is retained
+            # automatically. vLLM Ascend uses that identity to distinguish
+            # text models from multimodal models.
+            afd_model_config = deepcopy(model_config)
+            afd_model_config.hf_config.architectures = [f"AFD{model_arch}"]
             return afd_model_config
     return model_config
--- a/tests/unit/package/test_package.py
+++ b/tests/unit/package/test_package.py
@@ -77,8 +77,10 @@ def test_afd_model_config_preserves_nested_text_config():
 
     afd_model_config = get_afd_model_config(model_config)
 
+    # deepcopy privatizes the whole graph; a genuinely distinct nested
+    # hf_text_config stays distinct from hf_config.
     assert afd_model_config.hf_config is not model_config.hf_config
-    assert afd_model_config.hf_text_config is hf_text_config
+    assert afd_model_config.hf_text_config is not hf_text_config
     assert afd_model_config.hf_text_config is not afd_model_config.hf_config
 
 
--- a/tests/unit/package/test_package.py
+++ b/tests/unit/package/test_package.py
@@ -50,16 +50,36 @@ def test_afd_model_config_uses_private_architecture_copy():
     pytest.importorskip("vllm")
     from afd_plugin.model_executor.models.model_utils import get_afd_model_config
 
+    hf_config = SimpleNamespace(architectures=["DeepseekV2ForCausalLM"])
     model_config = SimpleNamespace(
-        hf_config=SimpleNamespace(architectures=["DeepseekV2ForCausalLM"]),
+        hf_config=hf_config,
+        hf_text_config=hf_config,
     )
 
     afd_model_config = get_afd_model_config(model_config)
 
     assert afd_model_config is not model_config
     assert afd_model_config.hf_config is not model_config.hf_config
+    assert afd_model_config.hf_text_config is afd_model_config.hf_config
     assert afd_model_config.hf_config.architectures == ["AFDDeepseekV2ForCausalLM"]
     assert model_config.hf_config.architectures == ["DeepseekV2ForCausalLM"]
+
+
+def test_afd_model_config_preserves_nested_text_config():
+    pytest.importorskip("vllm")
+    from afd_plugin.model_executor.models.model_utils import get_afd_model_config
+
+    hf_text_config = SimpleNamespace()
+    model_config = SimpleNamespace(
+        hf_config=SimpleNamespace(architectures=["DeepseekV2ForCausalLM"]),
+        hf_text_config=hf_text_config,
+    )
+
+    afd_model_config = get_afd_model_config(model_config)
+
+    assert afd_model_config.hf_config is not model_config.hf_config
+    assert afd_model_config.hf_text_config is hf_text_config
+    assert afd_model_config.hf_text_config is not afd_model_config.hf_config
 
 
 def test_entry_point_is_registered():


jiangkuaixue123:
image
Fix DCO

jiangkuaixue123 (commented):
Would it be safer to use `deepcopy(model_config)` here and then update `afd_model_config.hf_config.architectures`? `deepcopy` preserves aliasing within the copied object graph, so the pure-text invariant `hf_text_config is hf_config` would be retained automatically, and it would reduce the risk of missing other identity relationships if `ModelConfig` evolves.

The semantic difference is that a genuinely distinct/nested `hf_text_config` would also become a private copy, whereas this patch intentionally keeps it shared with the source config. Is that sharing required? If not, `deepcopy` may provide stronger isolation and a simpler implementation. Otherwise, could we document why only `hf_config` should be copied so this choice is explicit?

ShwStone:
I confused GPG signing with DCO; I always thought signing was enough to pass DCO. I've already recommitted.

ShwStone:
@jiangkuaixue123 Agreed — the deepcopy approach is cleaner and preserves the aliasing invariants automatically. I've switched the implementation to `deepcopy(model_config)` in 38df180 and validated it on real hardware (DeepSeekV2-Lite, Attention TP2 with FlashComm1/SP): the `npu_interleave_rope` failure is gone and the model runs correctly. Tests updated accordingly. PTAL.

PR #178, opened 2026-07-29
bugfix:DBO ubatch NPUGraph capture passes per-stage metadata to the standard full-graph updater

### LINKED ISSUE #177: [Bug]: DBO ubatch NPUGraph capture passes per-stage metadata to the standard full-graph updater
### Current environment

```text
afd-plugin commit: 9d5bffc5c1fcca2157befbc8ca4820cb6990344a
vLLM version/checkout:0.19.1
NPU
```

### Reproduction

1. Start the FFN service and Attention service.
3. Enable DBO with `ubatch-size=2`.
4. Enable full decode NPUGraph capture.
5. Use DeepSeek-V2-Lite and wait for graph capture.

The failure occurs during service initialization, before the API becomes ready.

### Expected behavior

Ascend's standard full-graph parameter updater should only run for normal, non-ubatch contexts.

For AFD ubatch contexts, graph parameters are handled by the AFD wrapper's per-stage capture/replay path. The outer per-stage metadata list should not be passed to the standard single-batch updater.

### Actual behavior and logs

```text
When AFD runs with DBO, `ubatch-size > 1`, and full decode NPUGraph capture, the Attention side can fail during graph warm-up with:


AttributeError: 'list' object has no attribute 'keys'
```

### Before submitting

- [x] I confirmed whether this reproduces against vLLM v0.19.1 or explained why not.

yujuancao07:
## Purpose

Add NPU support for MLA native DBO with Ascend ACL FULL graph replay,
including dense MLA models such as DeepSeek-V2-Lite.

With DBO enabled, the two child ubatches capture independent MLA graph
parameters. The existing vLLM-Ascend process-global graph parameter registry
cannot represent both child graphs safely. This change creates one registry
per ubatch, merges the captured records before replay, and preserves the
required eNPU/non-eNPU update ordering.

## Issue

- Related issue(s): #177 
- Closing keyword, only if fully resolved: Closes #177 

## Scope

- In scope:
  - Support MLA DBO FULL graph capture and replay on NPU.
  - Maintain independent MLA graph parameter registries for two ubatches.
  - Share the aggregate FIA workspace between the two child registries.
  - Merge graph records in layer-major, ubatch-minor order before replay.
  - Enable the existing DeepSeek-V2-Lite NPU DBO E2E cases.

- Out of scope:
  - Changing the existing sparse MLA path.
  - Changes to the vLLM source checkout.
  - Changes to CAMAsync scheduling behavior.

### Focused unit and runner tests

```bash
python -m pytest -vv \
  tests/unit/compat/test_runtime.py \
  tests/unit/v1/worker/test_npu_mla_graph.py \
  tests/unit/v1/worker/test_npu_runtime.py \
  tests/e2e/test_runner.py
```

### DIFF
--- a/afd_plugin/compat/npu/feature_validation.py
+++ b/afd_plugin/compat/npu/feature_validation.py
@@ -56,11 +56,31 @@ def fail_if_unsupported_npu_afd_features(
             )
         extra_info.validate_supported()
 
-    if bool(vllm_config.parallel_config.use_ubatching) and (
-        int(vllm_config.parallel_config.num_ubatches) != 2
-    ):
+    uses_ubatching = bool(vllm_config.parallel_config.use_ubatching)
+    if uses_ubatching and int(vllm_config.parallel_config.num_ubatches) != 2:
         raise RuntimeError(
             "AFD NPU runtime supports exactly two ubatches when DBO is enabled",
+        )
+    model_config = vllm_config.model_config
+    # Match the pinned NPUModelRunner's sparse-attention backend selection.
+    uses_sparse_mla = hasattr(
+        model_config.hf_text_config,
+        "index_topk",
+    )
+    cudagraph_mode = vllm_config.compilation_config.cudagraph_mode
+    uses_mla_dbo_full_graph = (
+        uses_ubatching
+        and model_config.use_mla
+        and not uses_sparse_mla
+        and cudagraph_mode.has_full_cudagraphs()
+    )
+    if uses_mla_dbo_full_graph and vllm_config.speculative_config is not None:
+        raise RuntimeError(
+            "AFD NPU MLA DBO FULL graph does not support speculative decoding",
+        )
+    if uses_mla_dbo_full_graph and cudagraph_mode.name != "FULL_DECODE_ONLY":
+        raise RuntimeError(
+            "AFD NPU MLA DBO graph execution requires FULL_DECODE_ONLY",
         )
 
 
--- a/afd_plugin/compat/npu/runtime.py
+++ b/afd_plugin/compat/npu/runtime.py
@@ -28,9 +28,19 @@ def apply_afd_ascend_patches_if_needed() -> None:
     from afd_plugin.compat.patches.npu.ascend_platform import (
         apply_afd_ascend_dbo_config_patch,
     )
+    from afd_plugin.compat.patches.npu.mla_graph import (
+        apply_afd_mla_graph_patch,
+    )
 
-    if apply_afd_ascend_dbo_config_patch():
-        _PATCHES_APPLIED = True
+    if not apply_afd_ascend_dbo_config_patch():
+        raise RuntimeError(
+            "AFD NPU DBO config patch requires vLLM-Ascend NPUPlatform",
+        )
+    if not apply_afd_mla_graph_patch():
+        raise RuntimeError(
+            "AFD NPU MLA graph patch requires the vLLM-Ascend MLA resolver",
+        )
+    _PATCHES_APPLIED = True
 
 
 __all__ = [
--- a/afd_plugin/v1/worker/npu/mla_graph.py
+++ b/afd_plugin/v1/worker/npu/mla_graph.py
@@ -1,0 +1,127 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
+"""Graph-local MLA parameter helpers for Ascend ubatch execution."""
+
+from __future__ import annotations
+
+from collections.abc import Iterator
+from contextlib import contextmanager
+from typing import TYPE_CHECKING, TypeVar
+
+from afd_plugin.compat.patches.npu.mla_graph import (
+    AFD_MLA_GRAPH_PARAMS_KEY,
+)
+
+if TYPE_CHECKING:
+    import torch
+    from vllm.forward_context import ForwardContext
+    from vllm_ascend.compilation.acl_graph import GraphParams
+
+MetadataT = TypeVar("MetadataT")
+
+
+def new_mla_graph_params(
+    num_tokens: int,
+    workspace: torch.Tensor,
+) -> GraphParams:
+    """Create an empty MLA registry for one token shape and FIA workspace."""
+    # Delay the NPU-only import so this helper remains CPU-import-safe.
+    from vllm_ascend.compilation.acl_graph import GraphParams
+
+    return GraphParams(
+        events={num_tokens: []},
+        workspaces={num_tokens: workspace},
+        handles={num_tokens: []},
+        attn_params={num_tokens: []},
+    )
+
+
+def merge_mla_graph_params(
+    attn_metadata: list[dict[str, MetadataT]],
+    graph_params: tuple[GraphParams, GraphParams],
+    num_tokens: int,
+) -> tuple[dict[tuple[str, int], MetadataT], GraphParams]:
+    """Validate and merge two stage-local MLA registries for graph replay."""
+    if len(attn_metadata) != 2 or len(graph_params) != 2:
+        raise RuntimeError(
+            "MLA DBO FULL graph requires exactly two metadata stages; "
+            f"got metadata={len(attn_metadata)}, params={len(graph_params)}",
+        )
+
+    layer_keys = tuple(attn_metadata[0])
+    if tuple(attn_metadata[1]) != layer_keys:
+        raise RuntimeError("MLA DBO FULL graph layer order differs by stage")
+
+    expected_records = len(layer_keys)
+    for stage_index, params in enumerate(graph_params):
+        record_counts = (
+            len(params.events[num_tokens]),
+            len(params.handles[num_tokens]),
+            len(params.attn_params[num_tokens]),
+        )
+        if record_counts != (
+            expected_records,
+            expected_records,
+            expected_records,
+        ):
+            raise RuntimeError(
+                "MLA DBO FULL graph record count mismatch for "
+                f"stage {stage_index}: metadata={expected_records}, "
+                f"events={record_counts[0]}, handles={record_counts[1]}, "
+                f"params={record_counts[2]}",
+            )
+
+    workspace = graph_params[0].workspaces[num_tokens]
+    if graph_params[1].workspaces[num_tokens] is not workspace:
+        raise RuntimeError(
+            "MLA DBO FULL graph requires one shared FIA workspace",
+        )
+
+    merged_metadata: dict[tuple[str, int], MetadataT] = {}
+    merged = new_mla_graph_params(num_tokens, workspace)
+
+    # Preserve the updater's layer-major, stage-minor record order.
+    for layer_index, layer_key in enumerate(layer_keys):
+        for stage_index in range(2):
+            params = graph_params[stage_index]
+            merged_key = (layer_key, stage_index)
+            merged_metadata[merged_key] = attn_metadata[stage_index][layer_key]
+            merged.events[num_tokens].append(
+                params.events[num_tokens][layer_index],
+            )
+            merged.handles[num_tokens].append(
+                params.handles[num_tokens][layer_index],
+            )
+            merged.attn_params[num_tokens].append(
+                params.attn_params[num_tokens][layer_index],
+            )
+
+    return merged_metadata, merged
+
+
+@contextmanager
+def override_mla_graph_params(
+    forward_context: ForwardContext,
+    attn_metadata: dict[tuple[str, int], MetadataT],
+    graph_params: GraphParams,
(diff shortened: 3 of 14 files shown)


jiangkuaixue123:
Thanks for the investigation. The root cause is not the shape of `attn_metadata`, but that MLA on Ascend has not yet been adapted to DBO + FULL graph.

In the previous AFD v0.13-based implementation, MLA had a dedicated mechanism to update the full-graph parameters. That mechanism should be ported to the current implementation. Skipping `_update_full_graph_params_if_needed()` for ubatches only avoids the immediate exception; it does not provide the MLA full-graph update semantics that are required.

This is why DeepSeek-V2-Lite cannot currently run with DBO + FULL graph. DeepSeek-V3.2 can run with the same configuration because it uses DSA rather than MLA, so it does not hit this MLA-specific gap.

Therefore, we will not accept the current workaround as the fix. Please port the previous MLA-specific full-graph update mechanism and validate DeepSeek-V2-Lite with DBO + FULL graph.

afd_plugin/compat/npu/feature_validation.py
```diff
@@ -56,12 +56,32 @@ def fail_if_unsupported_npu_afd_features(
...
+    model_config = vllm_config.model_config
+    # Match the pinned NPUModelRunner's sparse-attention backend selection.
+    uses_sparse_mla = hasattr(
+        model_config.hf_text_config,
+        "index_topk",
+    )
+    supports_mla_dbo_full_graph = (
+        model_config.use_mla
+        and not uses_sparse_mla
+        and vllm_config.speculative_config is None
```
jiangkuaixue123: [P1] Reject or support speculative MLA before enabling DBO FULL graph

With `use_mla=True`, DBO enabled, a FULL graph mode, and a non-`None` `speculative_config`, this condition makes `supports_mla_dbo_full_graph` false, so the configuration is accepted. `_install_ascend_ubatch_wrapper()` then also sets `mla_full_graph_enabled=False`; once the DBO threshold creates two ubatches, `_model_forward()` leaves the update to the upstream MLA path. The parent `forward_context.attn_metadata` is a `list[dict]` from `_build_attention_metadata_with_ubatches()`, while vLLM-Ascend 0.19.1 `MLAImpl.update_graph_params()` takes the main-model branch and calls `attn_metadata.keys()`, so replay fails instead of using the new per-stage registries. Either fail fast for speculative MLA + DBO FULL graph, or extend the per-ubatch registry/update path to this case; add a regression that reaches `_model_forward()` rather than asserting this configuration is allowed.

jiangkuaixue123 (changes requested):
Reviewed commit `602ad8a6e01d18ac7355806099a9e8e8a028373d` with InferMatrixCopilot Direct.

Requesting changes for the inline P1 below.

Validation performed against this snapshot:
- `compileall` passed for every changed Python file.
- `tests/unit/v1/worker/test_npu_mla_graph.py`: 25 passed on CPython 3.12 / macOS arm64 with PyTorch.
- The broader NPU runtime test set could not be collected on this host because the pinned `vllm` / `vllm_ascend` / `torch_npu` runtime is unavailable; GitHub currently exposes no combined CI statuses for this SHA.

Cleanup: please remove the unrelated `docs/assets/WeChat.jpg` binary replacement (272,151 → 412,617 bytes). It is outside the MLA graph fix and adds repository bloat/review noise.

afd_plugin/compat/npu/runtime.py
```diff
@@ -28,8 +28,12 @@ def apply_afd_ascend_patches_if_needed() -> None:
     from afd_plugin.compat.patches.npu.ascend_platform import (
         apply_afd_ascend_dbo_config_patch,
     )
+    from afd_plugin.compat.patches.npu.mla_graph import (
+        apply_afd_mla_graph_patch,
+    )
 
     apply_afd_ascend_dbo_config_patch()
+    apply_afd_mla_graph_patch()
```
jiangkuaixue123: [P1] Do not silently mark a required MLA patch as applied

`apply_afd_mla_graph_patch()` returns `None` both when installation succeeds and when its broad import `except Exception` path silently returns. The next line then sets `_PATCHES_APPLIED = True`. If installation fails, the wrapper can still claim ownership of MLA full-graph updates, but the child forward contexts will continue using the upstream process-global registry instead of their per-ubatch registries; the guard also prevents a later retry. Please make installation status explicit and fail fast when MLA DBO FULL graph requires this patch (while preserving the optional no-Ascend path). Add a regression covering a failed or unavailable resolver installation.

afd_plugin/v1/worker/npu/npu_ubatch_wrapper.py
```diff
@@ -59,21 +83,29 @@ def __init__(
         vllm_config: VllmConfig,
         runtime_mode: CUDAGraphMode,
         device: torch.device,
+        *,
+        mla_full_graph_enabled: bool = False,
```
jiangkuaixue123, line 87: [P2] Keep MLA graph policy out of the generic ubatch wrapper

These constructor additions are the start of a much broader ownership inversion: `AscendUBatchWrapper` now decides MLA support, imports and constructs `GraphParams`, selects the aggregate FIA workspace, merges layer records, calls a bound private runner updater, and enforces speculative/PCP policy. The duplicated policy has already diverged from feature validation in the speculative-MLA case. Please move this behavior into an MLA-specific coordinator/adapter owned by the NPU attention runner, and let the wrapper interact with it through opaque capture state and small capture/replay hooks. The generic wrapper should continue to own splitting, scheduling, capture, replay, and stream fencing—not MLA parameter semantics.

jiangkuaixue123 (changes requested):
The per-ubatch registry direction is reasonable, but I do not think the current ownership boundary is ready to merge.

The existing P1 about speculative MLA remains unresolved: validation accepts speculative MLA + DBO + FULL graph while the per-ubatch registry path is disabled, so execution falls back to the upstream single-batch updater with `list[dict]` metadata.

More broadly, the generic Ascend ubatch wrapper now owns MLA feature policy, `GraphParams` internals, FIA workspace selection, layer-record merging, a bound runner updater, and eNPU/non-eNPU update ordering. Please extract these MLA-specific responsibilities into a dedicated graph coordinator/adapter. The wrapper should retain only generic ubatch graph orchestration: an opaque capture state plus small capture/replay lifecycle hooks. Completely avoiding wrapper changes is not necessary, but the wrapper should not become the owner of MLA graph semantics.

Before approval, please also provide real NPU E2E evidence for DeepSeek-V2-Lite with DBO + `FULL_DECODE_ONLY`. The stub-heavy unit tests cannot validate the external-event, two-thread forward-context switching, and ACL graph update/replay ordering exercised by this change. If both eNPU and non-eNPU paths are supported, please validate both orderings.

Correction to my earlier review: the current PR diff does not contain a `docs/assets/WeChat.jpg` change, so please disregard that cleanup note.

jiangkuaixue123:
Please append fully e2e test on NPU. And Dsv2-lite profiling.

jiangkuaixue123:
fix
image

afd_plugin/v1/worker/npu/mla_graph.py
```diff
@@ -0,0 +1,122 @@
...
+
+if TYPE_CHECKING:
+    import torch
+    from vllm.forward_context import ForwardContext
+    from vllm_ascend.compilation.acl_graph import GraphParams
+
+MetadataT = TypeVar("MetadataT")
+
+
+def new_mla_graph_params(
```
jiangkuaixue123, line 23: Add comment and docstring

yujuancao07: fixed

yujuancao07:
> Please append fully e2e test on NPU. And Dsv2-lite profiling.

e2e test：
image
unit test:
image
profile:
image

afd_plugin/compat/npu/runtime.py
```diff
@@ -28,9 +28,17 @@ def apply_afd_ascend_patches_if_needed() -> None:
     from afd_plugin.compat.patches.npu.ascend_platform import (
         apply_afd_ascend_dbo_config_patch,
     )
+    from afd_plugin.compat.patches.npu.mla_graph import (
+        apply_afd_mla_graph_patch,
+    )
 
-    if apply_afd_ascend_dbo_config_patch():
-        _PATCHES_APPLIED = True
+    if not apply_afd_ascend_dbo_config_patch():
```
jiangkuaixue123, line 35: apply_afd_ascend_dbo_config_patch apply_afd_mla_graph_patch both raise RuntimeError

afd_plugin/v1/worker/npu/npu_ubatch_wrapper.py
```diff
@@ -154,15 +218,31 @@ def __call__(self, *args, **kwargs):
...
+            if mla_full_graph_active:
+                self._replay_mla_graph(
+                    cudagraph_metadata,
+                    forward_context,
+                    stage_num_tokens[0],
+                    positions,
+                )
+            else:
+                if not self.enable_enpu:
+                    torch.npu.current_stream().synchronize()
```
jiangkuaixue123: why here need synchronize

afd_plugin/v1/worker/npu/npu_ubatch_wrapper.py
```diff
@@ -59,21 +83,29 @@ def __init__(
...
                 runtime_mode=runtime_mode,
+                enable_enpu=enable_enpu,
             )
         self.device = device
+        self.mla_full_graph_enabled = mla_full_graph_enabled
+        self.full_graph_params_updater = full_graph_params_updater
+        self.enable_enpu = enable_enpu
 
     @property
     def graph_pool(self):
```
jiangkuaixue123, line 110: ENPU is not supported by this wrapper yet. Could we assert `not enable_enpu` here to make that limitation explicit, and then remove all of the downstream `self.enable_enpu` conditional branches?

jiangkuaixue123 (commented):
Please see the inline comment.

afd_plugin/v1/worker/npu/npu_ubatch_wrapper.py
```diff
@@ -96,24 +128,51 @@ def __getattr__(self, key: str):
...
+        return True
+
     def __call__(self, *args, **kwargs):
         forward_context = get_forward_context()
         batch_descriptor = forward_context.batch_descriptor
         ubatch_slices = forward_context.ubatch_slices
         cudagraph_runtime_mode = forward_context.cudagraph_runtime_mode
 
         if ubatch_slices is None:
-            if cudagraph_runtime_mode is CUDAGraphMode.FULL:
```
jiangkuaixue123, line 106: Explain why remove this.

afd_plugin/v1/worker/npu/npu_ubatch_wrapper.py
```diff
@@ -179,6 +259,75 @@ def __call__(self, *args, **kwargs):
...
+                self.full_graph_params_updater(
+                    forward_context,
+                    num_tokens,
+                    positions,
+                )
+
+        if self.enable_enpu:
+            update_params()
+        else:
+            torch.npu.current_stream().synchronize()
```
jiangkuaixue123: why here need synchronize

afd_plugin/v1/worker/npu/npu_ubatch_wrapper.py
yujuancao07: [https://github.com/vllm-project/vllm-ascend/blob/v0.19.1rc1/vllm_ascend/compilation/acl_graph.py#L199-L212](url)
This synchronization ensures that the previous iteration has finished consuming them before they are reused by the current iteration. Otherwise, under async CPU scheduling, the updater may modify them while the previous graph is still running.

afd_plugin/v1/worker/npu/npu_ubatch_wrapper.py
yujuancao07: [https://github.com/vllm-project/vllm-ascend/blob/v0.19.1rc1/vllm_ascend/compilation/acl_graph.py#L199-L212](url)
This synchronization ensures that the previous iteration has finished consuming them before they are reused by the current iteration. Otherwise, under async CPU scheduling, the updater may modify them while the previous graph is still running.

afd_plugin/v1/worker/npu/npu_ubatch_wrapper.py
yujuancao07: The removed fallback was tied to the old dict[int, ...] ubatch graph cache
when ubatch_slices is None, this is a standard single-batch FULL-graph execution and should be handled by the inner ACLGraphWrapper
The removal keeps the single-batch and ubatch graph caches independent. test_single_full_graph_uses_inner_wrapper_when_ubatch_total_matches covers this behavior.

Issue #169, opened 2026-07-28
[RFC]: Add DeepSeek V4 support on GPU and NPU for vLLM 0.26.0


jiangkuaixue123:
## Implementation tracking (updated 2026-08-14)

Legend: :white_check_mark: merged/done - :hourglass_flowing_sand: open/in progress - :x: TODO (no PR yet)

| Status | Workstream | Ref | Note |
|---|---|---|---|
| :white_check_mark: | GPU vLLM 0.26.0 prerequisite | #167 | Upgrade gate completed |
| :white_check_mark: | NPU vLLM/vLLM-Ascend prerequisite | #168 | Upgrade gate completed |
| :hourglass_flowing_sand: | GPU `P2pNcclAFDConnector` DeepSeek V4 support | #191 | Model adaptation and GPU connector integration in review |
| :x: | NPU `CAMP2pAFDConnector` DeepSeek V4 support | - | No implementation PR yet |
| :x: | NPU `CAMAsyncAFDConnector` DeepSeek V4 support | - | No implementation PR yet |

## Motivation

DeepSeek V4 is a committed model-support workstream in the project roadmap
(#155). vLLM `0.26.0` provides a native DeepSeek V4 implementation with a
substantially different model and attention lifecycle from the existing
DeepSeek V2/V3 adapter, including dedicated model construction, sparse MLA
state, indexer/compressor behavior, cache layout, MoE routing, quantization,
and weight loading.

AFD support must preserve those native semantics while inserting the
Attention/FFN stage boundary and connector communication. It must work on the
GPU `P2pNcclAFDConnector` path and both NPU connector paths:
`CAMP2pAFDConnector` and `CAMAsyncAFDConnector`.

Support also requires reproducible deployment recipes. A launch-only model
registration or smoke test is not sufficient to claim DeepSeek V4 support.

## Goals

- Support DeepSeek V4 with the exact vLLM `0.26.0` compatibility baseline.
- Support the GPU `P2pNcclAFDConnector` path after the GPU `0.26.0` upgrade in
  #167 is initially ready.
- Support both NPU connector paths after the NPU upgrade in #168 is complete:
  - synchronous `CAMP2pAFDConnector`;
  - asynchronous `CAMAsyncAFDConnector`.
- Reuse native vLLM/vLLM-Ascend DeepSeek V4 lifecycle and operators wherever
  possible.
- Keep AFD-specific behavior limited to role-aware model composition,
[...]

PR #191, opened 2026-08-04
V0.26.0 support for dsv4 (gpu)

lirx-pd:
## Purpose

Add plugin-owned DeepSeek V4 support for Attention/FFN Disaggregation (AFD). The change introduces the DeepSeek V4 model integration, routes model construction and worker execution according to the Attention or FFN role, and extends the P2P metadata and payload contracts needed to exchange the V4 expert-related data safely.

## Issue

- Related issue(s): #169 
- Closing keyword, only if fully resolved: N/A

## Scope

- In scope: 
   DeepSeek V4 model registration and construction; 
   role-aware Attention/FFN execution; 
   V4 weight-loading policy and proxy behavior;
   P2P metadata, payload, and expert-contract support;

- Out of scope:
  Compute gate on Attention side;
  Support for MegaMoE;
  Support for sequence-parallel MoE;
  Support for EPLB;
  PP>1;
 

## Implementation Notes

The implementation adds a plugin-owned `deepseek_v4` model module and exposes it through the existing package and connector registration points. The Attention and FFN runners select the V4-specific construction path while preserving the established AFD worker split. The GPU P2P connector carries token-aligned `input_ids` with the FFN activations so the remote V4 hash router uses the same token IDs as native vLLM. The IDs, including CUDA Graph padding slots, are forwarded without mutation to avoid changing hash-MoE routing or modifying vLLM's shared input buffer.

The change also keeps the DeepSeek V2 integration aligned with the new model-routing surface, and covers package discovery, connector factory behavior, P2P expert contracts, V4 construction, proxy behavior, and weight-loading policy with unit tests. No vLLM source files are modified. No new compatibility shim or monkey patch is introduced by this change.

## Test Plan

- Run the three focused DeepSeek V4 model test modules covering construction, proxy behavior, and weight-loading policy.
- Run Ruff lint and format checks for the modified code.
- Run Python bytecode compilation for the changed Python modules.
- Inspect the staged diff for unintended changes and verify the change set remains limited to the AFD implementation and its unit tests.
- Run end-to-end eager and CUDA Graph consistency tests against native vLLM with deterministic sampling (`temperature=0`) across prompt lengths 16/64/256 and batch sizes 1/2/4.
- Run a native-vLLM self-consistency control to distinguish AFD regressions from backend numerical nondeterminism.
- Run the DeepSeek V2 FFN runner regression to verify the V4 input-ID transport remains opt-in and V2 keeps its hidden-state-only connector contract.

## Test Result

- Focused DeepSeek V4 and P2P input-ID contract tests: 36 passed.
- Ruff lint check: passed.
- Ruff format check: passed.
- Python bytecode compilation: passed.
- Staged diff check: passed.
- Eager consistency 14/18 cases (36/42 exact choices).
- CUDA Graph consistency 12/18 cases (27/42 exact choices).
- Native CUDA Graph self-consistency was 10/18 cases (28/42 exact choices), confirming that the remaining token-level differences include native backend nondeterminism.
- DeepSeek V2 connector regression: passed; the FFN runner test module reports 27 passed.

## Docs Impact

- Files updated: None in the plugin repository.
- If none, reason: The change adds an internal model integration and test coverage without altering user-facing configuration or operational behavior that requires documentation updates.

---

<details>

- [x] Scope is bounded.
- [x] Any compat shim or monkey patch is isolated, idempotent, version-guarded, documented, and tested. N/A: this change introduces neither.

</details>

### DIFF
--- a/afd_plugin/connectors/gpu/p2p.py
+++ b/afd_plugin/connectors/gpu/p2p.py
@@ -80,7 +80,6 @@ from afd_plugin.connectors.base import (
 )
 from afd_plugin.connectors.metadata import (
     AFDA2FTransferPayload,
-    AFDA2FTransportSpec,
     AFDControlPayload,
     AFDDPMetadata,
     AFDExpertRoutingSpec,
@@ -99,6 +98,7 @@ if TYPE_CHECKING:
     from vllm.config import VllmConfig
 
 _AFD_COMMUNICATORS: dict[int, PyNcclCommunicator] = {}
+_INPUT_IDS_DTYPE = torch.int32
 _AFD_COMM_ID_COUNTER = 0
 _AFD_CUSTOM_OPS_REGISTERED = False
 
@@ -198,7 +198,7 @@ class P2pNcclAFDConnector(AFDConnectorBase):
             torch.Tensor,
         ] = {}
         self._recv_attn_input_ids_buffers: dict[
-            tuple[int, int, tuple[int, ...], torch.dtype],
+            tuple[int, int, tuple[int, ...]],
             torch.Tensor,
         ] = {}
         self.a2e_group: StatelessProcessGroup | None = None
@@ -315,8 +315,8 @@ class P2pNcclAFDConnector(AFDConnectorBase):
                 ``context.metadata.total_tokens``.
             context: Per-transfer context describing the token layout.
             **kwargs: Optional ``router_logits`` and token-aligned ``input_ids``
-                tensors. The stable version-1 wire order is hidden states,
-                router logits when present, then input IDs when present.
+                tensors. The wire order is hidden states, router logits when
+                present, then input IDs when present.
 
         Raises:
             ValueError: If the tensor shape does not match the metadata token
@@ -334,11 +334,6 @@ class P2pNcclAFDConnector(AFDConnectorBase):
             )
         router_logits: torch.Tensor | None = kwargs.get("router_logits")
         input_ids: torch.Tensor | None = kwargs.get("input_ids")
-        transport_spec: AFDA2FTransportSpec | None = kwargs.get("transport_spec")
-        if (transport_spec is None) != (input_ids is None):
-            raise ValueError(
-                "input_ids presence must exactly match the AFD transport spec",
-            )
         if (
             router_logits is not None
             and router_logits.shape[0] != hidden_states.shape[0]
@@ -352,11 +347,8 @@ class P2pNcclAFDConnector(AFDConnectorBase):
                     "input_ids must be one-dimensional and token-aligned with "
                     "hidden_states",
                 )
-            assert transport_spec is not None
-            if input_ids.dtype != transport_spec.input_ids_dtype:
-                raise ValueError(
-                    "P2P AFD input_ids dtype does not match the transport spec",
-                )
+            if input_ids.dtype != _INPUT_IDS_DTYPE:
+                raise ValueError("P2P AFD input_ids must use torch.int32")
         self._send_hidden_states(
             hidden_states,
             0,
@@ -430,8 +422,8 @@ class P2pNcclAFDConnector(AFDConnectorBase):
 
         Args:
             ubatch_idx: Stage/microbatch index to receive. Defaults to ``0``.
-            **kwargs: Optional ``routing_spec`` and versioned ``transport_spec``
-                describing model-owned Attention-to-FFN side tensors.
+            **kwargs: Optional ``routing_spec`` and ``recv_input_ids`` flag for
+                model-owned Attention-to-FFN tensors.
 
         Returns:
             ``AFDA2FTransferPayload`` with the concatenated hidden states and a
@@ -443,16 +435,7 @@ class P2pNcclAFDConnector(AFDConnectorBase):
                 has no Attention peers.
         """
         routing_spec: AFDExpertRoutingSpec | None = kwargs.get("routing_spec")
-        transport_spec: AFDA2FTransportSpec | None = kwargs.get("transport_spec")
-        if transport_spec is not None and transport_spec.version != 1:
-            raise ValueError(
-                f"unsupported AFD A2F transport version {transport_spec.version}",
-            )
-        if (
-            transport_spec is not None
-            and transport_spec.input_ids_dtype not in (None, torch.int64)
-        ):
-            raise ValueError("P2P AFD input_ids transport requires torch.int64")
+        recv_input_ids: bool = kwargs.get("recv_input_ids", False)
         hidden_states_list: list[torch.Tensor] = []
         router_logits_list: list[torch.Tensor] = []
         input_ids_list: list[torch.Tensor] = []
@@ -495,24 +478,16 @@ class P2pNcclAFDConnector(AFDConnectorBase):
                         router_metadata,
                     ),
                 )
-            if (
-                transport_spec is not None
-                and transport_spec.input_ids_dtype is not None
-            ):
+            if recv_input_ids:
                 input_ids_metadata = _TensorMetadata(
                     device=tensor_metadata.device,
-                    dtype=transport_spec.input_ids_dtype,
+                    dtype=_INPUT_IDS_DTYPE,
                     size=torch.Size([tensor_metadata.size[0]]),
                 )
                 input_ids_ref = None
                 if not self.vllm_config.model_config.enforce_eager:
                     input_ids_ref = self._recv_attn_input_ids_buffers.get(
-                        (
-                            ubatch_idx,
-                            src,
-                            tuple(input_ids_metadata.size),
-                            input_ids_metadata.dtype,
-                        ),
+                        (ubatch_idx, src, tuple(input_ids_metadata.size)),
                     )
                 input_ids_list.append(
                     self._recv_hidden_states(
@@ -816,21 +791,17 @@ class P2pNcclAFDControlPlane(AFDControlPlane):
                         dtype=tensor_metadata.dtype,
                         device=tensor_metadata.device,
                     )
-                if payload.transport_spec is not None:
-                    input_ids_key = (
-                        stage_idx,
-                        src_rank,
+                input_ids_key = (
+                    stage_idx,
+                    src_rank,
+                    (tensor_metadata.size[0],),
+                )
+                if input_ids_key not in connector._recv_attn_input_ids_buffers:
+                    connector._recv_attn_input_ids_buffers[input_ids_key] = torch.empty(
                         (tensor_metadata.size[0],),
-                        payload.transport_spec.input_ids_dtype,
+                        dtype=_INPUT_IDS_DTYPE,
+                        device=tensor_metadata.device,
                     )
-                    if input_ids_key not in connector._recv_attn_input_ids_buffers:
-                        connector._recv_attn_input_ids_buffers[input_ids_key] = (
-                            torch.empty(
-                                (tensor_metadata.size[0],),
-                                dtype=payload.transport_spec.input_ids_dtype,
-                                device=tensor_metadata.device,
-                            )
-                        )
 
     def send_dp_metadata_list(
         self,
--- a/afd_plugin/connectors/gpu/p2p.py
+++ b/afd_plugin/connectors/gpu/p2p.py
@@ -197,7 +197,7 @@ class P2pNcclAFDConnector(AFDConnectorBase):
             tuple[int, int, tuple[int, ...]],
             torch.Tensor,
         ] = {}
-        #This is used only for the hash MOE of deepseek v4 now.
+        # This is used only for the hash MOE of deepseek v4 now.
         self._recv_attn_input_ids_buffers: dict[
             tuple[int, int, tuple[int, ...]],
             torch.Tensor,
--- a/afd_plugin/connectors/gpu/p2p.py
+++ b/afd_plugin/connectors/gpu/p2p.py
@@ -80,6 +80,7 @@ from afd_plugin.connectors.base import (
 )
 from afd_plugin.connectors.metadata import (
     AFDA2FTransferPayload,
+    AFDA2FTransportSpec,
     AFDControlPayload,
     AFDDPMetadata,
     AFDExpertRoutingSpec,
@@ -196,6 +197,10 @@ class P2pNcclAFDConnector(AFDConnectorBase):
             tuple[int, int, tuple[int, ...]],
             torch.Tensor,
         ] = {}
+        self._recv_attn_input_ids_buffers: dict[
+            tuple[int, int, tuple[int, ...], torch.dtype],
+            torch.Tensor,
+        ] = {}
         self.a2e_group: StatelessProcessGroup | None = None
         self.e2a_group: StatelessProcessGroup | None = None
         self.p2p_pg: ProcessGroup | None = None
@@ -309,7 +314,9 @@ class P2pNcclAFDConnector(AFDConnectorBase):
                 whose leading dimension matches
                 ``context.metadata.total_tokens``.
             context: Per-transfer context describing the token layout.
-            **kwargs: Optional ``router_logits`` tensor sent after hidden states.
+            **kwargs: Optional ``router_logits`` and token-aligned ``input_ids``
+                tensors. The stable version-1 wire order is hidden states,
+                router logits when present, then input IDs when present.
 
         Raises:
             ValueError: If the tensor shape does not match the metadata token
@@ -326,6 +333,12 @@ class P2pNcclAFDConnector(AFDConnectorBase):
                 f"AFD metadata token count {metadata.total_tokens}",
             )
         router_logits: torch.Tensor | None = kwargs.get("router_logits")
+        input_ids: torch.Tensor | None = kwargs.get("input_ids")
+        transport_spec: AFDA2FTransportSpec | None = kwargs.get("transport_spec")
+        if (transport_spec is None) != (input_ids is None):
+            raise ValueError(
+                "input_ids presence must exactly match the AFD transport spec",
+            )
         if (
             router_logits is not None
             and router_logits.shape[0] != hidden_states.shape[0]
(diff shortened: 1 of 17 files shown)


jiangkuaixue123 (commented):
Reviewed at `7394b5e5110e9410af480606f979c52bd01b4526` with InferMatrixCopilot (Direct mode).

I did not find an actionable correctness issue in the inspected DeepSeek-V4 GPU split, A2F `input_ids` transport contract, P2P ordering changes, model registration, or Attention/FFN runner integration.

Validation performed:
- `git diff --check origin/main...HEAD`: passed
- `python -m compileall -q afd_plugin`: passed
- Focused unit tests were collected locally, but skipped because this review environment does not have `torch` / `vllm` installed. GitHub currently exposes only the passing DCO check, so the new tests and a GPU smoke/E2E run remain validation gaps before merge.

The PR is still a draft; please treat this as a code-review pass on the frozen head, not as confirmation that the runtime matrix has executed.

afd_plugin/compat/npu/runtime.py
```diff
@@ -29,8 +29,8 @@ def apply_afd_ascend_patches_if_needed() -> None:
         apply_afd_ascend_dbo_config_patch,
     )
 
-    apply_afd_ascend_dbo_config_patch()
-    _PATCHES_APPLIED = True
+    if apply_afd_ascend_dbo_config_patch():
```
jiangkuaixue123: **[major]** The diff changes `apply_afd_ascend_dbo_config_patch()` at afd_plugin/compat/npu/runtime.py:32 from unconditionally executing the patch to returning a bool (True on success, False if vLLM-Ascend is not importable). The new call site at config_validation.py:56-60 tests the return value with `if apply_afd_ascend_dbo_config_patch():` but does NOT handle the False case—when vLLM-Ascend is absent, the patch returns False and the code proceeds without applying the patch. This creates a silent no-op risk if the Ascend config is present in additional_config but the platform module is unavailable. Either (a) propagate the failure (raise an error when AFD config is present but the patch fails), or (b) document that the False case is expected and benign (e.g., when running CPU-only tests), and update the call site to log or skip the AFD-NPU path explicitly when the patch returns False.

Evidence: Read afd_plugin/compat/npu/runtime.py lines 17-34 and afd_plugin/compat/patches/config_validation.py lines 56-62; the new return-bool contract is not handled at the call site, creating a potential silent failure when vLLM-Ascend is expected but missing.

afd_plugin/connectors/metadata.py
```diff
@@ -324,6 +349,14 @@ def encode_control_payload(payload: AFDControlPayload) -> bytes:
         "dp_metadata_list": metadata_payload,
```
jiangkuaixue123, line 328: **[major]** The diff adds transport_spec to the AFDControlPayload wire encoding (encode_control_payload line 349-358 in the diff) and decode_control_payload (line 381-391), changing the JSON schema sent over the DP metadata control plane. If an Attention rank running the new code sends a payload with transport_spec to an FFN rank running the old code (pre-PR), the old decode_control_payload will call payload.get('transport_spec') which returns None for a missing key, and construct AFDControlPayload with transport_spec=None (line 396 after the if block). The FFN rank will then proceed with a None transport_spec even if the model requires input_ids transport. Later, when the FFN connector tries to allocate input_ids buffers (p2p.py:837-849 in the diff), it checks payload.transport_spec is not None before creating the buffer; if transport_spec is None, no buffer is allocated, and the subsequent recv will allocate a fresh tensor on every forward pass rather than reusing a graph-captured buffer. This is a behavior regression (performance loss) rather than a hard failure, but if future code paths assert transport_spec is not None when the model requires it, mixed-version Attention/FFN ranks will fail at runtime. Document the wire-protocol versioning policy: either (a) require homogeneous Attention/FFN versions during rollout, or (b) add a version field to the JSON schema and have decode_control_payload handle missing transport_spec by inferring a default (e.g., AFDA2FTransportSpec() with default version/dtype) when the sender is old.

Evidence: diff metadata.py:349-391 adds transport_spec to wire schema; grep afd_plugin/connectors/gpu/p2p.py 'payload.transport_spec' shows usage at line 837; no version migration logic in decode_control_payload.

afd_plugin/compat/patches/npu/force_load_balance.py
```diff
@@ -181,181 +176,63 @@ def _get_force_lb_topk_ids(
...
-    # expert ids for Ascend W8A8 MoE layers.
+    # ### PATCH START: capture AFD force-load-balance configuration
+    additional_config = vllm_config.additional_config or {}
     self.enable_force_load_balance = bool(
         additional_config.get("enable_force_load_balance", False)
     )
     self.force_load_balance_topn_per_rank = int(
         additional_config.get("force_load_balance_topn_per_rank", 0)
     )
     self.max_force_lb_tokens = _get_force_lb_max_tokens(vllm_config)
```
jiangkuaixue123, line 227: **[major]** The diff moves the force_lb_fake_topk_buffer initialization from the layer's __init__ (old line 243-250 in the diff: initialized on layer after layer.w13_weight existed) to the quant method's __init__ (new line 227 in the diff: self.force_lb_fake_topk_buffer = None) with lazy initialization in apply() (new line 337-344). The old code patched AscendFusedMoE.__init__ and stored the buffer on self (the layer instance); the new code patches AscendW8A8DynamicFusedMoEMethod.__init__ and stores the buffer on self (the method instance). The diff does not show that the method instance is retained by the layer across forward passes. If vLLM-Ascend recreates the quant method instance on each forward pass (or if the method is instantiated once but not held by the layer), the buffer at self.force_lb_fake_topk_buffer will be None on every apply() call, triggering re-initialization at line 337-344 and defeating the point of the patch (deterministic buffer reuse). Verify that AscendW8A8DynamicFusedMoEMethod instances are created once at layer construction and retained by the layer (e.g., layer.quant_method = method), so self.force_lb_fake_topk_buffer survives across calls. If the method lifecycle is per-call or per-batch, move the buffer storage back to the layer instance (layer.force_lb_fake_topk_buffer) and have apply() read/write it there.

Evidence: diff force_load_balance.py:149-227 shows patch target change from AscendFusedMoE.__init__ to AscendW8A8DynamicFusedMoEMethod.__init__; buffer stored on self (the method instance) at line 227; lazy init at line 337-344 checks self.force_lb_fake_topk_buffer is None; no proof method instance is retained across forward passes.

afd_plugin/compat/patches/npu/ascend_platform.py
```diff
@@ -17,83 +18,95 @@
...
-    saved: dict[str, bool | int],
+    saved: _AFDDBOConfigSnapshot,
 ) -> None:
     parallel_config = vllm_config.parallel_config
-    if not (
-        saved["enable_dbo"]
-        or saved["use_ubatching"]
-        or int(saved["ubatch_size"] or 0) != 0
-    ):
+    if not saved.enable_dbo and saved.ubatch_size == 0:
```
jiangkuaixue123, line 100: **[major]** The `_restore_afd_dbo_config` guard at afd_plugin/compat/patches/npu/ascend_platform.py line 99 in the diff changed from checking `enable_dbo OR use_ubatching OR ubatch_size != 0` to `not saved.enable_dbo and saved.ubatch_size == 0`. The new guard skips the restore (including restoring `all2all_backend`) when `enable_dbo=False` and `ubatch_size=0`, even if `all2all_backend` was snapshotted as a non-default value. The old code also checked `use_ubatching`, which is no longer in the snapshot. A config that has `enable_dbo=False`, `ubatch_size=0`, but `all2all_backend` set to `_AFD_TEMP_BACKEND` would silently leave the temporary backend on the parallel config after the NPU platform pass. Add a test for this edge case and expand the guard to include `all2all_backend` restoration independently: `if not saved.enable_dbo and saved.ubatch_size == 0 and saved.all2all_backend == parallel_config.all2all_backend: return`.

Evidence: Diff hunk in afd_plugin/compat/patches/npu/ascend_platform.py shows the old dict-based guard checked three fields; the new dataclass guard checks only two. The `all2all_backend` field is now snapshotted (`_AFDDBOConfigSnapshot` includes it) but the restore-skip condition does not account for it. No test in the diff exercises enable_dbo=False, ubatch_size=0 with a non-default all2all_backend.

jiangkuaixue123 (changes requested):
**Scan:**
| Category | Result |
|---|---|
| Tests / verification | 3 finding(s) below |
| Docs / comments | no finding reported |
| Correctness | 2 finding(s) below |

**Validated:**
- [sweep] afd_plugin/compat/npu/runtime.py:32 — new branch `if apply_afd_ascend_dbo_config_patch():` tests the return value; verified the function now returns bool (True on success, False on import failure), and the call site in config_validation.py:56-60 wraps the test but does NOT handle the False case explicitly — when vLLM-Ascend is absent, the patch returns False and the surrounding code proceeds without the patch, which may be intended but is a silent no-op risk if the Ascend config is present in additional_config yet the platform module is unavailable.
- [validated] afd_plugin/compat/patches/async_dp_engine.py:92 — new branch `if parallel_config.numa_bind:` is reachable because upstream vLLM 0.26 added the numa_bind field; verified by reading the diff hunk that copies the upstream pattern from the target release.
- [validated] afd_plugin/compat/patches/async_dp_engine.py:356 — new branch `if not self.engines_running and not is_afd_async_dp(self.vllm_config):` is the patched version of the FIRST_REQ wakeup; both arms (AFD async-DP and non-AFD) are reachable, and the logic preserves the upstream wakeup for non-AFD configs — no dead branch.
- [validated] afd_plugin/compat/patches/async_dp_forward_context.py:100-114 — the two new branches `if (num_tokens_across_dp is None and vllm_config.parallel_config.data_parallel_size > 1)` and `elif num_tokens_across_dp is None` handle the DP-size-1 and DP-size>1 cases separately; both arms are reachable, and the logic matches the upstream 0.26 pattern for sequence-parallel MoE — no dead branch.
- [validated] afd_plugin/compat/patches/config_validation.py:56-60 — the new `if parse_optional_afd_config(self.additional_config) is not None:` branch is reachable when AFD config is present; the nested `if current_platform.device_type == 'npu':` is reachable on NPU platforms; verified by reading the diff context.
- [validated] afd_plugin/compat/patches/config_validation.py:64-90 — the `if not _should_relax_engine_args_backend(self):` / `else:` branches handle AFD-ubatching and non-AFD configs; both arms are reachable, and the logic correctly wraps the upstream call in the non-AFD case and the backend-relaxation patch in the AFD case — no dead branch.
- [sweep] afd_plugin/compat/patches/config_validation.py:154 — new guard `if not engine_args.enable_dbo and engine_args.ubatch_size <= 1: return False` short-circuits the relaxation when DBO is disabled and ubatch_size is 1 or less; this is correct and reachable.
- [validated] afd_plugin/compat/patches/config_validation.py:168-172 — the checks `if parse_optional_afd_config(vllm_config) is None:` and `if not parallel_config.use_ubatching:` are both reachable and correctly guard the backend relaxation for AFD configs only — no dead branch.

Retained 17 of 18 candidates. Dropped one (i=8) because the diff does not touch the file it references. Kept two blockers (i=4, i=5 deduplicated) for missing NPU connector signature updates, and six majors covering silent failure paths, wire-protocol versioning gaps, lifecycle assumptions, and guard logic completeness. The remaining minors request docstring updates, explicit assertions for indexed accesses, and test coverage for non-exercised branches. No candidate was dropped for being a pure stylistic alternative; all retained items cite concrete evidence from the diff or named repo files.

**Verdict:** REQUEST CHANGES

### Findings without a valid current diff anchor

These findings are preserved in the review body because their file/line could not be mapped to the fetched PR diff.

- `afd_plugin/connectors/npu/camp2p.py:395` [blocker] — This diff adds router_logits, input_ids, and transport_spec kwargs to P2pNcclAFDConnector.send_attn_output (p2p.py:326-388 in the diff) and routing_spec/transport_spec to recv_attn_output (p2p.py:454-565), changing the connector data-path contract. CAMP2pAFDConnector.send_attn_output and recv_attn_output in afd_plugin/connectors/npu/camp2p.py (line 395 cited) are NOT updated in the diff to accept or produce those fields. Grep evidence shows afd_plugin/v1/worker/npu/ffn_model_runner.py:252 calls connector.recv_attn_output, and afd_plugin/model_executor/models/npu/deepseek_v2_async_cam_forward.py:168+288 call send_attn_output. If any Attention-side NPU code supplies router_logits= or input_ids= (for example a future DeepSeek V4 NPU path), the call will fail with an unexpected-keyword-argument error because this connector's signature is **kwargs but does not handle those keys. If any FFN-side code reads payload.router_logits or payload.input_ids, it will get None on NPU even when the model requires those fields. Update CAMP2pAFDConnector.send_attn_output to accept router_logits, input_ids, and transport_spec in kwargs and extend the CAMP2P custom op to transmit them in the same sequence as P2P. Update recv_attn_output to accept routing_spec and transport_spec, receive those tensors when the specs are present, and return them in AFDA2FTransferPayload. Evidence: grep afd_plugin/connectors/npu/camp2p.py 'def send_attn_output' shows line 395; grep afd_plugin/v1/worker/npu/ 'recv_attn_output' shows ffn_model_runner.py:252; diff shows P2P signature change at p2p.py:326-388 and 454-565 but no corresponding CAMP2P update.

jiangkuaixue123:
Please rebase to main

afd_plugin/connectors/gpu/p2p.py
```diff
@@ -195,6 +196,10 @@ def __init__(
             tuple[int, int, tuple[int, ...]],
             torch.Tensor,
         ] = {}
+        self._recv_attn_input_ids_buffers: dict[
```
jiangkuaixue123, line 201: Why we need _recv_attn_input_ids_buffers

afd_plugin/compat/npu/runtime.py
lirx-pd: It is not related with the support for deepseekv4 GPU and is consistent with the main branch now

afd_plugin/compat/patches/npu/force_load_balance.py
lirx-pd: This pr does not change this file. It remians the same as the main branch

afd_plugin/compat/patches/npu/ascend_platform.py
lirx-pd: This pr does not change this file. It remians the same as the main branch

afd_plugin/connectors/gpu/p2p.py
lirx-pd: The hash moe need the input_ids after computing the gate. And the gate computation is placed at FFN side for GPU. Therefore, the input_ids should be sent to the FFN side

afd_plugin/connectors/metadata.py
lirx-pd: The readme.md has explicitly documented that Attention and FFN ranks must run a homogeneous plugin version now.

afd_plugin/connectors/gpu/p2p.py
jiangkuaixue123: Please add a comment: this is used only for dsv4

jiangkuaixue123 (commented):
Please also include the GPU E2E test results.

afd_plugin/connectors/metadata.py
```diff
@@ -159,6 +160,26 @@ class AFDExpertRoutingSpec(NamedTuple):
     router_logits_dtype: torch.dtype
 
 
+@dataclass(frozen=True, slots=True)
+class AFDA2FTransportSpec:
```
specture724: what is `spec` short for?

afd_plugin/connectors/metadata.py
```diff
@@ -159,6 +160,26 @@ class AFDExpertRoutingSpec(NamedTuple):
...
+    """Versioned model-owned contract for Attention-to-FFN side inputs.
+
+    Version 1 preserves the historical hidden-state/router-logits ordering and
+    optionally appends token-aligned ``input_ids``. The explicit dtype prevents
+    transport implementations from coercing integer identifiers to the model
+    activation dtype.
+    """
+
+    version: int = 1
```
specture724: what case will cause `version != 1`

afd_plugin/connectors/metadata.py
lirx-pd: remove

afd_plugin/connectors/metadata.py
lirx-pd: remove

afd_plugin/connectors/gpu/p2p.py
lirx-pd: have added

afd_plugin/connectors/metadata.py
lirx-pd: transport_spec has been removed now

README.md
```diff
@@ -42,7 +42,7 @@ Model support:
...
 | --- | --- | --- | --- |
-| DeepSeekV2 / DeepSeekV3 / DeepSeekV3.2 | `DeepseekForCausalLM`, `DeepseekV2ForCausalLM`, `DeepseekV3ForCausalLM`, `DeepseekV32ForCausalLM` | `AFDDeepseekForCausalLM`, `AFDDeepseekV2ForCausalLM`, `AFDDeepseekV3ForCausalLM` | DeepSeekV3.2 uses `AFDDeepseekV3ForCausalLM`. Each AFD role constructs and loads only its role-required model components, while shared embedding, normalization, and output components remain available where required by the model lifecycle. |
+| DeepSeekV2 / DeepSeekV3 / DeepSeekV3.2 / DeepSeekV4 | `DeepseekForCausalLM`, `DeepseekV2ForCausalLM`, `DeepseekV3ForCausalLM`, `DeepseekV32ForCausalLM`,`DeepseekV32ForCausalLM` | `AFDDeepseekForCausalLM`, `AFDDeepseekV2ForCausalLM`, `AFDDeepseekV3ForCausalLM`, `AFDDeepseekV3ForCausalLM` | DeepSeekV3.2 uses `AFDDeepseekV4ForCausalLM`. Each AFD role constructs and loads only its role-required model components, while shared embedding, normalization, and output components remain available where required by the model lifecycle. DeepSeekV4 is currently only supported on the CUDA platform and the Attention and FFN ranks must run a homogeneous plugin version. |
```
jiangkuaixue123: Please remove this README change from the current PR. After the model implementation is merged, please submit a follow-up PR that adds the DeepSeek-V4 E2E test coverage and updates the README to reflect DeepSeek-V4 support in that same PR.

jiangkuaixue123 (commented):
Please address the README update in a follow-up PR.

README.md
lirx-pd: already removed

jiangkuaixue123:
Has the existing DeepSeek-V2 support been regression-tested with these changes? If so, please share the test results.

lirx-pd:
> Has the existing DeepSeek-V2 support been regression-tested with these changes? If so, please share the test results.

Add the test results in pr

PR #186, opened 2026-08-03
[v0.26.0] Upgrade GPU and NPU backends

jiangkuaixue123:
## Purpose

Upgrade the AFD GPU and NPU backends together from the vLLM 0.19.1 contract to vLLM 0.26.0.

This PR promotes the completed `update/v0.26.0` integration branch to `main` as one combined change. It brings together the GPU remote-experts refactor from #182 and the NPU runtime, PCP-removal, and DBO compatibility work from #183, #184, and #185.

## Issue

- Related PRs: #182, #183, #184, #185
- No issue is closed by this PR.

## Scope

- In scope:
  - Pin the plugin to vLLM 0.26.0 and refresh the lockfile.
  - Move the GPU DeepSeek MoE split to the remote-experts boundary while preserving the native model hierarchy.
  - Support both Attention-side and FFN-side gate placement, ordered P2P custom ops, CUDA Graph execution, and DBO paths on GPU.
  - Align the Ascend/NPU attention and FFN runners, forward context, fused-MoE selection, W8A8 force-load-balance path, and compatibility patches with vLLM/vLLM-Ascend 0.26 APIs.
  - Remove PCP-only model-runner-v1 paths that vLLM-Ascend 0.26 no longer supports.
  - Restore AFD NPU DBO and two-ubatch behavior while retaining the runtime communication backend.
  - Update focused unit coverage and design documentation for the new runtime contract.
- Out of scope:
  - Compatibility with vLLM 0.19.1; this is an intentional single-version upgrade.
  - GPU ModelRunner V2.
  - Quantized GPU MoE and EPLB validation.
  - NPU CAMAsync validation and a dedicated real-W8A8 force-load-balance E2E model.

## Implementation Notes

The GPU path keeps native `Model.forward`, `DecoderLayer.forward`, and `DeepseekV2MoE.forward`, and installs a parameter-free Attention experts proxy at the remote-experts boundary. Routed and shared experts remain owned by the FFN model runners, and the existing connector protocol is reused.

The NPU path rebases plugin-owned runners and isolated compatibility patches onto the v0.26 APIs. It removes stale PCP fields and hooks, refreshes the DeepSeek `FusedMoE` factory after plugin initialization, captures force-load-balance configuration during method construction, and preserves AFD DBO settings across Ascend platform validation.

## Test Plan

- Run the repository CPU-only CI on the combined PR.
- Retain the completed GPU eager, CUDA Graph, TP/EP, DP/EP, and DBO validation from #182.
[...]

PR #203, opened 2026-08-06
feat(npu): port token-balanced Async CAM MoE ubatching to vLLM 0.26

### LINKED ISSUE #149: [Feature]: Support token-based CAMAsync microbatch splitting for non-PCP DP+TP/SP topologies

ShwStone:
## Summary

This PR is the vLLM 0.26 counterpart of #180.

It implements the token-balanced Async CAM MoE ubatching requested by #149 on
top of the vLLM 0.26 architecture introduced by #186. This is a refactor, not
a mechanical rebase of the v0.19 implementation: the execution path follows
the vLLM 0.26 and vLLM-Ascend 0.26 forward-context, metadata-builder, and
sequence-parallel contracts.

The v0.19.1-specific implementation and its prior NPU validation remain in
#180, which targets `release/v0.19.1rc1`.

Closes #149.

## What changed

- Add request-boundary and token-balanced two-stage planning for NPU Async CAM
  MoE execution.
- Keep the stage planner independent of torch, vLLM, and the NPU runtime.
- Distinguish real scheduled tokens, parent-batch physical padding, and
  stage-local TP/SP padding.
- Isolate full-batch and per-stage Ascend attention metadata builders and their
  mutable metadata inputs/caches. This prevents one stage build from mutating
  another stage or the dense full-batch path.
- Use vLLM 0.26 `override_forward_context()` with shallow child contexts for
  stage attention execution.
- Convert full-batch FlashComm1/SP shards into stage-local shards and restore
  the parent layout after the MoE pipeline.
- For plain TP, shard replicated tokens only at the CAM boundary so every token
  is dispatched once, then restore the replicated model layout after FFN.
- Keep Attention and FFN topology validation independent. The primary small
  E2E topology is Attention DP3TP2 and FFN DP2TP1/EP2.
- Retain opt-in shape diagnostics through `AFD_ASYNC_MOE_LAYOUT_LOG=1`.
- Add a full 61-layer DeepSeek-V3.2 DP2TP8+EP16 NPU accuracy recipe and reserve
  20% device-memory headroom (`--gpu-memory-utilization 0.8`) on both roles.

No vLLM or vLLM-Ascend source-tree modification is required.

## Refactoring relative to #180

The v0.26 port removes plugin-owned layers that upstream v0.26 now makes
unnecessary:

- use vLLM's forward-context override instead of a custom mutation/rollback
  context manager;
- keep parent AFD connector metadata shared instead of cloning it per stage;
- keep real and physical stage extents in one immutable stage description;
- combine hidden-state and residual gather/restore operations when they share
  the same token layout;
- separate pure stage planning from NPU tensor-layout conversion without
  coupling either module to a model runner or connector.

## Supported v0.26 scope

- Ascend `CAMAsyncAFDConnector`;
- eager prefill and regular autoregressive decode execution;
- exactly two AFD-managed MoE stages;
- request and token split policies;
- FlashComm1/SP and plain-TP Attention layouts;
- Attention TP greater than one for token splitting;
- independently configured FFN TP/EP topology.

Explicitly unsupported:

- native vLLM DBO/ubatching with Async CAM;
- ACL graph execution;
- prefill or decode context parallelism;
- PCP on vLLM-Ascend 0.26 model runner v1;
- more than two Async CAM stages.

The legacy PCP behavior referenced by #149 remains version-specific to #180.

## Validation

Current-head local checks:

```bash
uv run --locked ruff check \
  afd_plugin/model_executor/models/npu/deepseek_v2_async_cam_forward.py \
  afd_plugin/v1/worker/npu/attention_model_runner.py
uv run --locked ruff format --check \
  afd_plugin/model_executor/models/npu/deepseek_v2_async_cam_forward.py \
  afd_plugin/v1/worker/npu/attention_model_runner.py
uv run --locked pytest -q \
  tests/unit/model_executor/models/test_forward_context.py \
  tests/unit/model_executor/test_async_cam_ubatching.py
bash -n \
  recipe/npu/CAMAsyncAFDConnector/deepseek_v3_2/v0_26_accuracy/attention_dp2tp8.sh \
  recipe/npu/CAMAsyncAFDConnector/deepseek_v3_2/v0_26_accuracy/ffn_ep16.sh
git diff --check
```

The focused CPU tests report 7 passed. The NPU runtime-only unit module is
skipped on the local non-NPU host.

NPU evidence is recorded at two distinct scopes:

- the DP3TP2 Attention + DP2TP1/EP2 FFN six-case matrix passed before the
  metadata-ownership fix;
- after the fix, the complete 61-layer DeepSeek-V3.2 DP2TP8+EP16 token-split
  deployment scored **97.5% GSM8K exact match (195/200)** on the fixed first
  200 samples.

The 200-sample result validates the corrected metadata path on the target full
model, but it is not a full-dataset accuracy claim. A post-fix rerun of the
six-case matrix and a full GSM8K evaluation remain follow-up validation. The
PR remains draft until the requested hardware matrix gate is closed.

Six-case matrix command:

```bash
AFD_NPU_ASYNC_CAM_RUN_DP3TP2=1 \
AFD_NPU_ASYNC_CAM_E2E_DEVICES=0,1,2,3,4,5,6,7 \
python -m pytest -svv \
  tests/e2e/accuracy/test_gsm8k_npu_async_cam.py::test_gsm8k_lm_eval_async_cam_dp3tp2_ep2
```

The matrix covers token/request/no-ubatch with Attention SP enabled and
disabled. The full-model launch and accuracy commands are documented in
`recipe/npu/CAMAsyncAFDConnector/deepseek_v3_2/v0_26_accuracy/README.md`.

## Review request

Review is requested in particular for:

1. real-token versus physical-padding semantics;
2. isolation of full-batch and stage metadata builders and mutable caches;
3. stage ForwardContext values used by MLA gather/unpad/reduce-scatter;
4. SP and plain-TP CAM boundary conversion;
5. two-stage dispatch/receive ordering;
6. compatibility with other vLLM-Ascend 0.26 changes.


jiangkuaixue123:
We should run an accuracy evaluation for this change, preferably on DeepSeek-V3.2 as well. A two-node setup can be used:

- Attention: DP2, TP8
- FFN: EP16

Please validate all layers with forced expert load balancing disabled.

tests/unit/model_executor/models/test_forward_context.py
```diff
@@ -7,10 +7,19 @@
 
 torch = pytest.importorskip("torch")
 pytest.importorskip("vllm")
+from vllm.forward_context import (  # noqa: E402
+    get_forward_context as get_current_forward_context,
```
jiangkuaixue123, line 12: **[blocker]** The diff removes ASYNC_MOE_UBATCH_METADATA_KEY and get_async_moe_ubatch_metadata_from_forward_context from afd_plugin/model_executor/models/forward_context.py and its __init__.py (they now live in models/npu/async_cam_layout.py). If tests/unit/model_executor/models/test_forward_context.py still imports these from afd_plugin.model_executor.models, the import fails at collection. Update the import to `from afd_plugin.model_executor.models.npu.async_cam_layout import ASYNC_MOE_UBATCH_METADATA_KEY, get_async_moe_ubatch_metadata_from_forward_context`.

Evidence: Grep showed models/__init__.py post-diff only exports get_afd_metadata_from_forward_context. Test file read showed lines 11-14 import both removed symbols from models. Diff archive confirms test_forward_context.py touched but imports unchanged.

docs/npu/CAM_ASYNC_CONNECTOR_USER_GUIDE.md
```diff
@@ -220,10 +252,12 @@ The CAM async v0.26 path was verified with:
 - the included `CAM_ascend910_93_openEuler_aarch64.run` installer;
```
jiangkuaixue123, line 252: **[minor]** The user guide (CAM_ASYNC_CONNECTOR_USER_GUIDE.md) now states the DP3TP2 NPU E2E matrix must pass before the v0.26 path is called hardware-validated, and the PR description confirms Ascend NPU E2E is not yet run. Before removing draft status, run `AFD_NPU_ASYNC_CAM_RUN_DP3TP2=1 AFD_NPU_ASYNC_CAM_E2E_DEVICES=0,1,2,3,4,5,6,7 pytest -svv tests/e2e/accuracy/test_gsm8k_npu_async_cam.py::test_gsm8k_lm_eval_async_cam_dp3tp2_ep2` and confirm the DP3TP2 x {request,token} x {SP on/off} cases meet the GSM8K threshold; the risk is shipping an unvalidated hardware path.

Evidence: diff at CAM_ASYNC_CONNECTOR_USER_GUIDE.md:252-260; PR description 'Not completed … Ascend NPU E2E validation'; execution_platforms.md:295 repeats requirement

afd_plugin/model_executor/npu/async_cam_ubatching.py
```diff
@@ -0,0 +1,122 @@
...
+        stage_bounds = (
+            (slice(0, split_request), 0, split_token),
+            (
+                slice(split_request, len(scheduled_tokens)),
+                split_token,
+                num_tokens,
+            ),
+        )
+    elif split == "token":
+        if tensor_parallel_size <= 1 or num_tokens < _ASYNC_MOE_STAGE_COUNT:
```
jiangkuaixue123, line 85: **[nit]** plan_async_moe_stages returns None in the token branch (async_cam_ubatching.py:82) when tensor_parallel_size <= 1 or num_tokens < _ASYNC_MOE_STAGE_COUNT, silently falling back to no ubatching. Add a docstring clause documenting the None fallback ('returns None when the batch cannot be split into two stages') so callers know the skip is intentional rather than an error.

Evidence: async_cam_ubatching.py:82–83 returns None for num_tokens < 2; test_async_cam_ubatching.py shows no single-token coverage

jiangkuaixue123 (changes requested):
**Scan:**
| Category | Result |
|---|---|
| Tests / verification | 4 finding(s) below |
| Docs / comments | 1 finding(s) below |
| Correctness | no finding reported |

**Validated:**
- [sweep] feature_validation.py:110 — checked new branch `if bool(parallel_config.enable_dbo) or bool(parallel_config.use_ubatching)`: both arms can occur (DBO-only, ubatch-only, both, neither); not dead.
- [sweep] feature_validation.py:147 — checked `if num_ubatches != ASYNC_MOE_NUM_STAGES`: branch occurs when config provides non-2 value; validated by test_npu_async_moe_ubatching_validation_rejects_unsupported_shape 'three-stages' case.
- [sweep] feature_validation.py:152 — checked `if split not in (ASYNC_MOE_REQUEST_SPLIT, ASYNC_MOE_TOKEN_SPLIT)`: branch occurs for unknown split strings; test does not show coverage for this exact ValueError.
- [sweep] feature_validation.py:159 — checked `if afd_config.is_ffn_server: return`: FFN role early-return skips attn_ranks_per_dp and context-parallel checks; logic correct for independent FFN topology but creates asymmetric validation—Attention rejects CP, FFN does not.
- [sweep] async_cam_ubatching.py:43–46 — checked `if not scheduled_tokens or any(token_count <= 0 for token_count in scheduled_tokens)`: both branches occur (empty list, nonpositive tokens); validated by test_stage_plan_rejects_nonpositive_token_counts.
- [sweep] async_cam_ubatching.py:50–54 — checked `if tensor_parallel_size <= 0`: branch occurs for invalid TP; raises ValueError; test does not show this exact case but constraint is reasonable.
- [sweep] async_cam_ubatching.py:99 — else branch for unknown split: raises ValueError; both token/request arms can occur, else is reachable for typos.
- [sweep] async_cam_layout.py:45–50 — AsyncMoeUbatchMetadata.__post_init__ validates num_stages matches attn_metadata length; both branches (empty, mismatch) can occur; raises ValueError correctly.

Verified 21 candidates against the diff and evidence pack. Kept two real blockers where the diff moves ASYNC_MOE_UBATCH_METADATA_KEY / get_async_moe_ubatch_metadata_from_forward_context out of forward_context.py and removes _use_async_moe_ubatch_forward_context, leaving touched-but-unupdated tests in test_forward_context.py that would break at import/reference (6, 7). Dropped eight candidates that directly contradict the diff or evidence (2 has a divisibility guard, 4 cites removed helpers, 5's runner builds the dataclass not a dict, 11/13/14 assert removals/updates the diff performs) or are already covered by the diff's own tests (16, 17, 20, 21). Deduplicated the shape[0] axis-0 assumption cluster (3 into 15) and demoted it to a documentation nit since the model-runner contract guarantees [tokens, hidden_size]. Remaining kept items are doc-consistency and coverage nits (8, 9, 10, 12, 15, 18) plus one minor NPU E2E verification reminder (19), all demoted below blocker because the author already flags the draft/validation status.

**Verdict:** REQUEST CHANGES

### Findings without a valid current diff anchor

These findings are preserved in the review body because their file/line could not be mapped to the fetched PR diff.

- `tests/unit/model_executor/models/test_forward_context.py:135` [blocker] — The diff removes the _use_async_moe_ubatch_forward_context context manager from deepseek_v2_async_cam_forward.py (imports switch from contextmanager/Iterator to copy + override_forward_context). If test_async_moe_ubatch_forward_context_restores_state in test_forward_context.py still references async_forward._use_async_moe_ubatch_forward_context, it raises AttributeError. Remove or rewrite the test to exercise the new copy(forward_context) + override_forward_context path. Evidence: Test file read showed lines 135-163 call async_forward._use_async_moe_ubatch_forward_context. Diff shows deepseek_v2_async_cam_forward.py removed the entire _use_async_moe_ubatch_forward_context context manager and replaced it with copy(forward_context) + override_forward_context. Grep of diff archive confirmed removal.
- `docs/design_zh/module/connector_contracts.md:77` [minor] — The diff updates docs/design/module/connector_contracts.md to describe async_moe_split as accepting request/token, and feature_validation.py now accepts both, but docs/design_zh/module/connector_contracts.md still reads '可选的请求边界流水线' (request-boundary only) and is not in the touched-files set. Update the Chinese translation to also mention token-balanced stages so the localized doc matches. Evidence: Read docs/design_zh/module/connector_contracts.md line 77 showing '可选的请求边界流水线'. Diff shows docs/design/module/connector_contracts.md:96 changed to 'request/token'. Grep confirmed Chinese doc not updated by this diff.

afd_plugin/model_executor/npu/async_cam_ubatching.py
ShwStone: Addressed in 7e2f3c3. The planner docstring now states that `None` means the selected policy cannot form two non-empty stages, and a single-token token-split case verifies this fallback. The dependency-free planner suite now reports 8 passed.

tests/unit/model_executor/models/test_forward_context.py
ShwStone: Fixed on the current head. `test_forward_context.py` now imports `ASYNC_MOE_UBATCH_METADATA_KEY` and `get_async_moe_ubatch_metadata_from_forward_context` directly from `models.npu.async_cam_layout`; the generic `models` namespace only exports AFD metadata. The updated CPU jobs and lint checks pass.

ShwStone:
Review follow-up on the current head (`7e2f3c3`):

- The stale `_use_async_moe_ubatch_forward_context` test has been removed. Its replacement, `test_async_moe_pipeline_preserves_stage_order`, exercises the current `copy(forward_context)` + `override_forward_context()` path, verifies stage metadata and padding, checks the two-stage send/receive order, and confirms the parent context is restored.
- The import blocker is fixed and its inline thread is resolved: NPU sidecar symbols are imported directly from `models.npu.async_cam_layout`.
- The planner `None` fallback is now documented and covered by a single-token test; that inline thread is resolved. The planner suite reports 8 passed.
- `docs/design_zh/module/connector_contracts.md` does not exist on the current `main` branch (nor elsewhere under `docs/`). The maintained English `docs/design/module/connector_contracts.md` is updated, so there is no current Chinese counterpart to synchronize.
- The NPU accuracy-validation thread remains open intentionally. The updated PR description distinguishes the six-case DP3TP2/EP2 matrix from the additionally requested DeepSeek-V3.2 DP2TP8/EP16 all-layer run with forced expert load balancing disabled.

ShwStone:
Accuracy Test Results: I tested on two 16-die 910C nodes with a topology of DP2TP8 on the Attention side and EP16 on the FFN side. Using the full DeepSeek v3.2 w8a8 model, the test results on the first 200 samples of the GSM8K dataset are as follows:

<table>
<tr>

<td> ubatch off

<td> request split

<td> token split

<tr>

<td> 88.5

<td> 74.5

<td> 63.5

</table>

I will continue to investigate the reasons for the accuracy degradation.

ShwStone:
### Accuracy regression: root cause and fix

The initial full-model validation exposed a clear accuracy regression on
DeepSeek-V3.2 W8A8 with Attention DP2TP8 and FFN EP16, using the first 200
GSM8K samples:

| Mode | Accuracy |
|---|---:|
| Async MoE ubatching disabled | 88.5% |
| Request split | 74.5% |
| Token split | 63.5% |

A deterministic 105-token TP8 smoke request localized the problem further:

- the layer-0 inputs were identical between token split and no ubatching;
- the layer-0 attention outputs already differed, before the first MoE layer;
- the large error transition occurred at zero-based token 52, adjacent to the
  53/52 stage boundary (`[0, 53)` and `[53, 105)`);
- request split was bitwise identical to no ubatching for this smoke request.

This ruled out weight loading, MoE dispatch, and FFN execution as the primary
cause. An NPU-side probe subsequently confirmed that constructing later
attention metadata changed data still referenced by metadata objects that had
already been built.

The root cause is a metadata ownership mismatch. vLLM-Ascend normally has only
one active attention metadata object, so several DeepSeek backends reuse mutable
runtime storage:

- SFA returns views into the global MLA RoPE runtime buffers and may update its
  C8 group metadata tensors in place;
- MLA decode, and no-RoPE MLA prefill, reuse runtime/identity RoPE buffers;
- DSA exposes reusable RoPE storage through `RopeDataProxy` and may modify
  block-table rows while building metadata;
- DSA ratio/SAS metadata caches are valid only for one token layout.

Async CAM constructs full-batch, stage-0, and stage-1 metadata before executing
the model. Later stage construction could therefore overwrite data referenced
by the full-batch or earlier-stage metadata. This also explains why the error
appeared in the first dense attention layer rather than only after MoE
ubatching began.

The fix is split across commits `65925b8` and `68bb799`:

1. reserve independent metadata builders for the full batch and both stages;
2. keep DSA prefill/decode/common ratio and SAS caches stage-local while still
   sharing them across attention groups with the same stage layout;
3. isolate backend inputs that are modified in place:
   - SFA C8 group metadata tensors;
   - DSA block tables;
4. materialize backend-owned outputs immediately after each build:
   - clone SFA and MLA RoPE tensors;
   - rebuild DSA RoPE proxies from the immutable full RoPE table;
5. apply the same ownership boundary to the full-batch metadata before either
   stage is built.

Read-only data such as KV caches, slot mappings, masks, and immutable RoPE
tables remain shared. The additional copies are restricted to AFD-managed Async
CAM stages; no-ubatch and native vLLM paths are unchanged.

The fix covers SFA, MLA, and DSA metadata, including prefill, decode, and mixed
batches. A single-token decode batch that cannot form two non-empty stages
continues through the normal unsplit path.

Focused unit and static checks pass. The full DeepSeek-V3.2 NPU accuracy
evaluation still needs to be rerun before considering the regression resolved.

afd_plugin/v1/worker/npu/attention_model_runner.py
```diff
@@ -1452,17 +1571,23 @@ def initialize_attn_backend(self, kv_cache_config: KVCacheConfig) -> None:
             )
             or self.afd_async_extra_info.async_moe_ubatching
         ):
-            self._ensure_two_metadata_builders()
+            num_metadata_builders = (
+                ASYNC_MOE_STAGE_METADATA_BUILDER_OFFSET
+                + int(self.afd_async_extra_info.async_moe_num_ubatches)
+                if self.afd_async_extra_info.async_moe_ubatching
+                else int(self.vllm_config.parallel_config.num_ubatches)
```
jiangkuaixue123, line 1581: For the current two-stage Async CAM path, this evaluates to three builders: builder 0 is the full-batch metadata builder used by the normal `_build_attention_metadata` path, while builders 1 and 2 are reserved for stage 0 and stage 1 respectively. The stage builders are selected later through `ASYNC_MOE_STAGE_METADATA_BUILDER_OFFSET + ubid`.

jiangkuaixue123 (commented):
Clarify the source and index mapping of the three Async CAM attention metadata builders.

afd_plugin/model_executor/models/npu/deepseek_v2_async_cam_forward.py
```diff
@@ -111,24 +121,28 @@ def run_attention_gate_afd_forward(
 
     afd_connector = afd_metadata.connector
     forward_context = get_forward_context()
-    stage_idx = int(
-        getattr(forward_context, "ubatch_idx", afd_metadata.stage_idx),
-    )
+    stage_idx = int(afd_metadata.stage_idx)
```
jiangkuaixue123: 不需要int()

ShwStone: Removed the redundant conversion in faf7ea4. `AFDForwardContextMetadata.stage_idx` is already typed as `int`, so the value is now used directly.

afd_plugin/v1/worker/npu/attention_model_runner.py
ShwStone: Clarified this mapping in faf7ea4 immediately above the offset constant: upstream builder 0 remains the full-batch builder, and Async CAM stage `ubid` selects builder `1 + ubid`.

docs/npu/CAM_ASYNC_CONNECTOR_USER_GUIDE.md
ShwStone: Updated the PR and user guide with the current hardware evidence. The pre-fix DP3TP2/EP2 six-case matrix is recorded separately from the post-fix full-model result: DeepSeek-V3.2 (all 61 layers), Attention DP2TP8 + FFN EP16, token split, scored 97.5% exact match (195/200) on the fixed first 200 GSM8K samples. I have not presented that limited-prefix run as a full-dataset result or as a post-fix rerun of the six-case matrix, and the PR remains draft pending that matrix gate.

ShwStone:
Accuracy and E2E update:

- The E2E changes have been removed from this feature PR. PR #222 has already merged the E2E framework refactor, so Async CAM E2E coverage will be submitted separately on top of that new framework. This keeps this PR focused on the runtime implementation, unit coverage, deployment recipe, and documentation.
- The post-fix full DeepSeek-V3.2 DP2TP8 Attention + EP16 FFN token-split deployment has now completed the full GSM8K evaluation and achieved **0.9522 strict match**.
- The documentation in this PR has been updated to replace the earlier limited 195/200 result with this complete-dataset result.

The E2E follow-up will preserve the two-step merge order: feature implementation first, then E2E coverage adapted to the framework introduced by #222.

Issue #209, opened 2026-08-07
[RFC]: Add Model-Owned Extensions to AFD Connectors


lirx-pd:
### Motivation

New model architectures may require the FFN side to receive model-specific data in addition to hidden states and routing tensors. For example, Deepseek-V4 requires token IDs to compute its FFN output correctly.

Encoding each new requirement directly in every connector creates several problems:

- Connector implementations become coupled to individual model architectures.
- Adding a model requires coordinated changes to GPU P2P, synchronous NPU CAMP2P, and asynchronous NPU CAM connectors.
- Model-specific tensor names, dtypes, shapes, buffers, and validation rules leak into the common transport layer.
- Existing models pay the maintenance cost for data that they never use.

The connector contract should therefore provide a small extension point while keeping all model-specific semantics in the model implementation.

### Proposed change

Introduce a plugin-owned `AFDConnectorExtension` with no-op defaults for the connector control and Attention-to-FFN data planes.

Each connector owns one default extension instance. After loading a model, the model runner calls `connector.bind_model(model)`. A model that requires additional communication exposes its implementation through:

```python
def get_afd_connector_extension(self) -> AFDConnectorExtension:
    return MyModelConnectorExtension()
```

The extension interface provides four hooks:

```python
class AFDConnectorExtension:
    def control_payload(self) -> dict[str, Any] | None:
        return None

    def update_state_from_control_payload(
        self,
        connector: AFDConnectorBase,
        payload: dict[str, Any] | None,
    ) -> None:
        pass

    def send_attn_output(
        self,
        connector: AFDConnectorBase,
        context: AFDTransferContext,
        **kwargs: Any,
    ) -> None:
        pass

    def recv_attn_output(
        self,
        connector: AFDConnectorBase,
        ubatch_idx: int,
    ) -> object | None:
        return None
```

The control plane carries one optional opaque JSON object in `AFDControlPayload.extension`. The FFN receive path stores the extension result in `AFDTransferMetadata.extension`, which is already propagated through the existing FFN forward context.

All current connectors invoke the same hooks:

- GPU P2P invokes both control-plane and data-plane hooks.
- Synchronous NPU CAMP2P invokes both control-plane and data-plane hooks.
- Asynchronous NPU CAM invokes data-plane hooks because it has no DP metadata control plane.

The common connector layer does not interpret extension tensor names, dtypes, shapes, buffer layouts, or model semantics. Models without an extension continue to use the no-op implementation and preserve their current behavior

[...]

PR #211, opened 2026-08-07
Add Model-Owned Extensions to AFD P2P Connectors

lirx-pd:
## Purpose

Add a model-owned extension mechanism to AFD connectors so new model architectures can transfer additional Attention-to-FFN control and data-plane information without adding model-specific fields and transport logic to every connector.

This change introduces a no-op `AFDConnectorExtension` in the common connector layer. Models may provide an implementation through `get_afd_connector_extension()`, while models without an extension preserve the existing communication behavior.

The proposal is described in **[RFC] Add Model-Owned Extensions to AFD Connectors**.

## Issue

- Related issue(s): #209 
- Closing keyword, only if fully resolved: N/A #209 

## Scope

- In scope:
  - Add the plugin-owned `AFDConnectorExtension` Null Object and model binding entry point.
  - Add an optional opaque extension object to the AFD control payload.
  - Add an optional opaque extension result to Attention-to-FFN transfer metadata.
  - Invoke extension hooks from GPU P2P, synchronous NPU CAMP2P, and asynchronous NPU CAM connectors.
  - Bind model-owned extensions after model loading in GPU and NPU Attention/FFN runners.
  - Preserve the current behavior for models that do not provide an extension.
  - Add unit coverage for binding, control-payload serialization, control-plane delegation, and data-plane hooks.
- Out of scope:
  - Model-specific extension implementations such as DeepSeek V4 `input_ids` transport.
  - A generic tensor schema, dtype/shape registry, or connector-managed extension buffers.
  - Per-layer extension schemas; one model-owned extension is used for the complete model.
  - Changes to vLLM source code or public vLLM interfaces.
  - Real multi-process GPU or NPU communication validation in this PR.

## Implementation Notes

- `AFDConnectorExtension` is defined in the plugin connector base module and exported from `afd_plugin.connectors`.
- `AFDConnectorBase` creates a no-op extension by default. `bind_model()` replaces it only when the loaded model exposes `get_afd_connector_extension()`.
- The extension interface contains four hooks:
  - `control_payload()` produces optional model-owned JSON control data.
  - `update_state_from_control_payload()` applies received control data.
  - `send_attn_extention()` sends model-owned Attention-to-FFN data after the connector's standard payload.
  - `recv_attn_extention()` receives model-owned data and returns the value exposed to FFN model code.
- `AFDControlPayload.extension` transports opaque JSON-compatible control data through the existing plugin-owned serializer.
- `AFDTransferMetadata.extension` carries the opaque data-plane result through the existing FFN forward context, avoiding runner-specific model arguments.
- GPU P2P and synchronous NPU CAMP2P invoke both control-plane and data-plane hooks.
- Asynchronous NPU CAM has no DP metadata control plane and therefore invokes only data-plane hooks.
- No model-specific tensor name, dtype, shape, validation, or buffer policy is added to the common connector layer.
- The implementation remains entirely plugin-owned and does not add compatibility patches or monkey patches.

## Test Plan

Static and CPU-safe unit validation:
Accelerator-gated validation:

- Run GPU P2P Attention/FFN multi-process smoke coverage when a CUDA/NCCL environment is available.
- Run CAMP2P and asynchronous CAM smoke coverage when a torch-npu/CANN environment is available.
- Add model-specific end-to-end coverage together with the first model extension implementation.

## Test Result

- Ruff lint: passed.
- Ruff format check: passed.
- Connector, GPU/NPU runner, and package unit suite: passed; optional accelerator-dependent tests were skipped when their runtime was unavailable.
- Python compilation checks: passed.
- `git diff --check`: passed.
- Real multi-process GPU P2P E2E: not run; no CUDA/NCCL deployment was available in the validation environment.
- Real NPU CAMP2P/Async CAM E2E: not run; no torch-npu/CANN deployment was available in the validation environment.

## Docs Impact

- Files updated: None.
- If none, reason: The public extension contract is documented by English class and method docstrings in the plugin-owned connector modules. The linked RFC provides the design rationale. Model-specific usage documentation should be added with the first model extension implementation.

---

<details>

- [x] Scope is bounded.

</details>

### DIFF
--- a/afd_plugin/connectors/base.py
+++ b/afd_plugin/connectors/base.py
@@ -51,7 +51,7 @@ class AFDConnectorExtension:
     ) -> None:
         pass
 
-    def send_attn_output(
+    def send_attn_extention(
         self,
         connector: AFDConnectorBase,
         context: AFDTransferContext,
@@ -59,7 +59,7 @@ class AFDConnectorExtension:
     ) -> None:
         pass
 
-    def recv_attn_output(
+    def recv_attn_extention(
         self,
         connector: AFDConnectorBase,
         ubatch_idx: int,
--- a/afd_plugin/connectors/base.py
+++ b/afd_plugin/connectors/base.py
@@ -34,6 +34,39 @@ class ConnectorExtraInfo:
         return {}
 
 
+class AFDConnectorExtension:
+    """No-op model extension for connector control and data planes.
+
+    A model may return a subclass from ``get_afd_connector_extension``. The
+    connector transports its control payload and delegates model-specific data.
+    """
+
+    def control_payload(self) -> dict[str, Any] | None:
+        return None
+
+    def update_state_from_control_payload(
+        self,
+        connector: AFDConnectorBase,
+        payload: dict[str, Any] | None,
+    ) -> None:
+        pass
+
+    def send_attn_output(
+        self,
+        connector: AFDConnectorBase,
+        context: AFDTransferContext,
+        **kwargs: Any,
+    ) -> None:
+        pass
+
+    def recv_attn_output(
+        self,
+        connector: AFDConnectorBase,
+        ubatch_idx: int,
+    ) -> object | None:
+        return None
+
+
 class AFDConnectorBase(ABC):
     """Base class for plugin-owned AFD Attention/FFN connectors.
 
@@ -105,10 +138,17 @@ class AFDConnectorBase(ABC):
         self.extra_info = self.parse_extra_config(
             connector_extra_config_from_source(vllm_config),
         )
+        self.extension = AFDConnectorExtension()
 
     # ==============================
     # Lifecycle methods
     # ==============================
+
+    def bind_model(self, model: Any) -> None:
+        """Bind an optional model-owned connector extension."""
+        extension_factory = getattr(model, "get_afd_connector_extension", None)
+        if extension_factory is not None:
+            self.extension = extension_factory()
 
     @abstractmethod
     def close(self) -> None:
@@ -307,4 +347,9 @@ class AFDControlPlane(ABC):
         raise NotImplementedError
 
 
-__all__ = ["AFDConnectorBase", "AFDControlPlane", "ConnectorExtraInfo"]
+__all__ = [
+    "AFDConnectorBase",
+    "AFDConnectorExtension",
+    "AFDControlPlane",
+    "ConnectorExtraInfo",
+]
--- a/afd_plugin/connectors/gpu/p2p.py
+++ b/afd_plugin/connectors/gpu/p2p.py
@@ -352,6 +352,7 @@ class P2pNcclAFDConnector(AFDConnectorBase):
                 self.a2e_group,
                 self.a2e_comm_id,
             )
+        self.extension.send_attn_output(self, context, **kwargs)
 
     def recv_ffn_output(
         self,
@@ -479,6 +480,7 @@ class P2pNcclAFDConnector(AFDConnectorBase):
             stage_idx=ubatch_idx,
             seq_lens=[tensor.shape[0] for tensor in hidden_states_list],
         )
+        metadata.extension = self.extension.recv_attn_output(self, ubatch_idx)
         return AFDA2FTransferPayload(
             hidden_states=hidden_states,
             context=AFDTransferContext(
@@ -751,6 +753,10 @@ class P2pNcclAFDControlPlane(AFDControlPlane):
                     dtype=tensor_metadata.dtype,
                     device=tensor_metadata.device,
                 )
+        connector.extension.update_state_from_control_payload(
+            connector,
+            payload.extension,
+        )
 
     def send_dp_metadata_list(
         self,
@@ -767,6 +773,7 @@ class P2pNcclAFDControlPlane(AFDControlPlane):
                 connector loop.
         """
         connector = self.connector
+        payload.extension = connector.extension.control_payload()
         if connector.p2p_pg is None:
             return
         if not (
@@ -946,4 +953,7 @@ def _register_p2p_custom_ops() -> None:
     _AFD_CUSTOM_OPS_REGISTERED = True
 
 
-__all__ = ["P2pNcclAFDConnector", "P2pNcclAFDControlPlane"]
+__all__ = [
+    "P2pNcclAFDConnector",
+    "P2pNcclAFDControlPlane",
+]
--- a/afd_plugin/connectors/gpu/p2p.py
+++ b/afd_plugin/connectors/gpu/p2p.py
@@ -352,7 +352,7 @@ class P2pNcclAFDConnector(AFDConnectorBase):
                 self.a2e_group,
                 self.a2e_comm_id,
             )
-        self.extension.send_attn_output(self, context, **kwargs)
+        self.extension.send_attn_extention(self, context, **kwargs)
 
     def recv_ffn_output(
         self,
@@ -480,7 +480,7 @@ class P2pNcclAFDConnector(AFDConnectorBase):
             stage_idx=ubatch_idx,
             seq_lens=[tensor.shape[0] for tensor in hidden_states_list],
         )
-        metadata.extension = self.extension.recv_attn_output(self, ubatch_idx)
+        metadata.extension = self.extension.recv_attn_extention(self, ubatch_idx)
         return AFDA2FTransferPayload(
             hidden_states=hidden_states,
             context=AFDTransferContext(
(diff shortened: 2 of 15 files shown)


afd_plugin/connectors/base.py
```diff
@@ -34,6 +34,39 @@ def to_mapping(self) -> dict[str, Any]:
...
+        return None
+
+    def update_state_from_control_payload(
+        self,
+        connector: AFDConnectorBase,
+        payload: dict[str, Any] | None,
+    ) -> None:
+        pass
+
+    def send_attn_output(
```
specture724: send_attn_extention

afd_plugin/connectors/base.py
```diff
@@ -34,6 +34,39 @@ def to_mapping(self) -> dict[str, Any]:
...
+        self,
+        connector: AFDConnectorBase,
+        context: AFDTransferContext,
+        **kwargs: Any,
+    ) -> None:
+        pass
+
+    def recv_attn_output(
```
specture724: recv_attn_extention

specture724:
cc @jiangkuaixue123

afd_plugin/connectors/base.py
lirx-pd: Already changed

afd_plugin/connectors/base.py
lirx-pd: Already changed

afd_plugin/connectors/base.py
```diff
@@ -34,6 +34,39 @@ def to_mapping(self) -> dict[str, Any]:
         return {}
 
 
+class AFDConnectorExtension:
```
specture724, line 37: Will `update_state_from_control_payload`, `send_attn_extention` and `recv_attn_extention` be the same logic to different models? If so, just make the extension dict into the payload?

afd_plugin/connectors/base.py
```diff
@@ -34,6 +34,39 @@ def to_mapping(self) -> dict[str, Any]:
...
+        return None
+
+    def update_state_from_control_payload(
+        self,
+        connector: AFDConnectorBase,
+        payload: dict[str, Any] | None,
+    ) -> None:
+        pass
+
+    def send_attn_extention(
```
jiangkuaixue123, line 54: **Typo: `extention` → `extension`**

`send_attn_extention` (and `recv_attn_extention` below) misspell "extension". These are public API names on the extension contract, so the typo will be locked in once a model implementation ships against them. Renaming now is the cheapest point — this is the first and only release of the interface.

Beyond the spelling, the two hooks are also asymmetric:
- `send_attn_extention(connector, context, **kwargs)` gets the full forward context and call kwargs, so a model can read `input_ids` from `kwargs` and send it.
- `recv_attn_extention(connector, ubatch_idx)` gets only `ubatch_idx` — no `context`, no way to receive anything the send side transmitted.

For a DeepSeek-V4-style `input_ids` extension this means the receive hook has no channel to read the data back from. Consider unifying the signatures (e.g. give `recv_attn_extention` the same `(connector, context, **kwargs)` shape) and defining the actual wire transport for the data-plane payload.

afd_plugin/connectors/base.py
```diff
@@ -34,6 +34,39 @@ def to_mapping(self) -> dict[str, Any]:
...
+        self,
+        connector: AFDConnectorBase,
+        context: AFDTransferContext,
+        **kwargs: Any,
+    ) -> None:
+        pass
+
+    def recv_attn_extention(
```
jiangkuaixue123, line 62: **Typo + `object` return type**

Same `extention` typo as `send_attn_extention`. Two additional points:

1. **Return type `object | None`** violates the project convention to avoid `object`/`Any` in public signatures (see AGENTS.md). The result is consumed as `metadata.extension = self.extension.recv_attn_extention(...)`, so a concrete protocol or at least a named type alias would be preferable.
2. **No transport channel for the data-plane payload.** `AFDTransferMetadata.extension` only carries the *receive-side* result across processes within the same worker (hidden-states transfer), but nothing moves the *send-side* data (`input_ids` from the Attention forward kwargs) to the FFN process. The send hook writes into `**kwargs` which never crosses a process boundary. If the intended DeepSeek-V4 use case is to transport `input_ids` Attention→FFN, this hook needs a real wire path (e.g. piggyback on the existing P2P tensor send, like `router_logits`), not a process-local return value.

afd_plugin/connectors/npu/camp2p.py
```diff
@@ -638,12 +640,17 @@ def update_state_from_dp_metadata(
...
+            connector,
+            payload.extension,
+        )
 
     def send_dp_metadata_list(
         self,
         payload: AFDControlPayload,
     ) -> None:
         connector = self.connector
+        payload.extension = connector.extension.control_payload()
```
jiangkuaixue123, line 653: **Control-plane callback fires on the wrong process (asymmetric)**

`update_state_from_dp_metadata` is invoked on **both** sides:
- **Attention side** (`npu/attention_model_runner.py:1384` calls `update_state_from_dp_metadata(payload)` then `send_dp_metadata_list(payload)`): applies the *Attention* process's own bound extension.
- **FFN side** (`npu/ffn_worker.py:142` `recv_dp_metadata_list()` → `execute_ffn_step` → `update_state_from_dp_metadata`): applies the *FFN* process's own bound extension.

So the `payload.extension` written by the send side is consumed by a **different process's** extension instance. If the FFN-side model does not expose `get_afd_connector_extension()` (a DeepSeek-V4-style extension may only be meaningful on the Attention side, which needs to send `input_ids`), the FFN side binds a no-op extension and `payload.extension` crosses the wire unread — while the Attention side already applied it *locally* before sending, so the send side never actually transmits its own payload.

Suggested fix: make the semantics symmetric — `send_dp_metadata_list` only **writes** `payload.extension`; the apply path (`update_state_from_control_payload`) should run only on the connector that actually **receives** (i.e. after `recv_dp_metadata_list()` decodes on the FFN side). The integration test `test_model_extension_owns_control_and_data_plane_payloads` shares one extension instance across both connectors, which masks this cross-process gap; a test with two independently-bound extensions would expose it.

afd_plugin/connectors/gpu/p2p.py
```diff
@@ -767,6 +773,7 @@ def send_dp_metadata_list(
                 connector loop.
         """
         connector = self.connector
+        payload.extension = connector.extension.control_payload()
```
jiangkuaixue123, line 776: **`payload.extension` overwrite semantics**

`send_dp_metadata_list` unconditionally overwrites `payload.extension` with `connector.extension.control_payload()`. Two consequences:

1. **Silent clobber**: if the caller constructed the payload with a pre-set `extension`, it is overwritten with no warning. `_send_dp_metadata` (`attention_model_runner.py:144`) always builds the payload fresh, so today this is benign — but the write-then-read pattern here invites the field being set from two places.
2. **No-op still writes `None`**: when the model has no extension, `control_payload()` returns `None`, and this line writes `payload.extension = None`. Combined with the symmetric apply suggestion, it would be cleaner to leave the field untouched when there is no extension (skip assignment), so `payload.extension` remains `None` from construction rather than an explicit overwrite.

Recommendation: move the extension filling into the wire-serialization boundary (`encode_control_payload`) or guard the assignment, so the control-plane extension has exactly one writer with clear ownership.

afd_plugin/connectors/npu/camp2p.py
```diff
@@ -638,12 +640,17 @@ def update_state_from_dp_metadata(
         connector.dp_metadata_list = payload.dp_metadata_list
         connector.is_graph_capturing = payload.is_graph_capturing
         connector.is_warmup = payload.is_warmup
+        connector.extension.update_state_from_control_payload(
+            connector,
+            payload.extension,
```
jiangkuaixue123, line 645: **Duplicate of the P2P control-plane asymmetry (camp2p variant)**

Same issue as my comment on `gpu/p2p.py:756`. In the camp2p path, `recv_dp_metadata_list()` (`camp2p.py:661`) decodes the wire payload but applies the extension callback only later via `update_state_from_dp_metadata`, which runs on whichever process calls it. If the symmetric semantics from the other comment are adopted, this apply should be moved to the recv path (decode → apply the *local* FFN-side extension), and the Attention-side `update_state_from_dp_metadata` (called from `npu/attention_model_runner.py:1384`) should not apply the extension it is about to send.

Keeping the two connector implementations consistent on the same control-plane contract matters here — P2P and camp2p are both control-plane-driven connectors, and diverging apply semantics between them would be a maintenance trap.

PR #213, opened 2026-08-07
fix(npu): skip empty Async CAM FFN work

### LINKED ISSUE #210: [Bug] Async CAM FFN crashes on ranks with zero shared-expert tokens
## Summary

On the Ascend Async CAM FFN path, a rank can legitimately receive zero shared-expert tokens from `async_dispatch_recv`. The connector slices `expand_x_shared` and `dynamic_scales_shared` to zero rows, but the DeepSeek attention-gate MoE implementation still invokes the W8A8 shared-expert path. This passes an empty INT8 tensor to `torch_npu.npu_quant_matmul`, which fails on NPU.

## Failure path

```text
CAM async_dispatch_recv
  -> expert_token_nums_shared[0] == 0
  -> CAMAsyncAFDConnector.recv_ffn_work_item
  -> expand_x_shared = expand_x_shared[:0]
  -> FFNModelRunner.compute_ffn_output
  -> compute_attention_gate_moe_ffn
  -> _compute_w8a8_shared_experts_from_int8
  -> torch_npu.npu_quant_matmul(empty_tensor, ...)
```

The relevant shared-expert branch currently checks only whether the model owns shared experts:

```python
if experts._shared_experts is not None:
    ...
```

That is a model capability check, not a batch-local work check. It does not account for a rank receiving zero shared tokens in the current dispatch.

## Regression history

The missing zero-row guard predates the vLLM/vLLM-Ascend 0.26 upgrade. Commit `c88babf` exposed it by correctly preserving the CAM-reported shared-token count (`max(0, count)`) and removing the previous fake-token slicing workaround. Restoring `max(1, count)` or the old fixed-size fallback would hide the crash by computing invalid placeholder tokens and should not be the fix.

This is present on `main` and is independent of TP/SP async MoE ubatching.

## Expected behavior

When the current FFN rank has zero shared-expert tokens:

- skip all shared-expert GEMM/quantized matmul work;
- keep `shared_output` absent so the existing CAM send placeholder path is used;
- preserve the real zero token count reported by CAM.

The routed-token path should also be audited for the symmetric zero-token case, because it can currently pass a zero-row tensor to the routed MoE implementation.

## Test coverage

Add coverage for the four rank-local work combinations:

- routed > 0, shared > 0;
- routed > 0, shared = 0;
- routed = 0, shared > 0;
- routed = 0, shared = 0.

The observed failure is the second case.

ShwStone:
## Summary

- skip shared-expert GEMMs when CAM assigns zero shared tokens to the current FFN rank
- skip `unified_apply_mlp` when the rank has zero routed tokens
- preserve CAM's exact zero token counts and leave placeholder handling to the existing connector send path
- cover all four rank-local routed/shared token combinations in a parameterized regression test

## Root cause

The FFN computation checked whether the model owns shared experts, but not whether the current rank received shared-expert work for this dispatch. A zero-row INT8 `expand_x_shared` tensor therefore reached `torch_npu.npu_quant_matmul`, which rejects empty inputs.

The routed path had the symmetric risk of passing an empty tensor to `unified_apply_mlp`. Both kernels are now bypassed only for zero-row rank-local inputs; non-empty behavior is unchanged. This deliberately does not restore the old fake-token compute workaround.

## Validation

- `python -m compileall -q afd_plugin/model_executor/models/npu/deepseek_v2_attention_gate.py tests/unit/model_executor/models/test_forward_context.py`
- `ruff check afd_plugin/model_executor/models/npu/deepseek_v2_attention_gate.py tests/unit/model_executor/models/test_forward_context.py`
- `git diff --check upstream/main...HEAD`
- the target pytest module could not be collected locally because the available environment lacks torch/vLLM; the regression test still requires execution in the pinned NPU test environment

Fixes #210

### DIFF
--- a/afd_plugin/model_executor/models/npu/deepseek_v2_attention_gate.py
+++ b/afd_plugin/model_executor/models/npu/deepseek_v2_attention_gate.py
@@ -167,40 +167,51 @@ def compute_attention_gate_moe_ffn(
         and _gmmswigluquant_fusion_enabled()
     )
 
+    # CAM reports exact rank-local work. An EP rank can legitimately receive
+    # zero routed or shared tokens, while Ascend MoE kernels require non-empty
+    # inputs.
     shared_output = None
     if experts._shared_experts is not None:
+        if expand_x_shared is None:
+            raise RuntimeError(
+                "AFD shared experts require expand_x_shared from CAM dispatch",
+            )
         shared_input = expand_x_shared
         shared_scales = dynamic_scales_shared
-        if shared_input.dtype == torch.int8 and quant_type == QuantType.W8A8:
-            shared_output = _compute_w8a8_shared_experts_from_int8(
-                experts._shared_experts,
-                shared_input,
-                shared_scales,
-                output_dtype=torch.bfloat16,
-            )
-        else:
-            shared_input = _dequantize_int8_activation(
-                shared_input,
-                shared_scales,
-                output_dtype=torch.bfloat16,
-            )
-            shared_output = experts._shared_experts(shared_input)
+        if shared_input.shape[0] > 0:
+            if shared_input.dtype == torch.int8 and quant_type == QuantType.W8A8:
+                shared_output = _compute_w8a8_shared_experts_from_int8(
+                    experts._shared_experts,
+                    shared_input,
+                    shared_scales,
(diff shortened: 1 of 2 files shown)


jiangkuaixue123:
LGTM

Issue #215, opened 2026-08-08
[RFC]: Add Qwen3 MoE CUDA AFD support on vLLM 0.26.0


sinpeyw:
## Motivation

Add a CUDA adaptation path for vLLM 0.26.0's native `Qwen3MoeForCausalLM` architecture. The initial scope covers:

- `Qwen/Qwen3-30B-A3B`
- `Qwen/Qwen3-30B-A3B-Instruct-2507`
- `Qwen/Qwen3-30B-A3B-Thinking-2507`
- `Qwen/Qwen3-Coder-30B-A3B-Instruct`
- `Qwen/Qwen3-235B-A22B`

Their native decoder invokes the complete `mlp(hidden_states)` stage after Attention, residual handling, and post-Attention normalization. AFD can split at that boundary while retaining native Qwen forward and weight-loading semantics and the existing connector contract.

### Initial scope

- vLLM: exactly `0.26.0`
- runtime: NVIDIA CUDA, single host, vLLM V1
- connector: synchronous `P2pNcclAFDConnector`
- gate placement: `compute_gate_on_attention=false`
- weights: unquantized BF16 (`quantization=None`)
- modes: eager, `FULL_DECODE_ONLY` CUDA graph, and native two-ubatch DBO
- 30B topologies: 1A1F, 2A2F, 2A1F, TP=1, and TP=2
- 235B validation topology: 4A4F with TP=4/EP=4
- validated maximum model length: 1024 tokens

### Plugin boundary

1. Register `AFDQwen3MoeForCausalLM` under an AFD-only alias without replacing vLLM's native registration.
2. Subclass native `Qwen3MoeModel` and inject a role-aware decoder through its `decoder_layer_type` hook; inherit native decoder/model forward behavior.
3. On Attention, construct native Attention and decoder norms and use the existing parameter-free `RemoteFFNProxy` at `.mlp`.
4. On FFN, construct the complete native dense MLP or `Qwen3MoeSparseMoeBlock`, including routing and experts.
5. Filter checkpoint paths once by role and delegate mapping, packing, and loaded-parameter results to the native loader.

No Qwen-specific connector, transport payload, distributed loop, FusedMoE, MoERunner, quantization, or kernel change is proposed.

### Explicit non-goals

- dense Qwen3, Qwen3-Next, Qwen3-VL, Qwen3.5, or Qwen3.6 architectures
- NPU, ROCm, MUSA, multi-node, or asynchronous connectors
[...]

PR #181, opened 2026-08-01
Add Qwen3.6 MoE CUDA AFD implementation support

shanyulu:
## Purpose

Qwen3.6-35B-A3B CUDA AFD implementation support on vLLM 0.26.

## Scope

In scope:

- CUDA, V1, synchronous NCCL P2P, and text-only Qwen3.5/3.6 MoE;
- AFD-only registration and native non-AFD registration preservation;
- role-aware construction and role-filtered native checkpoint loading;
- FFN-local native gate/router/expert execution;
- `hf_text_config` compatibility, direct pinned-vLLM API access, and fail-closed unsupported configurations;
- explicit CUDA-only Qwen3.5/3.6 backend resolution while retaining DeepSeek GPU/NPU registration behavior;
- coexistence with merged Qwen3 MoE support from #216 as a separate native architecture and adapter.

Out of scope:

- multimodal execution;
- `compute_gate_on_attention=true` and Attention-side routing transport for Qwen3.5/3.6;
- NPU Qwen3.5/3.6, DBO implementation, DP/EP/PP expansion, async communication, multi-node, quantization, and performance;
- checked-in Qwen3.5/3.6 E2E/capability infrastructure;
- README support-status changes.

## Implementation

The Qwen3.5/3.6 adapter retains native Qwen forward and loader lifecycles. Attention owns the text model embedding, norms, and attention state, then sends hidden states. FFN owns the native gate, routed experts, shared expert, and shared-expert gate. Qwen3.5/3.6 rejects `compute_gate_on_attention=true` and `pipeline_parallel_size != 1` during model construction before visual/model construction. Non-text-only execution is rejected before visual construction. CUDA Qwen3.5/3.6 resolves to its AFD alias; NPU Qwen3.5/3.6 fails during model-config resolution before model loading.

The rebased registration preserves #216's `Qwen3MoeForCausalLM` adapter unchanged and adds `Qwen3_5MoeForConditionalGeneration` as a distinct model family.

## Validation

Rebased on current `main` `a04d9da356eedfdfbd7a062054173eeae11e497b`. Final head: `9939ab22396fcb8a0b2a83424974f7305c8961fb`.

- static: `ruff check`, `ruff format --check`, `compileall`, `pip check`, and `git diff --check` passed;
- focused coexistence: 115 passed, 0 failed;
- full CPU: 577 passed, 48 skipped, 31 deselected, 0 failed;
- independent exact-final-SHA clean checkout: same static / focused / full CPU results.

[...]

Issue #179, opened 2026-07-30
[RFC] Experimental Qwen3.5 MoE AFD support: CUDA correctness baseline


shanyulu:
## Summary

This RFC proposes an **experimental, CUDA-only Qwen3.5 MoE AFD validation lane** for the AFD plugin.

The purpose is to establish a reviewed model/architecture baseline before continuing the MUSA work proposed in #173. It is **not** a request to change the project-wide vLLM alignment policy in #155, and it does not claim production readiness or performance improvement.

A local CUDA implementation has completed deterministic correctness validation for **Qwen/Qwen3.5-35B-A3B** on vLLM 0.23.0. The implementation follows native vLLM model/decoder/weight-loading lifecycles and disaggregates only at the MoE experts boundary.

## Why an RFC before an implementation PR?

Qwen3.5 is a hybrid full-attention + linear-attention/Mamba MoE model. AFD must keep attention/KV and hybrid state on the Attention role while moving only routed/shared expert execution to the FFN role.

Maintainer feedback is requested now on:

1. whether Qwen3.5 belongs in the AFD model-support scope;
2. whether an eager, text-only correctness-first contribution is an acceptable first step;
3. whether the Attention-side router boundary is the preferred design; and
4. whether this experimental vLLM 0.23 validation should be treated solely as an architecture/evidence lane while the project follows the alignment policy in #155.

## Proposed initial scope

- Model: `Qwen/Qwen3.5-35B-A3B`, original BF16 checkpoint.
- Backend: NVIDIA CUDA only.
- Topology: single host, Attention GPUs 0–1 / TP=2; FFN GPUs 2–3 / TP=2.
- Runtime: vLLM V1, BF16, eager execution, synchronous NCCL/P2P.
- Routing: gate on Attention; hidden states plus router logits transferred to FFN.
- Correctness only: no performance claim.

### Explicit non-goals

- no CUDA Graph, DBO, async communication, microbatching, EP/SP/PP, multi-node, quantisation, or performance tuning;
- no multimodal image/video support;
- no claim of broad Qwen or backend support;
- no MUSA implementation in this contribution.

MUSA remains a separate follow-up track under #173, where the CUDA result can serve as the model-boundary and correctness reference.

## Proposed plugin boundary

The proposed adapter keeps native vLLM Qwen control flow and loading authoritative:

```text
native Qwen decoder
  -> native Attention / linear-attention state
  -> native Qwen router on Attention
  -> parameter-free remote-experts proxy
[...]

PR #216, opened 2026-08-08
[Feature] Add Qwen3 MoE CUDA AFD support

sinpeyw:
## Purpose

Add CUDA AFD support for vLLM 0.26.0's native `Qwen3MoeForCausalLM` architecture, covering:

- `Qwen/Qwen3-30B-A3B`
- `Qwen/Qwen3-30B-A3B-Instruct-2507`
- `Qwen/Qwen3-30B-A3B-Thinking-2507`
- `Qwen/Qwen3-Coder-30B-A3B-Instruct`
- `Qwen/Qwen3-235B-A22B`

The adapter splits at the native complete-MLP boundary, keeps routing and experts together on FFN, and preserves the native Qwen forward and weight-loading lifecycle. It does not reuse the Qwen3.5/3.6 hybrid-state and router-logits transport introduced by the separate work discussed in #179 and draft PR #181.

## Issue

- Related issue: #215
- Closing keyword: none; keep the RFC open through implementation review

## Scope

- In scope:
  - CUDA, single-host vLLM V1, and synchronous `P2pNcclAFDConnector`;
  - AFD-only Qwen3 MoE registration and role-aware module construction;
  - one-pass role-filtered checkpoint loading through the native loader;
  - eager, `FULL_DECODE_ONLY` CUDA graph, and native two-ubatch DBO;
  - 1A1F, 2A2F, 2A1F, TP=1/2, and a 235B 4A4F/TP4/EP4 validation profile.
- Out of scope:
  - Qwen3.5/3.6, dense Qwen3, Qwen3-Next, and Qwen3-VL architectures;
  - NPU, ROCm, MUSA, multi-node, and asynchronous connectors;
  - Attention-side routing or a new connector/wire payload;
  - sequence-parallel MoE, EPLB, pipeline parallelism, speculative decoding, MTP/Eagle, LoRA, quantized checkpoints, long-context, or performance claims.

## Implementation Notes

- Register `AFDQwen3MoeForCausalLM` under an AFD-only alias; native `Qwen3MoeForCausalLM` registration remains unchanged.
- Inject `AFDQwen3MoeDecoderLayer` through the native `Qwen3MoeModel.decoder_layer_type` hook and inherit native decoder/model forward behavior.
- Attention constructs native Attention and norms plus the existing parameter-free `RemoteFFNProxy`; FFN constructs the complete native dense MLP or `Qwen3MoeSparseMoeBlock`.
- Filter checkpoint paths once by role, retaining native parameter names, packed mappings, TP/EP behavior, and `AutoWeightsLoader` results.
- No vLLM source file, connector contract, communication payload, FusedMoE, MoERunner, quantization path, or kernel is changed.

## Test Plan

Static and CPU contracts:

```bash
uv run ruff check .
uv run ruff format --check .
git diff --check
uv run pytest -q tests/unit
```

The GPU validation evidence below was collected before the checked-in E2E harness was split from this model-only PR. It covers a 10-case matrix with eager, 64-request/concurrency-32 pressure, CUDA graph, DBO, TP=2, 2A2F, and 2A1F. Additional controlled checks compare native TP1 with AFD 1A1F on Qwen3-30B-A3B BF16 and exercise the 235B profile as 4A4F/TP4/EP4.

## Test Result

- GPU validation source: `b5e91a8184af9e45e05d1175c3485d239a32625b`
- Rebased model-only source: `0b9d2dd856b57ca999074be7104d77686f6fd84c`
- Base: `upstream/main@18197b3b36aed7868cfac5b092545d862140b945`

Environment:

- Python 3.12.12
- PyTorch 2.11.0+cu129
- vLLM 0.26.0
- 8 x NVIDIA A800-SXM4-80GB, driver 550.54.14

| Gate | Result |
| --- | --- |
| Ruff check / format / diff check | PASS |
| Full CPU unit suite on the rebased model-only source | PASS |
| Focused registration, construction, and weight contracts | PASS, 47/47 |
| 30B GPU validation matrix | PASS, 10/10 for Base with the strengthened runtime assertions; Instruct-2507, Thinking-2507, and Coder identities passed the earlier pre-review 10-case matrix |
| Qwen3-30B-A3B BF16 oracle | PASS for English, Chinese, code, and arithmetic prompts; identical generated token IDs and top-5 membership, maximum sampled/top-5 logprob error `0.0` |
| Pressure | PASS, 64/64 requests at concurrency 32 |
| Qwen3-235B-A22B execution | PASS, 4A4F/TP4/EP4 eager, four concurrent requests |
| Cleanup | PASS, no AFD/vLLM worker, listening port, or test-owned GPU allocation remained |

This is correctness and execution evidence, not a throughput or production-readiness claim.

## Docs Impact

- Updated `README.md` model support and known limitations.
- Updated model-integration documentation.

---

<details>

- [x] Purpose is clear and linked to public context.
- [x] Scope is bounded.
- [x] Plugin-owned classes and explicit dotted class paths are used.
- [x] Imports remain CPU-safe; CUDA execution is GPU-gated.
- [x] Rebased CPU and pre-rebase GPU validation evidence is attached.

</details>

### DIFF
--- a/afd_plugin/model_executor/models/qwen3_moe.py
+++ b/afd_plugin/model_executor/models/qwen3_moe.py
@@ -1,0 +1,325 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
+"""Qwen3 MoE AFD model wrapper for CUDA execution.
+
+The wrapper keeps the native vLLM Qwen3 MoE lifecycle and injects a
+role-aware decoder layer through the native ``decoder_layer_type`` hook.
+Attention executes the native Attention, residual, and normalization path;
+FFN executes the complete native dense MLP or MoE block.
+"""
+
+from collections.abc import Iterable, Iterator
+
+import torch
+import torch.nn as nn
+from transformers import Qwen3MoeConfig
+from vllm.config import VllmConfig
+from vllm.model_executor.models import qwen3_moe as native
+from vllm.platforms import current_platform
+
+from afd_plugin.config import AFDConfig, parse_afd_config
+from afd_plugin.model_executor.models.deepseek_v2 import RemoteFFNProxy
+
+_ATTENTION_ROLE = frozenset(("attention",))
+_FFN_ROLE = frozenset(("ffn",))
+_BOTH_ROLES = frozenset(("attention", "ffn"))
+
+
+def _is_moe_layer(config: Qwen3MoeConfig, layer_idx: int) -> bool:
+    """Return whether the native Qwen3 schedule selects MoE for one layer."""
+    return (
+        layer_idx not in config.mlp_only_layers
+        and config.num_experts > 0
+        and (layer_idx + 1) % config.decoder_sparse_step == 0
+    )
+
+
+def _weight_layer_path(name: str) -> tuple[int, str] | None:
+    """Return ``(layer index, stage)`` for a native decoder weight path."""
+    parts = name.split(".")
+    for marker_idx, part in enumerate(parts[:-2]):
+        if part != "layers":
+            continue
+        try:
+            layer_idx = int(parts[marker_idx + 1])
+        except ValueError:
+            continue
+        return layer_idx, parts[marker_idx + 2]
+    return None
+
+
+def _checkpoint_weight_roles(name: str) -> frozenset[str]:
+    """Classify one native checkpoint path by its AFD execution owner."""
+    layer_path = _weight_layer_path(name)
+    if layer_path is None:
+        return _BOTH_ROLES
+
+    _layer_idx, stage = layer_path
+    if stage == "mlp":
+        return _FFN_ROLE
+    if stage in {"self_attn", "input_layernorm", "post_attention_layernorm"}:
+        return _ATTENTION_ROLE
+    return _BOTH_ROLES
+
+
+def _iter_role_weights(
+    weights: Iterable[tuple[str, torch.Tensor]],
+    *,
+    role: str,
+) -> Iterator[tuple[str, torch.Tensor]]:
+    """Consume a checkpoint iterator once and retain this role's paths."""
+    for name, loaded_weight in weights:
+        if role in _checkpoint_weight_roles(name):
+            yield name, loaded_weight
+
+
+def _validate_supported_config(
+    vllm_config: VllmConfig,
+    afd_config: AFDConfig,
+) -> None:
+    """Reject Qwen3 MoE modes not covered by the initial CUDA contract."""
+    parallel_config = vllm_config.parallel_config
+    if current_platform.device_type != "cuda":
+        raise RuntimeError("AFD Qwen3 MoE currently supports CUDA only")
+    if afd_config.compute_gate_on_attention:
+        raise RuntimeError(
+            "AFD Qwen3 MoE requires compute_gate_on_attention=false",
+        )
+    if parallel_config.use_sequence_parallel_moe:
+        raise RuntimeError(
+            "AFD Qwen3 MoE does not support sequence-parallel MoE",
+        )
+    if parallel_config.enable_eplb:
+        raise RuntimeError("AFD Qwen3 MoE does not support EPLB")
+    if parallel_config.pipeline_parallel_size != 1:
+        raise RuntimeError("AFD Qwen3 MoE does not support pipeline parallelism")
+    if vllm_config.speculative_config is not None:
+        raise RuntimeError("AFD Qwen3 MoE does not support speculative decoding")
+    if vllm_config.lora_config is not None:
+        raise RuntimeError("AFD Qwen3 MoE does not support LoRA")
+
+
+class AFDQwen3MoeDecoderLayer(native.Qwen3MoeDecoderLayer):
+    """Qwen3 MoE decoder layer with separable Attention and FFN execution."""
+
+    # Patch reason: native Qwen3 MoE constructs both Attention and FFN modules.
+    # Patch functionality: construct only the large modules owned by the AFD role.
+    # Signature: matches upstream; no added parameters.
+    # Upstream: vLLM v0.26.0, vllm/model_executor/models/qwen3_moe.py
+    # Commit: 568afb3a13806beb53bb2e6bd518269357b237c0
+    def __init__(self, vllm_config: VllmConfig, prefix: str = "") -> None:
+        # ### PATCH START: initialize the role-aware layer without native allocation.
+        nn.Module.__init__(self)
+        afd_config = parse_afd_config(vllm_config, validate=False)
+        self.afd_role = afd_config.role
+        # ### PATCH END: initialize the role-aware layer without native allocation.
+
+        config = vllm_config.model_config.hf_text_config
+        cache_config = vllm_config.cache_config
+        quant_config = vllm_config.quant_config
+
+        self.hidden_size = config.hidden_size
+        max_position_embeddings = getattr(config, "max_position_embeddings", 8192)
+        dual_chunk_attention_config = getattr(
+            config,
+            "dual_chunk_attention_config",
+            None,
+        )
+        # ### PATCH START: retain the native layer schedule for AFD role dispatch.
+        layer_idx = native.extract_layer_index(prefix)
+        self.layer_idx = layer_idx
+        self.is_moe_layer = _is_moe_layer(config, layer_idx)
+        # ### PATCH END: retain the native layer schedule for AFD role dispatch.
+
+        # ### PATCH START: allocate only the active role's execution stage.
+        if self.afd_role == "attention":
+            self.self_attn = native.Qwen3MoeAttention(
+                hidden_size=self.hidden_size,
+                num_heads=config.num_attention_heads,
+                num_kv_heads=config.num_key_value_heads,
+                rope_parameters=config.rope_parameters,
+                max_position_embeddings=max_position_embeddings,
+                rms_norm_eps=config.rms_norm_eps,
+                qkv_bias=getattr(config, "attention_bias", False),
+                head_dim=getattr(config, "head_dim", None),
+                cache_config=cache_config,
+                quant_config=quant_config,
+                prefix=f"{prefix}.self_attn",
+                dual_chunk_attention_config=dual_chunk_attention_config,
+            )
+            self.mlp = RemoteFFNProxy(layer_idx=layer_idx)
+            self.input_layernorm = native.RMSNorm(
+                config.hidden_size,
+                eps=config.rms_norm_eps,
+            )
+            self.post_attention_layernorm = native.RMSNorm(
+                config.hidden_size,
+                eps=config.rms_norm_eps,
+            )
+        else:
+            self.self_attn = native.PPMissingLayer()
+            if self.is_moe_layer:
+                self.mlp = native.Qwen3MoeSparseMoeBlock(
+                    vllm_config=vllm_config,
+                    prefix=f"{prefix}.mlp",
+                )
+            else:
+                self.mlp = native.Qwen3MoeMLP(
+                    hidden_size=config.hidden_size,
+                    intermediate_size=config.intermediate_size,
+                    hidden_act=config.hidden_act,
+                    quant_config=quant_config,
+                    prefix=f"{prefix}.mlp",
+                )
+            self.input_layernorm = native.PPMissingLayer()
+            self.post_attention_layernorm = native.PPMissingLayer()
+        # ### PATCH END: allocate only the active role's execution stage.
+
+    def compute_ffn_output(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        """Execute this layer's complete native FFN block on the FFN role."""
+        if self.afd_role != "ffn":
+            raise RuntimeError("Qwen3 MoE FFN compute requires the AFD FFN role")
+        return self.mlp(hidden_states)
+
+
+@native.support_torch_compile
+class AFDQwen3MoeModel(native.Qwen3MoeModel):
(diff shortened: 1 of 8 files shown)


hsliuustc0106:
is there any perf comparison and profiler for compute and network in A/F

jiangkuaixue123:
@yujuancao07 Maybe you can help review the change of e2e-test.

afd_plugin/v1/worker/attention_model_runner.py
```diff
@@ -264,24 +274,39 @@ def _build_attention_metadata(
...
-            logits_indices,
-            use_spec_decode,
-            for_cudagraph_capture,
-            num_scheduled_tokens,
-            cascade_attn_prefix_lens,
-            slot_mappings,
-        )
+        disabled_metadata_builders = {}
+        try:
+            if ubatch_slices is not None and len(ubatch_slices) > 1:
```
jiangkuaixue123, line 290: Could you clarify why this change to `AFDAttentionModelRunner` is necessary? Is disabling `supports_update_block_table` a general DBO compatibility issue introduced by upgrading to vLLM 0.26.0, or is it specific to Qwen3 MoE? If it affects AFD DBO generally, would it make sense to handle it as a separate compatibility fix rather than coupling it to the Qwen3 MoE integration?

jiangkuaixue123 (commented):
Question about the `AFDAttentionModelRunner` compatibility change.

sinpeyw:
> is there any perf comparison and profiler for compute and network in A/F

This PR currently focuses on correctness and does not make performance claims. I’ll follow up with a performance comparison and A/F compute and network profiling results.

afd_plugin/v1/worker/attention_model_runner.py
sinpeyw: You’re right. This is a general vLLM 0.26 DBO metadata-cache issue exposed by Qwen3/FlashAttention, not Qwen3 adapter logic. I’ll move the workaround and its tests into a separate compatibility PR and keep #216 focused on Qwen3 support.

sinpeyw: I’ve moved the compatibility fix into a standalone bugfix PR, #217, with the trade-off against directly backporting vLLM #48659 documented there. I’ll remove the workaround and its generic tests/docs from #216 so this PR remains focused on Qwen3 support, then rebase it after #217 lands

sinpeyw: Done in 93cc04e. No longer contains the DBO workaround or its generic tests/docs; the compatibility fix is isolated in #217.

sinpeyw:
### Qwen3-30B-A3B CUDA AFD throughput comparison

Following up on [the performance and A/F profiling question](https://github.com/vllm-project/afd-plugin/pull/216#issuecomment-5226319066), this update provides the end-to-end throughput comparison.

I reproduced the synchronous decode experiment from the [AFD performance blog](https://vllm.ai/blog/2026-07-23-vllm-afd-plugin) at an eight-GPU scale using Qwen3-30B-A3B on NVIDIA A800 GPUs with vLLM 0.26.0. The tested code combines #216, #217, and #219.

Native DP8/EP8 and 6A2F with A16/rank are the scaled counterparts of the published EP64 and 48A16F cases. They preserve the 3:1 Attention-to-FFN ratio, one-quarter active expert pool, per-Attention-rank batch sizes (16/28 for 16K and 8/14 for 32K), fixed input lengths, 512–1,536-token uniform outputs, forced balanced routing, and DBO for AFD. A32/rank and A64/rank are additional expert-pool sensitivity cases. The published 64A16F topology was not reproduced because its scaled 8A2F form requires ten GPUs.

Each case used 1,000 requests. Stable throughput is measured over the longest contiguous interval where running requests remain at least 90% of the run's peak, normalized by all eight deployed GPUs. All runs completed with zero failed requests.

| Topology | 16K tok/s/GPU | Change | 32K tok/s/GPU | Change |
| --- | ---: | ---: | ---: | ---: |
| Native DP8/EP8 | 504.21 | baseline | 255.91 | baseline |
| 6A2F, A16/rank | 524.88 | +4.10% | 269.08 | +5.14% |
| 6A2F, A32/rank | 525.30 | +4.18% | 269.18 | +5.19% |
| 6A2F, A64/rank | 373.24 | -25.97% | 189.92 | -25.79% |

image
image

Within this controlled setup, 6A2F with A16/rank and A32/rank consistently outperformed native DP8/EP8 at both input lengths: by 4.10–4.18% at 16K and 5.14–5.19% at 32K.

However, the A16/rank and A32/rank improvements are not sufficient evidence of a general AFD advantage because these AFD cases restrict the active expert pool. The A64/rank regression further illustrates this limitation. Broader topology studies with natural routing are still needed.

Once #216 is merged, I will follow up with a separate PR containing the detailed experimental setup and reproduction instructions.

tests/e2e/models/qwen3_moe/test_e2e_gpu.py
```diff
@@ -0,0 +1,204 @@
...
+        str(afd_port + case_index),
+        "--tp-size",
+        str(case.tp_size),
+        "--startup-timeout",
+        os.environ.get("AFD_GPU_E2E_STARTUP_TIMEOUT", "900"),
+        "--max-tokens",
+        str(case.max_tokens),
+        "--prompt",
+        DEFAULT_PROMPT,
+        "--num-requests",
```
yujuancao07: 2a2f_eager_smoke uses the default requests=1 and concurrency=1. With DP=2, a single request may only exercise one Attention–FFN pair, leaving the second pair untested. Please use at least two concurrent requests for the 2A2F cases.

tests/e2e/models/qwen3_moe/test_e2e_gpu.py
```diff
@@ -0,0 +1,204 @@
...
+    model = os.environ.get("AFD_QWEN3_MOE_E2E_MODEL")
+    if not model:
+        pytest.skip("set AFD_QWEN3_MOE_E2E_MODEL to a Qwen3 MoE model path")
+    return model
+
+
+def _gpu_list() -> list[str]:
+    return [
+        item.strip()
+        for item in os.environ.get("AFD_GPU_E2E_GPUS", "0,1,2,3").split(",")
```
yujuancao07: _gpu_list() defaults to physical GPUs 0,1,2,3 and ignores CUDA_VISIBLE_DEVICES.  Please require an explicit device list or derive it from CUDA_VISIBLE_DEVICES.

tests/e2e/models/qwen3_moe/test_e2e_gpu.py
```diff
@@ -0,0 +1,204 @@
...
+        requests=4,
+        concurrency=4,
+    ),
+)
+
+
+def _model_path() -> str:
+    model = os.environ.get("AFD_QWEN3_MOE_E2E_MODEL")
+    if not model:
+        pytest.skip("set AFD_QWEN3_MOE_E2E_MODEL to a Qwen3 MoE model path")
```
yujuancao07: If AFD_QWEN3_MOE_E2E_MODEL is missing, all Qwen3 tests are skipped and the job can still be green without testing Qwen3 at all.

tests/e2e/models/qwen3_moe/test_e2e_gpu.py
```diff
@@ -0,0 +1,204 @@
...
+            [
+                "--enable-dbo",
+                "--dbo-decode-token-threshold",
+                "1",
+                "--dbo-prefill-token-threshold",
+                "8",
+            ],
+        )
+
+    subprocess.run(command, cwd=REPO_ROOT, check=True)
```
yujuancao07: These cases only check whether the subprocess exits successfully. A silent fallback to eager execution, or DBO running without forming multiple microbatches, would still pass. Please add an assertion based on an explicit runtime signal that confirms Graph replay and DBO execution.

tests/e2e/models/qwen3_moe/test_e2e_gpu.py
sinpeyw: Addressed in a4de153. Every 2A2F case now uses at least four concurrent requests; the pressure case uses 64 requests at concurrency 32.

tests/e2e/models/qwen3_moe/test_e2e_gpu.py
sinpeyw: Addressed in a4de153. AFD_GPU_E2E_GPUS takes precedence, then CUDA_VISIBLE_DEVICES is used; the test fails if neither provides a device allocation.

tests/e2e/models/qwen3_moe/test_e2e_gpu.py
sinpeyw: Addressed in a4de153. Missing AFD_QWEN3_MOE_E2E_MODEL now fails the Qwen3 GPU suite instead of skipping it.

tests/e2e/models/qwen3_moe/test_e2e_gpu.py
sinpeyw: Addressed in a4de153. Graph cases assert a FULL runtime dispatch; the DBO case asserts runtime microbatching and exactly two UBatchSlice entries after API readiness.

afd_plugin/model_executor/models/qwen3_moe.py
```diff
@@ -0,0 +1,325 @@
...
+        # ### PATCH END: allocate only the active role's execution stage.
+
+    def compute_ffn_output(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        """Execute this layer's complete native FFN block on the FFN role."""
+        if self.afd_role != "ffn":
+            raise RuntimeError("Qwen3 MoE FFN compute requires the AFD FFN role")
+        return self.mlp(hidden_states)
+
+
+@native.support_torch_compile
```
hsliuustc0106: [P1] Avoid double compile initialization

Native `Qwen3MoeModel` is already wrapped by `support_torch_compile`, and this subclass calls that wrapped `super().__init__()`. Decorating the subclass again initializes `TorchCompileWithNoGuardsWrapper` twice in non-eager mode, creating duplicate compiler backends and bytecode hooks while overwriting the first cleanup handle. Please inherit the native compile wrapper without applying the decorator again, and add a regression test that compile support initializes once.

afd_plugin/model_executor/models/qwen3_moe.py
```diff
@@ -0,0 +1,325 @@
...
+            yield name, loaded_weight
+
+
+def _validate_supported_config(
+    vllm_config: VllmConfig,
+    afd_config: AFDConfig,
+) -> None:
+    """Reject Qwen3 MoE modes not covered by the initial CUDA contract."""
+    parallel_config = vllm_config.parallel_config
+    if current_platform.device_type != "cuda":
```
hsliuustc0106: [P2] Distinguish CUDA from ROCm

vLLM's `RocmPlatform` also reports `device_type = "cuda"`, so this condition accepts the explicitly unsupported ROCm backend. The exact vLLM 0.26 validator accepted a `RocmPlatform` instance in a direct probe. Please use `if not current_platform.is_cuda():` and add a ROCm negative test.

tests/e2e/models/qwen3_moe/test_e2e_gpu.py
```diff
@@ -0,0 +1,261 @@
...
+        pytest.fail("GPU E2E device list must contain at least one device")
+    return gpus
+
+
+def _assert_runtime_signals(case: Qwen3MoeE2ECase, output: str) -> None:
+    _, ready, runtime_output = output.partition(API_READY_SIGNAL)
+    assert ready, "runner completed without reporting Attention API readiness"
+
+    if case.graph_capture_size:
+        assert GRAPH_RUNTIME_SIGNAL in runtime_output, (
```
hsliuustc0106: [P2] Verify FFN graph replay

This signal is emitted by the native Attention runner and does not prove that `GPUFFNModelRunner` selected `AFDGraphRunMode.REPLAY`. If the FFN graph key is absent, the FFN runner falls back to eager execution while this assertion still passes. Please emit an FFN-side replay signal and assert it after API readiness.

hsliuustc0106 (commented):
Three inline findings on the current head: one compiler-lifecycle issue and two validation gaps.

jiangkuaixue123:
Thanks for this PR.

Since the scope explicitly notes that this does not make any performance claims, I suggest we discuss the performance-related aspects of this change at our regular meeting. Please see the agenda for reference and add a topic/agenda item in the doc if you would like it discussed:

https://docs.google.com/document/d/1FMeSNdIyQEgeZUqj4ddalP__rZdzJSgzob1csEhPP3I

afd_plugin/model_executor/models/qwen3_moe.py
sinpeyw: Fixed. Removed the duplicate decorator. The regression verifies that compile support comes from the native wrapper and is not applied again by the AFD subclass.

afd_plugin/model_executor/models/qwen3_moe.py
sinpeyw: Fixed. The validator now uses current_platform.is_cuda(), with a negative test covering ROCm semantics.

tests/e2e/models/qwen3_moe/test_e2e_gpu.py
sinpeyw: Fixed. The FFN runner now emits a DEBUG signal only from AFDGraphRunMode.REPLAY, and the graph E2E asserts it after API readiness.

sinpeyw:
Rebased onto the latest main. E2E changes were removed and will follow in a separate PR.

jiangkuaixue123:
LGTM. Thanks for the contribution!

PR #217, opened 2026-08-08
[Bugfix] Avoid cross-ubatch Attention metadata reuse in AFD DBO

sinpeyw:
## Purpose

Fix incorrect Attention metadata reuse in AFD GPU DBO with vLLM 0.26.

## Root Cause

When DBO builds metadata for multiple ubatches, vLLM 0.26 caches it by KV-cache specification and builder type, but not by ubatch ID.

As a result, ubatch 1 can reuse sequence metadata from ubatch 0 while only updating its block table, producing inconsistent Attention metadata.

The upstream fix is [vllm-project/vllm#48659](https://github.com/vllm-project/vllm/pull/48659).

## Fix

During AFD multi-ubatch metadata construction, temporarily disable the block-table-only update path and restore every affected builder in `finally`.

This makes each ubatch build consistent metadata. Single-ubatch execution is unchanged.

## Why not backport the upstream cache-key change?

The upstream fix is preferable because it adds the ubatch ID to the cache key while preserving the optimization.

However, that cache key is local to the roughly 329-line private `_build_attention_metadata` method in the pinned vLLM release. An exact plugin backport would require copying that complete method, creating substantial version drift and maintenance cost. A global monkey patch would also affect non-AFD execution.

The scoped workaround avoids both problems. Its trade-off is rebuilding Attention metadata during AFD multi-ubatch execution; it does not disable DBO, CUDA Graph, KV cache, or model compute optimizations. It should be removed when vLLM #48659, or an equivalent fix, reaches the pinned release.

This compatibility fix was separated from #216 because it is not specific to the Qwen3 adapter.

## Validation

- Focused Attention runner tests: 45 passed.
- Complete unit suite: 409 passed, 39 skipped.
- Ruff check and format: passed.
- `git diff --check`: passed.

## Docs Impact

Updated the compatibility documentation with the workaround scope and removal condition.

### DIFF
--- a/afd_plugin/v1/worker/attention_model_runner.py
+++ b/afd_plugin/v1/worker/attention_model_runner.py
@@ -15,7 +15,10 @@ from vllm.config import CUDAGraphMode, VllmConfig
 from vllm.distributed.parallel_state import get_world_group
 from vllm.forward_context import BatchDescriptor, DPMetadata, get_forward_context
 from vllm.sequence import IntermediateTensors
-from vllm.v1.attention.backend import CommonAttentionMetadata
+from vllm.v1.attention.backend import (
+    AttentionMetadataBuilder,
+    CommonAttentionMetadata,
+)
 from vllm.v1.outputs import AsyncModelRunnerOutput, ModelRunnerOutput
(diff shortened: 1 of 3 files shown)


jiangkuaixue123:
LGTM

PR #219, opened 2026-08-08
[Bugfix] Fix CUDA FFN token metadata for asymmetric topologies

### LINKED ISSUE #218: [Bug]: CUDA FFN uses Attention-rank token counts in asymmetric topologies
# [Bug]: CUDA FFN uses Attention-rank token counts in asymmetric topologies

## Current environment

```text
Python: 3.12
OS: Linux
afd-plugin commit: 2e4378f2d478313992e9a1b850c61b1e888a1145
vLLM version/checkout: 0.26.0+cu129
CUDA/GPU: CUDA 12.9 / NVIDIA A800-SXM4-80GB
Install command: uv sync --group dev --extra vllm
```

## Reproduction

Run the existing E2E runner with a supported `4A2F` P2P topology:

```sh
python tests/e2e/runner.py \
  --model /path/to/DeepSeek-V2-Lite \
  --num-attention-ranks 4 \
  --num-ffn-ranks 2 \
  --attention-gpus 0,1,2,3 \
  --ffn-gpus 4,5 \
  --max-tokens 8 \
  --num-requests 8 \
  --request-concurrency 8 \
  --common-vllm-arg=--trust-remote-code
```

## Expected behavior

`4A2F` satisfies the P2P topology contract (`A >= F` and `A % F == 0`). Each
FFN rank should run with the combined token count of its two Attention peers.

## Actual behavior and logs

The FFN runner passes Attention-rank token counts directly into vLLM's FFN
forward context. An FFN rank therefore receives the combined hidden states from
two peers while all-to-all sees only one peer's token count and aborts:

```text
File "vllm/distributed/device_communicators/all2all.py", line 116, in dispatch
  assert sizes[dist_group.rank_in_group] == hidden_states.shape[0]
AssertionError
```

## AFD configuration

```yaml
afd:
  role: attention  # ffn on the FFN server
  connector: P2pNcclAFDConnector
  host: 127.0.0.1
  port: 1239
  num_attention_ranks: 4
  num_ffn_ranks: 2
```

## Additional context

The connector must retain the original per-Attention-rank counts for tensor
transfers; only the FFN execution metadata should be aggregated. NPU already
performs analogous Attention-to-FFN aggregation for FFN execution, introduced
for its runner and graph key in PR #18. CUDA needs the corresponding execution
metadata while matching its P2P peer shapes. This is distinct from PR #69,
which fixed peer-specific GPU transfer metadata for `2A1F`; the failure here is
FFN execution metadata when `F > 1` and `A:F > 1`.

## Before submitting

- [x] I confirmed this against vLLM v0.26.0.

sinpeyw:
# [Bugfix] Fix CUDA FFN token metadata for asymmetric topologies

## Purpose

Fix CUDA AFD execution for supported asymmetric topologies such as `4A2F`.
The FFN runner previously passed per-Attention-rank token counts into vLLM's
FFN forward context, causing all-to-all size assertions when each FFN rank
received tensors from multiple Attention peers.

NPU already performs analogous Attention-to-FFN aggregation for FFN
execution. This PR adds the CUDA counterpart while matching CUDA P2P peer
shapes.

## Issue

- Closes #218

## Scope

- In scope: CUDA FFN forward metadata for supported P2P topology ratios.
- Out of scope: connector transfer metadata, topology rules, changes to the
  existing NPU behavior, and model-specific changes.

## Implementation Notes

- Add a CPU-safe helper for Attention-to-FFN token-count aggregation and FFN
  role-rank to DP-rank projection.
- Build FFN DP metadata once per stage while preserving the original Attention
  metadata for connector control and tensor transfers.
- Preserve the original per-Attention-rank counts in CUDA graph keys because
  peer receive shapes are part of the captured graph.

No vLLM source changes, compatibility shims, or monkey patches are added.

## Test Plan

- Unit coverage for asymmetric per-stage aggregation, peer-preserving graph
  keys, and TP-to-DP projection.
- CUDA E2E coverage for `4A2F` eager, `FULL_DECODE_ONLY`, and
  `FULL_DECODE_ONLY` with two-way DBO.

## Test Result

- `uv run ruff check .`: passed.
- `uv run ruff format --check .`: passed (138 files unchanged).
- `git diff --check`: passed.
- `pytest tests/unit`: 413 passed, 39 skipped.
- Focused CUDA graph/FFN runner unit tests: 43 passed.
- Unmodified `main`, DeepSeek-V2-Lite `4A2F` eager: reproduced the vLLM
  all-to-all size assertion.
- Patched DeepSeek-V2-Lite `4A2F`: eager, CUDA graph capture/replay, and CUDA
  graph with two-way DBO each completed 8 concurrent requests successfully.

## Docs Impact

- Files updated: none.
- Reason: this restores behavior already allowed by the documented P2P topology
  contract and does not add or change a user-facing interface.

---

<details>

- [x] Scope is bounded.
- [x] Any compat shim or monkey patch is isolated, idempotent, version-guarded, documented, and tested. (N/A)

</details>

### DIFF
--- a/afd_plugin/v1/worker/ffn_metadata.py
+++ b/afd_plugin/v1/worker/ffn_metadata.py
@@ -37,9 +37,9 @@ def aggregate_ffn_token_counts(
 
     group_size = attention_size // ffn_size
     return tuple(
-        max(
-            1,
-            sum(expanded_counts[rank * group_size : (rank + 1) * group_size]),
+        sum(
+            max(1, int(count))
+            for count in expanded_counts[rank * group_size : (rank + 1) * group_size]
         )
         for rank in range(ffn_size)
     )
--- a/afd_plugin/v1/worker/ffn_metadata.py
+++ b/afd_plugin/v1/worker/ffn_metadata.py
@@ -1,0 +1,62 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
+"""CPU-safe token metadata helpers for AFD FFN runtimes."""
+
+from __future__ import annotations
+
+
+def aggregate_ffn_token_counts(
+    attention_counts: tuple[int, ...],
+    *,
+    attention_size: int,
+    ffn_size: int,
+    fallback: int = 1,
+) -> tuple[int, ...]:
+    """Aggregate consecutive Attention-rank counts for each FFN rank."""
+
+    fallback_counts = tuple(max(1, int(fallback)) for _ in range(ffn_size))
+    if not attention_counts:
+        return fallback_counts
+
+    expanded_counts = attention_counts
+    if (
+        len(attention_counts) < attention_size
+        and attention_size % len(attention_counts) == 0
+    ):
+        parallel_size = attention_size // len(attention_counts)
+        expanded_counts = tuple(
+            attention_counts[rank // parallel_size] for rank in range(attention_size)
+        )
+
+    if (
+        len(expanded_counts) < attention_size
+        or attention_size < ffn_size
+        or attention_size % ffn_size != 0
+    ):
+        return fallback_counts
+
+    group_size = attention_size // ffn_size
+    return tuple(
+        max(
+            1,
+            sum(expanded_counts[rank * group_size : (rank + 1) * group_size]),
+        )
+        for rank in range(ffn_size)
+    )
+
+
+def project_ffn_token_counts_to_dp(
+    ffn_counts: tuple[int, ...],
+    *,
+    dp_size: int,
+) -> tuple[int, ...]:
+    """Project FFN role-rank counts to vLLM data-parallel rank counts."""
+
+    if len(ffn_counts) == dp_size or dp_size <= 0 or len(ffn_counts) % dp_size != 0:
+        return ffn_counts
+
+    parallel_size = len(ffn_counts) // dp_size
+    return tuple(ffn_counts[rank * parallel_size] for rank in range(dp_size))
+
+
+__all__ = ["aggregate_ffn_token_counts", "project_ffn_token_counts_to_dp"]
--- a/afd_plugin/v1/worker/ffn_metadata.py
+++ b/afd_plugin/v1/worker/ffn_metadata.py
@@ -12,36 +12,42 @@ def aggregate_ffn_token_counts(
     ffn_size: int,
     fallback: int = 1,
 ) -> tuple[int, ...]:
-    """Aggregate consecutive Attention-rank counts for each FFN rank."""
+    """Aggregate consecutive Attention-rank counts for each FFN rank.
 
-    fallback_counts = tuple(max(1, int(fallback)) for _ in range(ffn_size))
-    if not attention_counts:
+    For example, ``4A2F`` counts ``(0, 4, 5, 6)`` become ``(5, 11)`` because
+    every zero-token Attention peer contributes one placeholder row. Missing
+    peers use the same per-peer fallback, so empty counts become ``(2, 2)``.
+    """
+
+    fallback_count = max(1, int(fallback))
+    fallback_counts = tuple(fallback_count for _ in range(max(0, ffn_size)))
+    if ffn_size <= 0 or attention_size < ffn_size or attention_size % ffn_size != 0:
         return fallback_counts
 
     expanded_counts = attention_counts
     if (
-        len(attention_counts) < attention_size
+        attention_counts
+        and len(attention_counts) < attention_size
         and attention_size % len(attention_counts) == 0
     ):
-        parallel_size = attention_size // len(attention_counts)
+        attention_ranks_per_count = attention_size // len(attention_counts)
         expanded_counts = tuple(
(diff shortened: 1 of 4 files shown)


afd_plugin/v1/worker/ffn_metadata.py
```diff
@@ -0,0 +1,62 @@
...
+        or attention_size < ffn_size
+        or attention_size % ffn_size != 0
+    ):
+        return fallback_counts
+
+    group_size = attention_size // ffn_size
+    return tuple(
+        max(
+            1,
+            sum(expanded_counts[rank * group_size : (rank + 1) * group_size]),
```
hsliuustc0106: [P1] Clamp every Attention peer before summing

The connector clamps each idle Attention peer to a one-row dummy tensor, but this code clamps only the final group sum. In `4A2F`, counts `(0, 4, 5, 6)` therefore produce FFN metadata `(4, 11)` while the connector receives `(5, 11)`, recreating the all-to-all size assertion this PR fixes. Apply `max(1, count)` before summing each group and add an idle-rank regression test.

afd_plugin/v1/worker/ffn_model_runner.py
```diff
@@ -236,6 +244,34 @@ def _execute_eager_mode(
     ) -> torch.Tensor:
         return self.model.compute_ffn_output(hidden_states, layer_idx)
 
+    def _make_graph_key(
+        self,
+        dp_metadata_list: dict[int, DPMetadata | AFDDPMetadata],
+    ) -> tuple:
+        return make_ffn_graph_key(
```
hsliuustc0106: [P1] Preserve peer shapes in the CUDA graph key

The captured FFN graph includes fixed-shape receives and return slices for every Attention peer, so aggregate totals are not a sufficient cache key. For example, `(1, 3, 5, 7)` and `(2, 2, 6, 6)` both become `((0, (4, 12)),)` but require different peer buffers. An eager or otherwise nonuniform step can consequently replay an incompatible captured graph. Keep the original Attention vector, or include it alongside the aggregate; only forward-context DP metadata should be aggregated.

hsliuustc0106 (commented):
Two blocking correctness findings on the current head, covering idle Attention ranks and CUDA graph cache-key safety.

afd_plugin/v1/worker/ffn_metadata.py
sinpeyw: Fixed. Each Attention peer is now clamped before grouping, so (0, 4, 5, 6) produces (5, 11). A regression test covers this case.

afd_plugin/v1/worker/ffn_model_runner.py
sinpeyw: Fixed. CUDA graph keys now preserve the original Attention peer vector; aggregation is limited to FFN execution metadata. A collision regression test was added.

afd_plugin/v1/worker/ffn_metadata.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def aggregate_ffn_token_counts(
+    attention_counts: tuple[int, ...],
+    *,
+    attention_size: int,
+    ffn_size: int,
+    fallback: int = 1,
+) -> tuple[int, ...]:
+    """Aggregate consecutive Attention-rank counts for each FFN rank."""
```
jiangkuaixue123: Maybe add some example input/output in the comments

afd_plugin/v1/worker/ffn_model_runner.py
```diff
@@ -236,6 +244,30 @@ def _execute_eager_mode(
     ) -> torch.Tensor:
         return self.model.compute_ffn_output(hidden_states, layer_idx)
 
+    def _make_graph_key(
```
jiangkuaixue123: Maybe this helper function is not needed.

afd_plugin/v1/worker/ffn_model_runner.py
```diff
@@ -405,6 +437,14 @@ def _set_moe_layer_index(forward_context: object, layer_idx: int) -> None:
             return
 
 
+def _metadata_values_tuple(
```
jiangkuaixue123: Is this the same as cuda_graph.py:155

afd_plugin/v1/worker/ffn_model_runner.py
sinpeyw: Removed the thin graph-key helper and call make_ffn_graph_key directly at both call sites.

afd_plugin/v1/worker/ffn_model_runner.py
sinpeyw: Yes. Removed the duplicate conversion helper and kept the direct tuple conversion at the FFN metadata call site; cuda_graph.py is now unchanged from upstream.

afd_plugin/v1/worker/ffn_metadata.py
sinpeyw: Added a concrete 4A2F example: (0, 4, 5, 6) becomes (5, 11), including the per-idle-peer placeholder behavior.

jiangkuaixue123:
LGTM!

PR #220, opened 2026-08-10
Update WeChat community QR-code image

david6666666:
## Summary

- Replace `docs/assets/WeChat.jpg` with the updated AFD plugin developer community QR-code image.
- Keep the repository asset as a valid `1203x1683` RGB JPEG.

## Validation

- `git diff --check`
- Verified the generated image decodes as JPEG with Pillow.


jiangkuaixue123:
@david6666666 Is this need merge?

PR #222, opened 2026-08-10
refactor e2e tests

yujuancao07:
## Purpose

Simplify and harden the DeepSeek-V2-Lite E2E tests before integrating them into the GPU/NPU CI gate.

## Scope

- In scope:
    - Four fixed E2E scenarios: baseline-graph, afd-eager, afd-graph, and afd-graph-dbo
    - Unified GPU/NPU configuration
    - GSM8K-7 PR coverage and weekly full GSM8K support
    - Strict failure and process cleanup handling
    - Runner unit tests and E2E documentation

## Implementation Notes

1. Use two Attention devices and one FFN device for AFD scenarios.
2. Remove redundant TP, profiler, serving, and async CAM E2E cases.
3. Move the NPU operator test to tests/unit/compat/npu/.

## Test Plan
```
python -m pytest -q tests/unit/test_e2e_runner.py
python -m pytest --collect-only -q tests/e2e/test_deepseek_v2_lite.py
python -m pytest -q -s tests/e2e/test_deepseek_v2_lite.py
```
Weekly full accuracy test:
```
AFD_GSM8K_LIMIT=all python -m pytest -q -s \
  "tests/e2e/test_deepseek_v2_lite.py::test_deepseek_v2_lite[afd-graph-dbo]"
```


tests/e2e/runner.py
```diff
@@ -464,6 +589,43 @@ def ffn_api_port(args: argparse.Namespace) -> int:
...
+    )
+    sample_count = _extract_gsm8k_sample_count(results)
+    assert sample_count == expected_sample_count, (
+        f"GSM8K evaluated {sample_count} samples; expected {expected_sample_count}"
+    )
+    minimum_accuracy = float(
+        os.environ.get(GSM8K_THRESHOLD_ENV, str(DEFAULT_GSM8K_THRESHOLD)),
+    ) - float(os.environ.get(GSM8K_TOLERANCE_ENV, str(DEFAULT_GSM8K_TOLERANCE)))
+    accuracy = _extract_gsm8k_accuracy(results)
+    assert accuracy >= minimum_accuracy, (
```
jiangkuaixue123: These checks run inside the subprocess script (`python -m tests.e2e.runner`), not under pytest, so they are not assert-rewritten and would be silently disabled under `python -O`. The rest of this file consistently raises explicit exceptions (`RuntimeError`/`ValueError`). Consider raising `RuntimeError` with the same messages here for consistency and robustness.

tests/e2e/runner.py
```diff
@@ -282,23 +320,63 @@ def parse_args() -> argparse.Namespace:
     return parser.parse_args()
 
 
+def configure_scenario(args: argparse.Namespace) -> None:
```
jiangkuaixue123, line 256: With the fixed-scenario design, several legacy flexible flags are now dead or silently overridden: `configure_scenario` unconditionally overwrites `--num-attention-ranks`, `--num-ffn-ranks`, `--tp-size*`, and the graph/DBO flags, while `--afd-async`, `--compute-gate-on-attention`, `--afd-connector-extra-config`, `--use-decode-bench-connector`, and the completion-request options (`--prompt`, `--max-tokens`, `--temperature`, `--num-requests`, `--request-concurrency`, `--expect-text`) are not exercised by any of the four scenarios. The completion smoke path (`request_completions`) is also no longer reachable from `main()`. The CLI surface advertises options that have no effect — consider removing the unused flags/helpers, or explicitly wiring them into the scenarios.

tests/e2e/models/deepseek_v2_lite/test_deepseek_v2_lite.py
```diff
@@ -0,0 +1,137 @@
...
+    ]
+
+
+def _run_runner(command: list[str]) -> None:
+    handled_signals = (signal.SIGTERM, signal.SIGINT)
+    previous_handlers = {signum: signal.getsignal(signum) for signum in handled_signals}
+    process: subprocess.Popen | None = None
+    received_signal: int | None = None
+    forwarded = False
+
```
jiangkuaixue123, line 92: For `baseline-graph`, the command still passes `--ffn-devices <third-device>`, but `validate_topology` ignores FFN devices entirely in baseline mode (`num_ffn_ranks` is forced to 0), so a device that appears to be used is silently unused. Minor, but passing only the first device (or omitting the flag) would make the baseline command reflect what actually runs.

tests/e2e/helpers_gsm8k.py
```diff
@@ -8,11 +8,80 @@
...
+    except ProcessLookupError:
+        group_alive = False
+    except OSError as exc:
+        failures.append(f"SIGTERM failed for lm-eval group {process.pid}: {exc}")
+
+    deadline = time.monotonic() + LM_EVAL_TERMINATION_TIMEOUT_S
+    while group_alive:
+        try:
+            os.killpg(process.pid, 0)
+        except ProcessLookupError:
```
jiangkuaixue123: This SIGTERM → liveness-poll → SIGKILL → reap sequence duplicates `runner.terminate_processes` (`tests/e2e/runner.py:806-870`); the two differ only in timeout constants and the extra reader join. Consider extracting a shared process-group termination helper so the two hand-rolled copies cannot drift apart.

jiangkuaixue123:
Thanks for consolidating and hardening the E2E framework in this PR. Since this design will become the foundation for future E2E coverage, could you please add a Markdown design document under `docs/design/module/` that defines the rules for adding new E2E cases?

The document should make the conventions introduced by this PR explicit, including at least:

- the acceptance criteria for adding a new scenario, model, backend, topology, or execution mode;
- how new cases must reuse the shared runner instead of introducing standalone launch and cleanup logic;
- the required pytest structure, scenario naming, environment variables, device-role ordering, and GPU/NPU separation;
- the rule that hardware-gate prerequisites fail clearly rather than silently skipping;
- sequential execution and device isolation requirements;
- process lifecycle requirements, including startup checks, signal handling, cleanup, log capture, and detection of leaked processes;
- the distinction between the bounded PR-gate accuracy run and the scheduled full-accuracy run;
- the unit tests required for any runner or scenario extension;
- the documentation and `run-e2e` skill updates required when the E2E contract changes;
- how unsupported combinations and backend-specific exclusions must be declared.

Please treat that Markdown file as the source of truth for subsequent E2E additions, so future contributors and reviewers can require every new E2E case to follow the same framework and rules established here.

.agents/skills/run-e2e/SKILL.md
```diff
@@ -1,173 +1,103 @@
...
-note "limit ignored — that category has no gsm8k".
+Honor an explicit backend. Otherwise inspect nvidia-smi -L and npu-smi info.
+If both are available, ask which to use. If neither is available, stop.
 
-## Workflow
+### 2. Validate prerequisites
+
+Before starting pytest, confirm:
+
+- AFD_E2E_DEVICES contains exactly 3 unique device IDs.
```
jiangkuaixue123: We may have more than 3 device test cases later.

yujuancao07: Current E2E tests use at most three devices. If a future E2E case requires more devices, you can update AFD_E2E_DEVICES handling and the topology accordingly.

PR #236, opened 2026-08-13
NPU Async Connector: tests and docs

ShwStone:
## Summary

- update NPU unit-test fixtures and module isolation for the vLLM v0.26 runtime
- add a fixed DeepSeek-V2-Lite async CAM ubatching E2E scenario:
  - Attention DP=1/TP=2
  - FFN DP=1/TP=1
  - two-stage token split
  - lm-eval batch size 2
  - GSM8K sample-count and accuracy gates
- consolidate NPU testing and troubleshooting documentation

The fixed scenario overrides conflicting connector settings so it always exercises the intended token-split configuration. Import and ABI errors from an installed NPU runtime are no longer silently ignored.

## Validation

- `uv run --frozen pytest tests/unit/test_e2e_runner.py tests/unit/v1/worker/test_npu_mla_graph.py`
  - 83 passed, 1 skipped because `torch` was unavailable
- `uv run --frozen ruff check ...`
- `uv run --frozen ruff format --check ...`
- `git diff --check`

## Known issue

NPU async workloads can complete evaluation, but process cleanup currently causes pytest to fail during teardown: #251.


ShwStone:
image

PR #243, opened 2026-08-13
[Feature] Add deterministic forced routing for CUDA MoE benchmarks

### LINKED ISSUE #242: [RFC]: Add deterministic forced routing for CUDA MoE benchmarks
## Motivation

Controlled AFD performance studies need native and disaggregated MoE runs to use deterministic routing with balanced token volume across EP ranks. The NPU benchmark path introduced in #40 provides this control and an optional local expert-pool limit, but CUDA has no equivalent.

Natural routing makes compute and communication load depend on model logits. A model-specific workaround would also be difficult to reuse and review.

## Proposed change

- Add an opt-in vLLM 0.26.0 CUDA compatibility patch.
- Enable it only when `AFD_BENCHMARK_FORCE_LOAD_BALANCE=1` is set before vLLM starts.
- Let `AFD_BENCHMARK_FORCE_LB_TOPN_PER_RANK` select the active local experts on each EP rank; `0` uses all local experts.
- Preserve router weights and replace only expert IDs with a cached, deterministic balanced cycle.
- Use distinct source-rank phases, matching the aggregation behavior repaired by #156 and #157.
- Require an evenly partitioned expert pool across EP ranks.
- Reject EPLB while the benchmark control is enabled.

The control applies to both native vLLM and AFD workers. It intentionally changes model outputs and is not intended for correctness or production use.

The CUDA activation uses an environment switch rather than NPU's `additional_config` key because the general plugin hook runs before a concrete `VllmConfig` exists. Using config-only activation would require installing a global CUDA router patch in every plugin-enabled process. The environment gate keeps the unconfigured CUDA path identical to upstream while preserving the same routing semantics once enabled.

## Plugin boundary

- **Plugin-owned:** activation, deterministic cycle construction, validation, per-router cache state, tests, and documentation.
- **Compatibility patch:** exact-version wrappers for `FusedMoERouter.__init__` and `BaseRouter._select_experts`, with expert IDs replaced before vLLM captures routing metadata.
- **Upstream-owned:** router weights, expert kernels, communication collectives, model implementations, and normal routing when the switch is absent.

No vLLM source file is modified. The patch should be removed when vLLM exposes a post-routing profiling hook that can replace expert IDs while preserving native routing computation and weights.

## Risks and alternatives

- vLLM's `RoutingSimulator` replaces routing computation and weights; this proposal preserves native routing and replaces only expert IDs to match the NPU benchmark semantics.
- The benchmark switch changes model outputs and patches private router internals, so it is opt-in, exact-version guarded, and covered by focused tests.

## Validation plan

- Runtime unit coverage for deterministic cycles, expert bounds, source-rank phases, cache reuse, native routing capture, and patch installation.
- CUDA smoke coverage for native EP and an asymmetric AFD topology.
- Re-run the controlled Qwen3 16K/32K native and 6A2F benchmark matrix before publishing the dependent recipe.

## Feedback period

Open for maintainer feedback before the implementation PR is submitted.

sinpeyw:
## Purpose

Add an opt-in CUDA benchmark control that gives native and AFD MoE runs deterministic routing with balanced token volume across EP ranks.

This extends the existing NPU benchmark semantics to CUDA without making the control model- or connector-specific.

## Issue

- Related issue(s): #242
- Closing keyword, only if fully resolved: Closes #242

## Scope

- In scope:
  - vLLM 0.26.0 CUDA routing;
  - deterministic source-rank phases across EP ranks;
  - an optional per-rank active-expert limit;
  - native and AFD benchmark processes.
- Out of scope:
  - correctness or production serving;
  - natural-routing performance claims;
  - uneven expert placement, EPLB, NPU changes, model adapters, connectors, and kernels.

## Implementation Notes

- `VLLM_MOE_ROUTING_SIMULATION_STRATEGY=afd_balanced` selects the plugin-owned strategy through vLLM's routing simulator.
- `AFD_BENCHMARK_FORCE_LB_TOPN_PER_RANK=0` uses all local experts; a positive value limits the deterministic pool on every EP rank.
- The strategy returns normalized weights and deterministic expert IDs; routing buffers are cached by topology, device, and dtype.
- vLLM continues to own strategy selection, routing capture, EPLB mapping, and index conversion. EPLB remains outside this benchmark's validated scope.
- The CUDA path uses vLLM's strategy selector rather than NPU's method-owned `additional_config` path because vLLM already exposes the required extension point.

No vLLM source file or router implementation is modified. Without the strategy selector, normal vLLM routing is unchanged.

## Test Plan

```bash
uv run ruff check .
uv run ruff format --check .
uv run pytest -q tests/unit/model_executor/test_routing_simulator.py
git diff --check
```

GPU validation:

- native CUDA EP smoke with the complete local expert pool;
- asymmetric AFD smoke with a restricted expert pool;
- controlled Qwen3 native and 6A2F throughput matrix.

## Test Result

- Focused routing-simulator tests: 4 passed.
- Complete unit suite with vLLM 0.26.0: 551 passed, 46 skipped.
- Native DP2/EP2 and asymmetric AFD 2A1F graph smokes: passed on A800.
- Ruff check, Ruff format, and `git diff --check`: passed.

## Docs Impact

- Files updated: `docs/design/module/compatibility_and_patches.md`.
- Documents activation, compatibility scope, validation, and removal criteria.

---

<details>

- [x] Purpose is clear and linked to public context.
- [x] Scope is bounded.
- [x] The implementation uses vLLM's routing-simulator registration interface and does not replace router internals.
- [x] Imports remain CPU-safe; vLLM routing is imported only during plugin
  registration when vLLM is installed.
- [x] Final CPU and CUDA validation evidence is included.

</details>

### DIFF
--- a/afd_plugin/__init__.py
+++ b/afd_plugin/__init__.py
@@ -114,6 +114,12 @@ def register_afd() -> None:
             exc_info=True,
         )
 
+    from afd_plugin.model_executor.routing_simulator import (
+        register_afd_balanced_routing_strategy,
+    )
+
+    register_afd_balanced_routing_strategy()
+
     try:
         from afd_plugin.v1.worker.dbo import register_dbo_yield_custom_op
 
--- a/afd_plugin/model_executor/routing_simulator.py
+++ b/afd_plugin/model_executor/routing_simulator.py
@@ -1,0 +1,178 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
+"""Deterministic routing strategy for controlled CUDA MoE benchmarks.
+
+Select ``afd_balanced`` with vLLM's
+``VLLM_MOE_ROUTING_SIMULATION_STRATEGY`` environment variable. The optional
+``AFD_BENCHMARK_FORCE_LB_TOPN_PER_RANK`` variable limits the local expert pool
+on every EP rank; ``0`` selects all local experts.
+
+The strategy returns normalized uniform weights and deterministic expert IDs.
+It changes model outputs and is intended only for benchmark and profiling runs.
+"""
+
+from __future__ import annotations
+
+import os
+from dataclasses import dataclass
+
+import torch
+from vllm.distributed.parallel_state import get_ep_group
+from vllm.model_executor.layers.fused_moe.router.routing_simulator_router import (
+    RoutingSimulator,
+    RoutingStrategy,
+)
+
+AFD_BALANCED_ROUTING_STRATEGY = "afd_balanced"
+_TOPN_PER_RANK_ENV = "AFD_BENCHMARK_FORCE_LB_TOPN_PER_RANK"
+_DETERMINISTIC_SEED = 1024
+
+
+@dataclass(frozen=True)
+class _RoutingConfig:
+    num_experts: int
+    ep_size: int
+    ep_rank: int
+    top_k: int
+    topn_per_rank: int
+
+
+@dataclass(frozen=True)
+class _RoutingBuffers:
+    weights: torch.Tensor
+    expert_ids: torch.Tensor
+
+
+def _validate_config(config: _RoutingConfig) -> None:
+    if config.num_experts <= 0:
+        raise ValueError("num_experts must be positive")
+    if config.ep_size <= 0:
+        raise ValueError("ep_size must be positive")
+    if not 0 <= config.ep_rank < config.ep_size:
+        raise ValueError("ep_rank must be within the expert-parallel group")
+    if config.top_k <= 0:
+        raise ValueError("top_k must be positive")
+    if config.num_experts % config.ep_size != 0:
+        raise ValueError("num_experts must be divisible by ep_size")
+
+    local_experts = config.num_experts // config.ep_size
+    selected_per_rank = config.topn_per_rank or local_experts
+    if not 0 < selected_per_rank <= local_experts:
+        raise ValueError(f"{_TOPN_PER_RANK_ENV} must be between 0 and {local_experts}")
+    if config.top_k > selected_per_rank * config.ep_size:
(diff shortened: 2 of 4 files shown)


jiangkuaixue123:
Could we avoid this version-specific monkey patch and use vLLM's built-in routing simulator instead? For example, AFD could register its own strategy via `RoutingSimulator.register_strategy(...)` and select it with `VLLM_MOE_ROUTING_SIMULATION_STRATEGY=<afd_strategy>`.

I understand that the simulator strategy returns both routing weights and expert IDs, whereas this patch preserves the native weights and only overrides the IDs. Is preserving the exact native weights essential for the benchmark results? If the AFD strategy can reproduce the required weight semantics, this approach would encapsulate the deterministic/load-balanced routing in plugin-owned code while avoiding patches to `FusedMoERouter` and `BaseRouter` internals, which should be easier to maintain across vLLM versions.

sinpeyw:
> Could we avoid this version-specific monkey patch and use vLLM's built-in routing simulator instead? For example, AFD could register its own strategy via `RoutingSimulator.register_strategy(...)` and select it with `VLLM_MOE_ROUTING_SIMULATION_STRATEGY=<afd_strategy>`.
> 
> I understand that the simulator strategy returns both routing weights and expert IDs, whereas this patch preserves the native weights and only overrides the IDs. Is preserving the exact native weights essential for the benchmark results? If the AFD strategy can reproduce the required weight semantics, this approach would encapsulate the deterministic/load-balanced routing in plugin-owned code while avoiding patches to `FusedMoERouter` and `BaseRouter` internals, which should be easier to maintain across vLLM versions.

I have replaced the version-specific router monkey patch with a plugin-owned afd_balanced strategy registered through vLLM's RoutingSimulator. Exact native weights are not required for this benchmark, so the strategy returns normalized uniform weights while preserving deterministic expert placement and balanced token volume.

PR #244, opened 2026-08-13
feat: Parallelize Attention and FFN model loading across AFD connectors

### LINKED ISSUE #237: [Feature]: Parallelize Attention and FFN model loading across AFD connectors
## Motivation

Attention and FFN model loading is currently serialized across all AFD connectors.

The common startup sequence is:

```text
Attention                               FFN
init_device                             init_device
  construct model runner                 construct model runner
  initialize connector (blocks)          load model
       |                                 finish EngineCore initialization
       +---- FFN initializes connector --+
cross-role rendezvous completes
load Attention model
```

This leaves the FFN ranks idle for most of the Attention weight-loading time and increases total service startup latency. The impact is especially visible with large NPU checkpoints, but the lifecycle issue is not specific to Async CAM.

## Affected connectors

All currently registered connectors use a collective cross-role initialization path:

| Connector | Platform | Cross-role initialization |
| --- | --- | --- |
| `P2pNcclAFDConnector` | CUDA | Joins the Attention/FFN NCCL world before creating data and metadata subgroups |
| `CAMP2pAFDConnector` | Ascend NPU | Joins Attention/FFN HCCL groups before creating the FFN-only and Gloo groups |
| `CAMAsyncAFDConnector` | Ascend NPU | Joins the combined Attention/FFN Async CAM HCCL world |

## Root cause

vLLM workers execute `init_device()` before `load_model()`.

The AFD roles initialize their connectors at different lifecycle stages:

- `AFDAttentionModelRunner.__init__()` initializes the GPU connector immediately.
- `AFDNPUAttentionModelRunner.__init__()` initializes the NPU connector immediately, for both synchronous and asynchronous CAM.
- The GPU and NPU FFN model runners only construct their connectors.
- FFN initializes its connector later from `start_ffn_server_loop()`, after its model has loaded.
- Each connector's `init_afd_connector()` requires participating Attention and FFN ranks to rendezvous.

As a result, Attention blocks inside `init_device()` and cannot enter `load_model()` until FFN completes model loading and reaches connector-loop startup.

There is also a misleading startup log in `_run_ffn_busy_loop()`: `AFD FFN EngineCore started; workers run connector loop.` is emitted before the `start_ffn_server_loop` collective RPC returns, so it does not prove that the connector loop is ready.

## Proposed direction

1. Keep connector object construction in the GPU and NPU Attention model-runner constructors.
2. Defer `init_afd_connector()` until the end of each Attention model runner's `load_model()`, after any AFD/Ascend ubatch wrapper is installed.
3. Preserve the existing FFN initialization fallback in `start_ffn_server_loop()`.
4. Move the FFN loop-started log after the `start_ffn_server_loop` RPC succeeds, or change the earlier message so it does not imply readiness.

The end of Attention `load_model()` appears to be the latest safe coordination point:

- it allows Attention and FFN weight loading to overlap fully;
- connector initialization still completes before Attention memory profiling or any dummy/real model forward;
- connector memory remains visible to the later available-memory/KV-cache calculation;
- moving initialization to `initialize_from_config()` would be too late because `determine_available_memory()` may run a profile forward first.

Moving FFN connector initialization into its constructor would also allow parallel model loading, but it would allocate communicator resources before FFN weight loading. Delaying the Attention rendezvous avoids changing FFN weight-loading memory behavior.

### Lifecycle tests

Add GPU and NPU model-runner coverage that records the order of:

1. connector construction;
2. model weight loading;
3. optional wrapper installation;
4. connector initialization;
5. memory profiling or the first model forward.

Verify that connector initialization remains idempotent and cleanup is safe after partial startup failures.

### Hardware coverage

Validate all supported paths:

- CUDA with `P2pNcclAFDConnector`;
- Ascend NPU with `CAMP2pAFDConnector`;
- Ascend NPU with `CAMAsyncAFDConnector`.

For each role and rank, record timestamps for:

- model load begin/end;
- connector initialization begin/end;
- FFN connector loop start;
- Attention memory profile begin.

Expected results:

- Attention and FFN model-loading intervals overlap;
- all connector ranks rendezvous successfully after model loading;
- Attention profiling starts only after connector initialization;
- no startup OOM, NCCL/HCCL timeout, graph/warmup failure, or accuracy regression;
- total startup latency decreases by approximately the shorter of the two overlapping model-load phases.

This proposal is based on static lifecycle analysis plus the observed startup ordering. GPU and NPU end-to-end validation is still required.

ShwStone:
## Summary

Fixes #237: parallelize Attention and FFN model loading across AFD connectors.

The AFD Attention model runners previously called the collective
`connector.init_afd_connector()` from their constructors, which blocked inside
`init_device()` until every FFN rank finished loading its model and reached
connector-loop startup. FFN ranks stayed idle during most of the Attention
weight-loading phase, inflating total service startup latency.

This PR moves the cross-role connector rendezvous to the end of each Attention
model runner's `load_model()`, so Attention and FFN weight loading overlap
fully before any communication group is created.

## Changes

- **`AFDAttentionModelRunner` (CUDA)**: connector construction stays in
  `__init__`; `init_afd_connector()` is deferred to the end of `load_model()`,
  after the AFD ubatch wrapper is installed.
- **`AFDNPUAttentionModelRunner` (Ascend)**: same deferral for both
  `CAMP2pAFDConnector` and `CAMAsyncAFDConnector`, after the Ascend ubatch
  wrapper is installed.
- Both runners skip re-initialization when the connector is already
  initialized (mirrors the existing FFN-side guard in
  `start_ffn_server_loop()`), so repeated `load_model()` calls never re-enter
  the collective rendezvous. The connectors themselves remain idempotent.
- **`engine_core.py` patch**: the `AFD FFN EngineCore started; workers run
  connector loop.` log is now emitted only after the
  `start_ffn_server_loop` collective RPC succeeds, instead of before it.

### Why the end of Attention `load_model()` is safe

- Attention and FFN weight loading overlap completely.
- The rendezvous still completes before Attention memory profiling or any
  dummy/real forward, so connector memory is visible to the
  available-memory/KV-cache calculation.
- Initialization stays before `initialize_from_config()`, which is too late
  because `determine_available_memory()` may run a profile forward first.
- FFN weight-loading memory behavior is unchanged: FFN still initializes its
  connector from `initialize_from_config()`.

## Validation

- New GPU and NPU model-runner lifecycle tests record the order of connector
  construction, model weight loading, wrapper installation (ubatch and
  non-ubatch variants), and connector initialization
  (`test_attention_model_runner.py`, `test_npu_runtime.py`). Connector-level
  init idempotency remains covered by the existing connector unit tests.
- New `engine_core` tests assert the readiness log is emitted only after the
  `start_ffn_server_loop` RPC succeeds, and is absent when the RPC fails
  (`test_engine_core.py`).
- CPU-only unit suite (`pytest -q tests/unit -m "not gpu and not vllm_runtime"`)
  and `ruff check` / `ruff format --check` pass.

## Notes

Hardware end-to-end validation (CUDA with `P2pNcclAFDConnector`, Ascend NPU
with `CAMP2pAFDConnector` and `CAMAsyncAFDConnector`) is still required per the
issue's validation plan: attention and FFN model-load intervals should
overlap, connector ranks should rendezvous after model loading, Attention
profiling should start only after connector initialization, and total startup
latency should decrease by roughly the shorter of the two overlapping
model-load phases.

### DIFF
--- a/afd_plugin/v1/worker/attention_model_runner.py
+++ b/afd_plugin/v1/worker/attention_model_runner.py
@@ -164,6 +164,10 @@ class AFDAttentionModelRunner(GPUModelRunner):
             super().load_model(load_dummy_weights)
         if use_ubatching:
             self._install_afd_ubatch_wrapper()
+        # Wrapper installation is local and non-blocking; the connector
+        # rendezvous is the blocking cross-role collective, so it is
+        # deliberately last: it doubles as the "both roles finished loading
+        # weights" barrier before memory profiling.
         if not self.connector.is_initialized:
             self.connector.init_afd_connector()
 
--- a/afd_plugin/v1/worker/attention_model_runner.py
+++ b/afd_plugin/v1/worker/attention_model_runner.py
@@ -75,7 +75,8 @@ class AFDAttentionModelRunner(GPUModelRunner):
             self.vllm_config,
             self.afd_config,
         )
-        self.connector.init_afd_connector()
+        # The connector rendezvous is deferred to the end of ``load_model()``
+        # so Attention and FFN weight loading overlap; see that method.
         # TODO: Async GPU connector will be supported in the future
         assert self.connector.control_plane is not None, (
             "GPU model runner only supports control-plane-driven connectors"
@@ -149,12 +150,22 @@ class AFDAttentionModelRunner(GPUModelRunner):
         self.connector.control_plane.update_state_from_dp_metadata(payload)
         self.connector.control_plane.send_dp_metadata_list(payload)
 
+    # Patch reason: upstream load_model has no AFD connector lifecycle.
+    # Patch functionality: after native weight loading and any AFD ubatch
+    # wrapper installation, collectively initialize the AFD connector so
+    # Attention and FFN model loading overlap across roles. Connector
+    # initialization is skipped when it already succeeded (the FFN side uses
+    # the same guard), and the rendezvous completes before memory profiling or
+    # the first model forward. The connector itself is idempotent.
+    # Signature: matches upstream; no added parameters.
     def load_model(self, load_dummy_weights: bool = False) -> None:
         use_ubatching = bool(self.vllm_config.parallel_config.use_ubatching)
         with _use_afd_ubatch_wrapper_during_load(use_ubatching):
             super().load_model(load_dummy_weights)
         if use_ubatching:
             self._install_afd_ubatch_wrapper()
+        if not self.connector.is_initialized:
+            self.connector.init_afd_connector()
 
     def _install_afd_ubatch_wrapper(self) -> None:
         if isinstance(self.model, AFDUBatchWrapper):
--- a/afd_plugin/v1/worker/npu/attention_model_runner.py
+++ b/afd_plugin/v1/worker/npu/attention_model_runner.py
@@ -151,7 +151,8 @@ class AFDNPUAttentionModelRunner(NPUModelRunner):
                     f"{type(connector_extra_info).__name__}",
                 )
             self.afd_async_extra_info = connector_extra_info
-        self.connector.init_afd_connector()
+        # The connector rendezvous is deferred to the end of ``load_model()``
+        # so Attention and FFN weight loading overlap; see that method.
         self._is_warmup = False
         self._afd_is_graph_capturing = False
         self._afd_pending_metadata: AFDForwardContextMetadata | None = None
@@ -1534,10 +1535,21 @@ class AFDNPUAttentionModelRunner(NPUModelRunner):
         dp_size = int(self.vllm_config.parallel_config.data_parallel_size)
         return _make_uniform_dp_metadata(dp_size, int(num_tokens))
 
+    # Patch reason: upstream NPUModelRunner.load_model has no AFD connector
+    # lifecycle.
+    # Patch functionality: after native weight loading and any Ascend ubatch
+    # wrapper installation, collectively initialize the AFD connector so
+    # Attention and FFN model loading overlap across roles. Connector
+    # initialization is skipped when it already succeeded (the FFN side uses
+    # the same guard), and the rendezvous completes before memory profiling or
+    # the first model forward. The connector itself is idempotent.
+    # Signature: matches upstream; no added parameters.
     def load_model(self) -> None:
         super().load_model()
         if bool(self.vllm_config.parallel_config.use_ubatching):
             self._install_ascend_ubatch_wrapper()
+        if not self.connector.is_initialized:
+            self.connector.init_afd_connector()
 
     def _install_ascend_ubatch_wrapper(self) -> None:
         if isinstance(self.model, AscendUBatchWrapper):
--- a/afd_plugin/v1/worker/npu/attention_model_runner.py
+++ b/afd_plugin/v1/worker/npu/attention_model_runner.py
@@ -1548,6 +1548,10 @@ class AFDNPUAttentionModelRunner(NPUModelRunner):
         super().load_model()
         if bool(self.vllm_config.parallel_config.use_ubatching):
             self._install_ascend_ubatch_wrapper()
+        # Wrapper installation is local and non-blocking; the connector
+        # rendezvous is the blocking cross-role collective, so it is
+        # deliberately last: it doubles as the "both roles finished loading
+        # weights" barrier before memory profiling.
         if not self.connector.is_initialized:
             self.connector.init_afd_connector()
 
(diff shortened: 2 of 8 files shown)


afd_plugin/v1/worker/npu/attention_model_runner.py
```diff
@@ -1534,10 +1535,21 @@ def _build_capture_dp_metadata(self, num_tokens: int) -> DPMetadata | AFDDPMetad
...
+    # Attention and FFN model loading overlap across roles. Connector
+    # initialization is skipped when it already succeeded (the FFN side uses
+    # the same guard), and the rendezvous completes before memory profiling or
+    # the first model forward. The connector itself is idempotent.
+    # Signature: matches upstream; no added parameters.
     def load_model(self) -> None:
         super().load_model()
         if bool(self.vllm_config.parallel_config.use_ubatching):
             self._install_ascend_ubatch_wrapper()
+        if not self.connector.is_initialized:
```
ShwStone, line 1555: Is this the most proper place to put init connector inside load_model?

afd_plugin/v1/worker/attention_model_runner.py
```diff
@@ -149,12 +150,22 @@ def _send_dp_metadata(
...
+    # the same guard), and the rendezvous completes before memory profiling or
+    # the first model forward. The connector itself is idempotent.
+    # Signature: matches upstream; no added parameters.
     def load_model(self, load_dummy_weights: bool = False) -> None:
         use_ubatching = bool(self.vllm_config.parallel_config.use_ubatching)
         with _use_afd_ubatch_wrapper_during_load(use_ubatching):
             super().load_model(load_dummy_weights)
         if use_ubatching:
             self._install_afd_ubatch_wrapper()
+        if not self.connector.is_initialized:
```
ShwStone, line 171: Can we init ubatch wrapper before  connector?

ShwStone: Yes — and that is already the order in the code: the ubatch wrapper is installed first (`_install_afd_ubatch_wrapper()`), and only then is the connector initialized. This ordering is intentional:

- Wrapper installation is **local and non-blocking**: it wraps `self.model` (which only exists after `super().load_model()`) and configures the AFD context provider. It touches no communicators.
- `init_afd_connector()` is the **blocking cross-role collective** rendezvous. Doing it last means the local wrapper setup is never held hostage by the other role, and it follows the issue's proposed direction verbatim: "Defer `init_afd_connector()` until the end of each Attention model runner's `load_model()`, after any AFD/Ascend ubatch wrapper is installed."
- Nothing in `_install_afd_ubatch_wrapper()` / `_install_ascend_ubatch_wrapper()` requires connector communicators, so the order does not affect correctness.

I also added a short comment in both `load_model()` overrides stating this rationale so the ordering is self-documenting.

afd_plugin/v1/worker/npu/attention_model_runner.py
ShwStone: Yes — the end of `load_model()` is the latest safe coordination point, per the analysis in #237:

- **Earlier (previous behavior) is too early**: the constructor blocked inside `init_device()` on the cross-role rendezvous, which is exactly the serialization being fixed — FFN ranks sat idle until they finished weight loading and reached connector-loop startup.
- **Later is too late**: `initialize_from_config()` is preceded by `determine_available_memory()` in the EngineCore init, which may run a profile forward. The connector must already be initialized for profiling/warmup, and its communicator/HCCL-buffer memory must be visible to the available-memory/KV-cache calculation.
- **At the end of `load_model()`**: both roles have finished weight loading, so the Attention and FFN load intervals overlap fully; the rendezvous still completes before memory profiling or any dummy/real forward.

One nuance: `load_model()` can run with `load_dummy_weights=True` and may be re-entered (e.g., weight reload). The `is_initialized` guard — mirroring the existing FFN-side guard in `start_ffn_server_loop()` — makes repeated calls safe, and all three connectors' `init_afd_connector()` are themselves idempotent.

ShwStone:
Tested this branch on Ascend NPU (A3) with the `afd-eager-async-cam` E2E scenario (DeepSeek-V2-Lite, attention DP1TP2 + FFN DP2, `CAMAsyncAFDConnector`, eager).

The loading order now matches the expected behavior:

- The attention side finishes loading its model weights faster and then waits for the FFN ranks (the `init_afd_connector()` rendezvous now happens at the end of attention `load_model()`, as intended).
- Once the FFN server loop starts, the attention side immediately proceeds to KV cache estimation / available-memory profiling.
- The smoke completion request finished successfully end-to-end.

Startup no longer serializes FFN weight loading behind attention weight loading — the two roles load in parallel.

yenuo26:
@ShwStone We have recently set up a CI system. To ensure it runs properly, you may need to update your code.

PR #245, opened 2026-08-14
Add AFD support for CUDA GPU ModelRunnerV2

jiaran-king:
## Summary

This PR adds opt-in AFD support for vLLM 0.26.0 CUDA GPU ModelRunnerV2 while preserving the native ModelRunnerV2 execution architecture.

The implementation keeps native V2 ownership of request state, input preparation, Attention execution, KV-cache handling, sampling, output construction, DP dispatch, and CUDA graph management. AFD adds only the connector metadata and control-plane integration required around those native paths.

The existing V1 path remains the default.

## Architecture

```mermaid
flowchart TD
    SO["SchedulerOutput"] --> AW["AFDAttentionWorker"]
    AW --> V["validate_v2_config()"]
    V --> MR["AFDAttentionModelRunnerV2"]

    MR --> NV2["Native GPUModelRunnerV2.execute_model()"]
    MR --> MM["AFDMetadataMixin"]

    NV2 --> FC["Native ForwardContext"]
    MM --> FC
    FC --> MODEL["Existing AFD DeepSeek model integration"]

    MM --> CP["AFD control payload"]
    MODEL --> P2P["Existing synchronous P2P connector"]
    P2P --> FFN["Existing GPUFFNModelRunner"]

    NV2 --> GM["Native CUDA graph manager"]
    MR -.->|capture and replay hooks| GM
```

### Architecture walkthrough

1. **Worker initialization and validation**

   `validate_v2_config()` enforces only constraints required by the AFD V2 seam, while existing AFD topology/model registration and native vLLM configuration validation remain authoritative.

   `AFDAttentionWorker.init_device()` and `AFDFFNWorker.init_device()` temporarily substitute the runner class at the exact native construction seam. The rest of device initialization remains owned by the native Worker, and the original class symbol is restored in `finally`.

2. **Native ModelRunnerV2 execution**

   `AFDAttentionModelRunnerV2` delegates request processing to native `GPUModelRunnerV2.execute_model()`. Native vLLM continues to own persistent request and model state, input buffers and block tables, DP dispatch and execution descriptors, Attention and KV-cache work, sampling, and output construction. The AFD runner does not duplicate that pipeline.

3. **ForwardContext metadata integration**

   During native ForwardContext construction, `AFDMetadataMixin` installs the AFD metadata sidecar used by the existing model integration. It carries connector, transaction, token-length, stage, and DP control state without adding a ModelRunnerV2-specific model branch.

4. **Attention-to-FFN data path**

   The existing synchronous `P2pNcclAFDConnector` transports activations to the existing `GPUFFNModelRunner`. No V2-specific FFN runner, connector API, or protocol field is introduced.

5. **CUDA graph ownership**

   Native vLLM retains ownership of graph descriptors, descriptor selection, capture, replay, and graph-managed buffers. Narrow AFD lifecycle hooks publish matching control state around warmup, capture, and replay, with connector control operations outside the `torch.cuda.graph` body.

## Hardware-validated evidence matrix

| AFD topology | Data parallel | Tensor parallel | Execution mode |
|---|---:|---:|---|
| 1A1F | DP1 | TP1 | eager |
| 2A2F | DP2 | TP1 | eager |
| 2A2F | DP1 | TP2 | eager |
| 1A1F | DP1 | TP1 | FULL_DECODE_ONLY |
| 2A2F | DP2 | TP1 | FULL_DECODE_ONLY |
| 2A2F | DP1 | TP2 | FULL_DECODE_ONLY |

These six rows are representative hardware evidence, not a production topology allowlist. Other legal DP/TP layouts that satisfy the existing AFD topology contract and native vLLM validation use the same runtime path, but are not claimed as hardware-validated by this PR.

The validated rows use CUDA, the existing AFD DeepSeek model registration (tested with DeepSeek-V2-Lite), static expert parallelism, synchronous `P2pNcclAFDConnector`, `compute_gate_on_attention=false`, and no DBO.

### Dedicated V1, V2, and shared metadata modules

- `attention_model_runner.py` contains the V1 Attention runner and its V1-specific execution and ubatching integration.
- `attention_model_runner_v2.py` contains `validate_v2_config()` and `AFDAttentionModelRunnerV2`, and imports the native V2 runner directly.
- `attention_metadata.py` contains the runner-neutral `AFDMetadataMixin` and shared rank, token, ForwardContext, and DP-metadata helpers. It imports neither runner module nor `ubatch_wrapper.py`.
- `ubatch_wrapper.py` preserves its existing `build_ubatch_dp_metadata_list` import path while sourcing the implementation from the shared metadata module.
- AFD metadata is injected through native ForwardContext construction.
- Native `execute_model()`, input preparation, sampling, output, and graph-manager implementations are not copied.
- The existing `_MODEL_REGISTRATIONS` table remains the model boundary; no checkpoint-specific architecture count or configuration fingerprint is added.

### Worker construction and FFN reuse

vLLM 0.26.0 does not expose a model-runner factory in `Worker.init_device()`. Attention and FFN workers therefore use temporary exact-module substitution while delegating the full initialization lifecycle to the native Worker. This constructs exactly one runner and restores the native symbol in `finally`.

The FFN side continues to use the existing `GPUFFNModelRunner`; no V2 FFN runner is added.

### CUDA graph integration

`FULL_DECODE_ONLY` support adds narrow hooks for graph warmup, formal capture, and replay. AFD control-plane work remains outside the `torch.cuda.graph` body, and native graph descriptors and graph-manager ownership are preserved.

Production validation does not pin an experimental capture size. Generic capture-size and descriptor consistency remain owned by native vLLM; AFD restricts only the graph mode required by its implemented lifecycle seam.

### DP and TP routing

The hardware-tested DP2/TP1 and DP1/TP2 rows preserve separate contracts for scheduler real tokens, native DP execution/padded tokens, AFD transmitted rows, per-rank control payloads, and Attention-to-FFN peers. The existing connector and protocol are reused without public API or wire-format changes.

## Runtime constraints

The current AFD ModelRunnerV2 seam requires:

- CUDA;
- synchronous `P2pNcclAFDConnector`;
- `compute_gate_on_attention=false`;
- pipeline and context parallel sizes equal to 1;
- each role's configured rank count equal to native DP × TP;
- a model present in the existing AFD model registration table;
- static expert parallelism without elastic EP/EPLB or sequence-parallel MoE;
- no DBO/ubatching;
- eager execution or `FULL_DECODE_ONLY` CUDA graph mode.

Native vLLM remains authoritative for its own feature/configuration validation. NPU ModelRunnerV2 and other RFC non-goals remain out of scope.

## Validation

The final PR snapshot was replayed onto the current `main` E2E harness and validated against the installed vLLM 0.26.0 runtime.

- The final validator/model-runner focused selection passed 64/64 after simplification.
- 26 FFN regression tests passed with zero failures.
- The 95-test E2E-runner consumer suite passed after harness adaptation.
- Local `compileall`, `git diff --check`, and the 88-column scan passed for the final three-file validator amend.
- The DeepSeek-V2-Lite E2E module collected 11 scenarios, including the six V2 evidence rows.
- Current-harness GPU replay passed for representative DP2 eager and TP2 `FULL_DECODE_ONLY` graph scenarios.
- The final graph run completed 7/7 GSM8K requests and left no target PID, port, GPU process, or SysV shared-memory residue.
- Earlier development evidence covers all six rows, deterministic comparisons, non-uniform DP/idle continuation, direct rank/token/control observations, graph warmup/capture/replay, serving, profiling, GSM8K, and fatal-path cleanup.

Performance measurements are characterization only. This PR does not claim a performance acceptance threshold or change the default execution path.

## Compatibility and review notes

- V1 remains the default unless ModelRunnerV2 is explicitly selected.
- No ModelRunnerV2-specific model-layer branch was added.
- No connector implementation, public connector API, or wire protocol was changed.
- No native ModelRunnerV2 execution method was copied.
- Pre-existing FFN blocking-receive shutdown semantics are unchanged by this PR and are not claimed as newly resolved here.

### DIFF
--- a/afd_plugin/v1/worker/attention_metadata.py
+++ b/afd_plugin/v1/worker/attention_metadata.py
@@ -1,0 +1,313 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
+"""Shared AFD runner metadata and rank helpers."""
+
+from __future__ import annotations
+
+import torch
+from vllm.config import CUDAGraphMode, VllmConfig
+from vllm.distributed.parallel_state import get_world_group
+from vllm.forward_context import DPMetadata, ForwardContext
+from vllm.v1.worker.ubatch_utils import UBatchSlices
+
+from afd_plugin.connectors import (
+    AFDControlPayload,
+    AFDDPMetadata,
+    AFDForwardContextMetadata,
+)
+
+
+class AFDMetadataProviderMixin:
+    """AFD metadata/control plumbing shared by the V1 and V2 runners.
+
+    This mixin owns only the connector-facing metadata sidecar and DP control
+    payload. ``AFDAttentionModelRunner`` consumes it alongside the existing V1
+    Attention/graph lifecycle, while ``AFDAttentionModelRunnerV2`` consumes it
+    alongside native MRV2 execution. Native runner state and model ownership
+    remain in each vLLM base class.
+    """
+
+    def build_afd_metadata(
+        self,
+        ubatch_slices: UBatchSlices | None,
+        num_tokens_unpadded: int,
+    ) -> AFDForwardContextMetadata:
+        """Build the AFD sidecar for one transaction without sending control.
+
+        Without multiple ubatch stages, ``num_tokens_unpadded`` is the token count
+        supplied by the caller.  With multiple stages, native slice lengths are
+        authoritative for both token fields.  The returned metadata references
+        the runner's connector and owns a fresh transaction id; this method does
+        not mutate a ``ForwardContext`` or perform control- or data-plane I/O.
+        """
+
+        if ubatch_slices and len(ubatch_slices) > 1:
+            tokens_start_loc = [ub.token_slice.start for ub in ubatch_slices]
+            requests_start_loc = [ub.request_slice.start for ub in ubatch_slices]
+            tokens_lens = [ub.num_tokens for ub in ubatch_slices]
+            tokens_unpadded_lens = [int(ub.num_tokens) for ub in ubatch_slices]
+            num_stages = len(ubatch_slices)
+        else:
+            tokens_start_loc = [0]
+            requests_start_loc = [0]
+            tokens_lens = [num_tokens_unpadded]
+            tokens_unpadded_lens = [num_tokens_unpadded]
+            num_stages = 1
+
+        return AFDForwardContextMetadata(
+            tokens_start_loc=tokens_start_loc,
+            requests_start_loc=requests_start_loc,
+            stage_idx=0,
+            connector=self.connector,
+            tokens_lens=tokens_lens,
+            num_stages=num_stages,
+            transaction_id=self._next_afd_transaction_id(),
+            tokens_unpadded_lens=tokens_unpadded_lens,
+        )
+
+    def send_dp_metadata(
+        self,
+        dp_metadata: DPMetadata | AFDDPMetadata | None,
+        ubatch_slices: UBatchSlices | None,
+    ) -> None:
+        """Publish DP control for the pending AFD transaction.
+
+        ``dp_metadata`` contains the native rank-specific counts selected for the
+        current execution; graph lifecycle callers may supply metadata built from
+        a padded descriptor.  ``ubatch_slices`` instead selects per-stage control
+        payloads.  This method updates connector control state and then sends one
+        payload; it does not allocate a transaction or perform data-plane work.
+        """
+
+        assert self.connector.control_plane is not None, (
+            "send_dp_metadata needs control plane driven connectors"
+        )
+
+        if ubatch_slices and len(ubatch_slices) > 1:
+            dp_metadata_list = {
+                idx: metadata
+                for idx, metadata in enumerate(
+                    build_ubatch_dp_metadata_list(self.vllm_config, ubatch_slices),
+                )
+            }
+        else:
+            dp_metadata = self._ensure_dp_metadata(dp_metadata)
+            dp_metadata_list = {0: dp_metadata}
+        is_warmup = self._is_warmup
+        # Keep the V1 object.__new__ test seam and older graph lifecycle
+        # callers compatible with runners created before this mixin existed.
+        is_graph_capturing = getattr(self, "_afd_is_graph_capturing", False)
+        payload = AFDControlPayload(
+            dp_metadata_list=dp_metadata_list,
+            is_graph_capturing=is_graph_capturing,
+            is_warmup=is_warmup,
+        )
+        self.connector.control_plane.update_state_from_dp_metadata(payload)
+        self.connector.control_plane.send_dp_metadata_list(payload)
+
+    def _ensure_dp_metadata(
+        self,
+        dp_metadata: DPMetadata | AFDDPMetadata | None,
+    ) -> DPMetadata | AFDDPMetadata:
+        if dp_metadata is not None:
+            return dp_metadata
+
+        dp_size = int(self.vllm_config.parallel_config.data_parallel_size)
+        if dp_size != 1:
+            raise RuntimeError("AFD expected vLLM DPMetadata for attention DP > 1")
+
+        if self._afd_pending_metadata is None:
+            raise RuntimeError("AFD metadata is not available for DP metadata fallback")
+        if len(self._afd_pending_metadata.tokens_lens) != 1:
+            raise RuntimeError("AFD DP=1 fallback only supports one stage")
+
+        num_tokens = int(self._afd_pending_metadata.tokens_lens[0])
+        num_tokens_across_dp_cpu = torch.tensor(
+            [num_tokens],
+            dtype=torch.int32,
+            device="cpu",
+        )
+        return AFDDPMetadata(
+            num_tokens_across_dp_cpu=num_tokens_across_dp_cpu,
+            max_tokens_across_dp_cpu=torch.max(num_tokens_across_dp_cpu),
+        )
+
+    def build_capture_dp_metadata(
+        self,
+        num_tokens: int,
+    ) -> DPMetadata | AFDDPMetadata:
+        """Build, but do not send, uniform padded graph DP metadata.
+
+        ``num_tokens`` is the padded CUDA graph token count executed on every DP
+        rank during capture or replay.  The returned metadata is suitable for
+        :meth:`send_dp_metadata`; this method has no transaction, context, or
+        connector side effects.
+        """
+
+        dp_size = int(self.vllm_config.parallel_config.data_parallel_size)
+        num_tokens_across_dp_cpu = torch.full(
+            (dp_size,),
+            int(num_tokens),
+            dtype=torch.int32,
+            device="cpu",
+        )
+        if dp_size > 1:
+            return DPMetadata.make(
+                self.vllm_config.parallel_config,
+                int(num_tokens),
+                num_tokens_across_dp_cpu,
+            )
+        max_tokens_across_dp_cpu = torch.max(num_tokens_across_dp_cpu)
+        return AFDDPMetadata(
+            num_tokens_across_dp_cpu=num_tokens_across_dp_cpu,
+            max_tokens_across_dp_cpu=max_tokens_across_dp_cpu,
+        )
+
+    def _afd_num_tokens_for_context(self, forward_context: ForwardContext) -> int:
+        return _forward_context_num_tokens(forward_context, self.vllm_config)
+
+    def install_afd_metadata_on_forward_context(
+        self,
+        forward_context: ForwardContext,
+    ) -> None:
+        """Install transaction metadata and publish its matching control.
+
+        On the ordinary eager path, the method creates a pending transaction from
+        the current context, writes it to ``ForwardContext.additional_kwargs``,
+        and sends the context's native DP control before the model/data-plane
+        forward.  Graph lifecycles may instead pre-stage the transaction and
+        control payload; their suppression scope makes this method attach the
+        staged sidecar without a duplicate send.  A FULL graph context that was
+        not pre-sent uses its padded descriptor count for DP control.  Ubatch
+        child contexts reuse their parent transaction without another send, and
+        native FULL replay uses its pre-replay hook because it creates no context.
+        """
+
+        if forward_context.additional_kwargs is None:
+            forward_context.additional_kwargs = {}
+        existing_metadata: AFDForwardContextMetadata | None = (
+            forward_context.additional_kwargs.get("afd_metadata")
+        )
+        if existing_metadata is not None and _is_ubatch_child_afd_context(
+            forward_context,
+            existing_metadata,
+        ):
+            return
+
+        if self._afd_pending_metadata is None:
+            self._afd_pending_metadata = self.build_afd_metadata(
+                forward_context.ubatch_slices,
+                self._afd_num_tokens_for_context(forward_context),
+            )
+        if self._afd_pending_metadata is not None:
+            forward_context.additional_kwargs["afd_metadata"] = (
+                self._afd_pending_metadata
+            )
+        if getattr(self, "_afd_suppress_metadata_send", False):
+            return
+        dp_metadata = forward_context.dp_metadata
+        ubatch_slices = forward_context.ubatch_slices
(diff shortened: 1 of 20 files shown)


jiangkuaixue123:
Could we consider adding a dedicated model runner file for `ModelRunnerV2` instead of adding the V2 implementation to the existing V1 `attention_model_runner.py`? vLLM keeps the two implementations separate in `vllm/v1/worker/gpu_model_runner.py` (V1) and `vllm/v1/worker/gpu/model_runner.py` (V2). It would be good to follow the same structure here so that the V1 and V2 code paths remain clearly separated and easier to maintain.

jiaran-king:
> Could we consider adding a dedicated model runner file for `ModelRunnerV2` instead of adding the V2 implementation to the existing V1 `attention_model_runner.py`? vLLM keeps the two implementations separate in `vllm/v1/worker/gpu_model_runner.py` (V1) and `vllm/v1/worker/gpu/model_runner.py` (V2). It would be good to follow the same structure here so that the V1 and V2 code paths remain clearly separated and easier to maintain.

Thanks for the suggestion. We have now separated the two implementations:
AFDAttentionModelRunner remains in attention_model_runner.py.
AFDAttentionModelRunnerV2 has been moved to the dedicated attention_model_runner_v2.py.
To avoid duplicating the AFD-specific metadata and control-plane plumbing, the logic shared by both runners has been extracted into attention_metadata.py.
The V1 and V2 execution paths, model-runner state, input preparation, and graph lifecycles remain separate and continue to be owned by their respective vLLM base classes. Only the connector-facing AFD metadata contract is shared.

jiangkuaixue123 (commented):
Reviewed at `85764854272696126f01b4ffb6c0bd058e161ae9` with InferMatrixCopilot (Direct mode).

I did not find an actionable correctness issue in the inspected ModelRunnerV2 integration. I focused on the native V2 delegation boundary, ForwardContext metadata/control ordering, DP/TP token accounting, worker runner substitution/restoration, CUDA graph warmup/capture/replay hooks, V1 compatibility, and FFN runner reuse. The earlier request to separate the V1 and V2 runner implementations is addressed by `attention_model_runner_v2.py` plus the shared `attention_metadata.py` layer.

Validation performed:

- `git diff --check origin/main...HEAD`: passed
- `python3 -m compileall -q afd_plugin`: passed
- GitHub DCO and lint checks: passed
- Compared the capture/replay seam against the pinned vLLM upstream commit `568afb3a13806beb53bb2e6bd518269357b237c0`; the expected one warmup + one formal capture callback per descriptor and `run_fullgraph` hook match that snapshot
- Focused pytest could not be rerun in the temporary review checkout because this environment does not have pytest/vLLM installed; the PR's reported GPU/runtime matrix therefore remains externally supplied validation evidence

This is a code-review pass on the frozen head, not an independent rerun of the CUDA hardware matrix.

jiangkuaixue123 (commented):
One API-design comment on the shared metadata abstraction.

afd_plugin/v1/worker/attention_metadata.py
```diff
@@ -0,0 +1,270 @@
...
+from vllm.v1.worker.ubatch_utils import UBatchSlices
+
+from afd_plugin.connectors import (
+    AFDControlPayload,
+    AFDDPMetadata,
+    AFDForwardContextMetadata,
+)
+
+
+class AFDMetadataMixin:
```
jiangkuaixue123: Could we reconsider the name `AFDMetadataMixin`? More importantly, this class is being introduced as a shared, public abstraction for both the V1 and V2 runners, but all of its methods currently use private (`_`-prefixed) names. If these methods define the abstraction's supported contract, they should have public API semantics rather than appearing to be internal implementation details. Please rename the relevant methods accordingly and add docstrings that clearly describe their responsibilities, inputs, outputs, and lifecycle expectations.

afd_plugin/v1/worker/attention_worker.py
```diff
@@ -37,28 +43,62 @@ def __init__(
             is_driver_worker,
         )
 
+    # Patch reason: vLLM 0.26.0 constructs its runner inside Worker.init_device
```
jiangkuaixue123, line 48: This does represent a real compatibility patch: the override wraps upstream `Worker.init_device()` and temporarily substitutes the runner class, so the structured patch metadata and `PATCH START/END` markers are useful and consistent with the repository guidelines. However, the same rationale is repeated in the header, the synchronization note, and the comment immediately before the substitution. Please retain the required patch reason, functionality, signature, upstream source, and removal plan here, but consolidate the duplicate commentary around lines 62–64 and 91–94 into one concise explanation. That would preserve the upgrade context while reducing visual noise in this small method.

jiangkuaixue123 (commented):
One maintainability comment on the patch documentation in `attention_worker.py`.

afd_plugin/v1/worker/ffn_worker.py
```diff
@@ -55,28 +62,59 @@ def __init__(
...
         assert_compatible_afd_stack(
             self.vllm_config,
             caller="AFDFFNWorker.init_device",
             expected_role="ffn",
         )
         if self.use_v2_model_runner:
-            raise RuntimeError(
-                "AFD FFN runtime currently supports only the vLLM v1 "
-                "GPUModelRunner interface; set VLLM_USE_V2_MODEL_RUNNER=0",
+            validate_v2_config(
```
jiangkuaixue123: `GPUFFNModelRunner` does not inherit from or otherwise implement the native ModelRunnerV2 execution contract; this branch appears to be needed only because `Worker.init_device()` uses a different module-level runner construction seam when `use_v2_model_runner` is enabled. Could we make that distinction explicit in the surrounding comments and move `validate_v2_config` out of `attention_model_runner_v2.py` into a shared validation or compatibility module? The current dependency makes the FFN runtime appear coupled to the attention-side V2 runner implementation, even though FFN execution remains connector-driven and uses its own runner.

jiangkuaixue123 (commented):
One architectural clarification on the FFN worker's ModelRunnerV2 compatibility path.

afd_plugin/v1/worker/attention_model_runner_v2.py
```diff
@@ -0,0 +1,406 @@
...
+                    afd_runner._is_warmup = False
+                    afd_runner._afd_is_graph_capturing = False
+                    afd_runner._send_dp_metadata(
+                        afd_runner._build_capture_dp_metadata(padded_tokens),
+                        None,
+                    )
+                return original_run_fullgraph(self, desc)
+                # ### PATCH END: publish one AFD pre-replay payload.
+
+            v2_cudagraph_utils.CudaGraphManager.run_fullgraph = run_fullgraph
```
jiangkuaixue123: These nested patch regions target different upstream owners: the outer override patches `GPUModelRunnerV2.execute_model`, while the inner closure temporarily replaces the class-level `CudaGraphManager.run_fullgraph`. Beyond making the patch boundaries difficult to audit, changing a global class method on every execution is not concurrency- or reentrancy-safe; another runner could observe the temporary closure, or a nested invocation could capture and later restore the wrong "original" method. The current tests verify restoration after normal and exceptional execution, but do not cover concurrent or nested calls. Could we separate each upstream patch into an independently scoped helper/context manager and avoid the class-level mutation, or explicitly enforce, document, and test the required single-runner, non-reentrant invariant?

jiangkuaixue123 (commented):
One correctness and maintainability concern about the nested runtime patch in `execute_model()`.

afd_plugin/v1/worker/attention_metadata.py
jiaran-king: Thanks. We renamed the class to AFDMetadataProviderMixin, made the supported V1/V2 contract methods public, and added lifecycle and token-semantics docstrings. Internal helpers remain private. We also replaced private provider access with a typed ForwardContext installer callable.

afd_plugin/v1/worker/attention_worker.py
jiaran-king: Thanks. We consolidated the synchronization and substitution rationale into the structured patch header and removed the duplicate inline comments. The required patch metadata and PATCH markers remain unchanged; runtime behavior is unaffected.

afd_plugin/v1/worker/ffn_worker.py
jiaran-king: Thanks. We clarified that use_v2_model_runner only selects the module-level construction seam used by Worker.init_device(); GPUFFNModelRunner remains connector-driven and does not implement native ModelRunnerV2 execution. We also moved the shared validator to afd_plugin.validation as validate_gpu_model_runner_v2_config().

afd_plugin/v1/worker/attention_model_runner_v2.py
jiaran-king: Thanks. We moved the replay patch into a dedicated _use_afd_fullgraph_replay_hook() context manager that wraps only the current ModelCudaGraphManager instance; the class method is no longer modified. It restores the exact instance state, rejects same-manager reentry, and allows different manager scopes to overlap independently. We added normal, exceptional, overlapping, nested, and pre-manager profile coverage, plus a passing four-GPU DP2 graph E2E run.

afd_plugin/v1/worker/attention_model_runner_v2.py
```diff
@@ -0,0 +1,398 @@
...
+        is_profile: bool = False,
+    ) -> ModelRunnerOutput | IntermediateTensors | None:
+        # ### PATCH START: scope AFD metadata provider/replay and profiler step.
+        step_afd_gpu_profiler(self.prof)
+        use_fullgraph_replay_hook = (
+            self.vllm_config.compilation_config.cudagraph_mode
+            == CUDAGraphMode.FULL_DECODE_ONLY
+            and self.cudagraph_manager is not None
+        )
+        previous_metadata = self._afd_pending_metadata
```
jiangkuaixue123: Could we encapsulate the temporary AFD execution state and replay-hook setup in a context manager? The multiple `previous_*` variables, conditional replay scope, metadata-provider scope, and matching `finally` restoration make it difficult to identify the actual `execute_model()` flow and verify that every temporary state is restored correctly. A dedicated execution context would make the lifecycle and exception-safety guarantees explicit, while keeping this method focused on delegating to the native runner.

afd_plugin/v1/worker/attention_model_runner_v2.py
```diff
@@ -0,0 +1,398 @@
...
+                "AFD ModelRunnerV2 CUDA Graph expected at least one FULL "
+                "capture descriptor",
+            )
+        for desc in capture_descs:
+            if desc.cg_mode != CUDAGraphMode.FULL:
+                raise RuntimeError(
+                    "AFD ModelRunnerV2 CUDA Graph expected only FULL capture "
+                    "descriptors",
+                )
+        expected_events = [
```
jiangkuaixue123: Could we encapsulate the capture event sequencing and validation in a small tracker/helper? `expected_events`, `event_index`, the nested hook, and the final count check collectively implement a state machine that assumes each descriptor is processed exactly twice in warmup/capture order. Keeping that protocol implicit inside `capture_model()` makes this version-sensitive path difficult to review and maintain. A named abstraction would make the upstream assumptions and state transitions explicit while leaving `capture_model()` focused on lifecycle orchestration.

afd_plugin/v1/worker/attention_model_runner_v2.py
jiaran-king: Thanks. We added a private _use_afd_execution_context() context manager that encapsulates replay-hook selection, the metadata-provider scope, and temporary AFD state preservation/restoration. execute_model() is now focused on the profiler step and delegation to native super().execute_model(). The context restores all temporary state on both success and exception paths without changing execution order or public APIs. Focused unit tests and the four-GPU afd-v2-graph-dp2 E2E passed.

afd_plugin/v1/worker/attention_model_runner_v2.py
jiaran-king: Thanks. We added a private _AFDCaptureEventTracker that explicitly models one warmup followed by one formal-capture event per FULL descriptor, with checks for extra, missing, reordered, and shape-mismatched events. We also moved the temporary prepare_inputs_to_capture patch into _use_afd_capture_input_preparation(). capture_model() now focuses on descriptor validation, lifecycle scopes, native capture delegation, and final tracker verification. Focused unit tests and the four-GPU afd-v2-graph-dp2 E2E passed.

PR #248, opened 2026-08-14
[Bugfix]Fix async CAM first-request deadlock

### LINKED ISSUE #247: [Bug]: Async CAM first request hangs with Attention DP2 until a second request is submitted
### Current environment

- afd-plugin: latest `main` (`714fc298d01591ab6d9fdb44dc1631089e349df9` at the time of testing)
- vLLM: 0.26.0
- Model: DeepSeek-V2-Lite
- Connector: `CAMAsyncAFDConnector`
- Attention topology: DP2 TP1
- FFN topology: DP1 TP1
- AFD MoE ubatching: disabled
- FlashComm1: disabled
- Native DBO: disabled

The issue reproduces with both `ASCEND_LAUNCH_BLOCKING=0` and `ASCEND_LAUNCH_BLOCKING=1`. Enabling blocking only made the actual blocking operation easier to locate.

### Reproduction

1. Start the 2-Attention/1-FFN topology above.
2. Submit exactly one request after startup.
3. The request hangs indefinitely.
4. Submit one additional request.
5. The first request immediately resumes.
6. After this one-time unblock, subsequent lm-eval requests complete normally when submitted serially.

The equivalent 1-Attention/1-FFN topology completes normally.

### Actual behavior

Process sampling with `py-spy` shows:

```text
Active Attention worker: blocked in async_dispatch_send
FFN worker:              blocked in async_dispatch_recv
Other Attention worker: idle, waiting for an engine request
```

No Python exception is raised at the original blocking point.

### Expected behavior

A request routed to one Attention DP replica should complete independently. The first request after a cold start should not require a second client request to make progress.

### Preliminary analysis

This appears to be a one-time CAM initialization or rendezvous problem rather than a requirement for concurrent requests on every dispatch round. Connector initialization creates the HCCL process group and buffers, while profiling skips the CAM operators. The first live request is therefore the first real CAM operation.

One possibility is that the second request reaches the previously idle Attention DP and completes lazy CAM/HCCL operator-resource initialization. This requires confirmation from first-request rank logs and the CAM operator contract.

### Before submitting

- [x] I confirmed that this reproduces on the latest afd-plugin `main` targeting vLLM 0.26.0.

jiangkuaixue123:
## Summary

- preserve AFD's independent async-DP engine semantics after vLLM-Ascend platform patching
- restore matched async CAM connector I/O during startup profile runs
- add lifecycle and profile-I/O regression coverage

Fixes #247.

## Root cause

The v0.26 upgrade introduced two interacting regressions:

1. vLLM-Ascend installs its balance-schedule wrapper after the AFD general plugin, replacing `EngineCoreProc.run_engine_core`. The subprocess target was therefore captured as the Ascend/upstream wrapper and Attention DP ranks instantiated `DPEngineCoreProc`, bypassing AFD's async-DP selection.
2. AFD began skipping CAM connector I/O during `profile_run()`. In the known-good v0.19.1 stack, all Attention ranks participate in matched profile traffic with the FFN daemon before serving. Removing that startup lifecycle step deferred CANN/CAM communicator initialization until the first real request; with DP2/TP1 only one Attention replica entered the first request, so initialization waited for the other rank and the request completed only after a second request arrived.

This change makes the async-DP patch installer idempotent and reapplies it only for NPU async Attention after the final Ascend config is built, before vLLM captures process targets. It also restores matched profile CAM traffic.

## Validation

NPU DeepSeek-V2-Lite, eager, `ASCEND_LAUNCH_BLOCKING=1`, one cold `max_tokens=32` completion, no second request:

- A DP2/TP1; F DP2/TP1/EP2: HTTP 200 in about 5 seconds.
- A DP1/TP2 with FlashComm1; F DP2/TP1/EP2 without FlashComm1: HTTP 200 in about 7 seconds.
- Both returned non-empty/correct completion text.
- Focused unit tests: 52 passed.
- Ruff check and format check passed for all changed files.
- `git diff --check` passed.

The existing NPU shutdown path can still report allocator/process-group cleanup timeouts after successful inference; this is independent of the first-request deadlock and is not changed here.

## Impact

Async-DP remains enabled; this does not revert to native DP wave/FIRST_REQ scheduling. Non-AFD configs and the FFN Ascend scheduling path are not rebound by the late installer.

### DIFF
--- a/afd_plugin/compat/npu/__init__.py
+++ b/afd_plugin/compat/npu/__init__.py
@@ -21,6 +21,7 @@ from afd_plugin.compat.npu.ops import (
 from afd_plugin.compat.npu.runtime import (
     apply_afd_ascend_config_patch_if_needed,
     apply_afd_ascend_patches_if_needed,
+    apply_afd_async_dp_engine_patch_if_needed,
     ascend_forward_context,
     fail_if_unsupported_npu_afd_features,
     fix_all2all_backend_for_afd,
@@ -30,6 +31,7 @@ from afd_plugin.compat.npu.runtime import (
 __all__ = [
     "apply_afd_ascend_config_patch_if_needed",
     "apply_afd_ascend_patches_if_needed",
+    "apply_afd_async_dp_engine_patch_if_needed",
     "ascend_forward_context",
     "AFD_ASCEND_OPS_NAMESPACE",
     "AFD_ASCEND_VENDOR_NAME",
--- a/afd_plugin/compat/npu/runtime.py
+++ b/afd_plugin/compat/npu/runtime.py
@@ -4,6 +4,8 @@
 
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
 from afd_plugin.compat.npu.feature_validation import (
     fail_if_unsupported_npu_afd_features,
 )
@@ -14,6 +16,10 @@ from afd_plugin.compat.npu.runtime_config import (
     fix_all2all_backend_for_afd,
     npu_afd_num_ubatches,
 )
+from afd_plugin.config import is_afd_async_dp, parse_optional_afd_config
+
+if TYPE_CHECKING:
+    from vllm.config import VllmConfig
 
 _PATCHES_APPLIED = False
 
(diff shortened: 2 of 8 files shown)


jiangkuaixue123:
cc @ShwStone

PR #253, opened 2026-08-17
fix(recipe): correct DeepSeek-V2-Lite proxy path

like955:
## Purpose

Fix the proxy server path in the DeepSeek-V2-Lite 2P1A1F eager and
CUDA graph recipes.

The recipes currently resolve the proxy path to:

    recipe/tools/proxy_server.py

which does not exist. The intended file is:

    tools/proxy_server.py

The path was valid when introduced in commit 0256946, but commit 4ec514b
nested the GPU recipes under an additional connector directory without
updating the relative path.

## Issue

- Related issue(s): N/A

## Scope

- In scope:
  - Correct the proxy path in the eager 2P1A1F recipe.
  - Correct the proxy path in the graph 2P1A1F recipe.
- Out of scope:
  - Changes to proxy behavior.
  - Changes to AFD workers or connectors.
  - Other recipe refactoring.

## Implementation Notes

Add one parent-directory traversal to both proxy paths so they resolve
to the repository-level `tools/proxy_server.py`.

No vLLM source or third-party code is changed.

## Test Plan

CPU-only:

- `bash -n` on both modified recipes.
- Resolve both proxy paths and verify the target exists.
- `uv run pytest`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run pre-commit run --all-files`

GPU-gated:

- Not required for path resolution; the original failure occurs before
  the proxy process can start.

## Test Result

- Proxy paths resolve to tools/proxy_server.py for both recipes.
- uv run pytest --ignore=tests/e2e: 263 passed, 63 skipped.
- uv run ruff check .: passed.
- uv run ruff format --check .: passed.
- Targeted pre-commit hooks: passed.
- git diff --check: passed.
- GPU E2E was not rerun because this change only corrects shell-level path resolution.

## Docs Impact

- Files updated: two executable recipe scripts.
- No documentation update is required because the user-facing command
  and recipe behavior are unchanged.

---

<details>

- [x] Scope is bounded.

</details>

### DIFF
--- a/recipe/gpu/P2pNcclAFDConnector/deepseek_v2_lite/prefill_decode_disaggregation/2p1a1f_eager_dbo.sh
+++ b/recipe/gpu/P2pNcclAFDConnector/deepseek_v2_lite/prefill_decode_disaggregation/2p1a1f_eager_dbo.sh
@@ -112,7 +112,7 @@ CUDA_VISIBLE_DEVICES=3 uv run vllm serve "$MODEL_PATH" \
 # decoder pulls the KV. Send benchmark/client traffic to 18305, NOT to 18303.
 # Wait for the prefill/attn/ffn servers to print "Application startup complete"
 # before this proxy starts accepting traffic.
-uv run python "$SCRIPT_DIR/../../../../tools/proxy_server.py" \
+uv run python "$SCRIPT_DIR/../../../../../tools/proxy_server.py" \
     --host 127.0.0.1 \
     --port 18305 \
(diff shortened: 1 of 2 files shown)


jiangkuaixue123:
Thanks! I see DCO failed.

like955:
Thanks for pointing that out. I've added the DCO sign-off and force-pushed the amended commit.

jiangkuaixue123:
LGTM！ Thanks very much

PR #254, opened 2026-08-17
[WIP][FEAT][NPU]Add DSV4 CAM async connector

bjf-frz:
PLEASE FILL IN THE PR DESCRIPTION AND MAKE SURE THE CHECKLIST ITEMS HAVE BEEN CONSIDERED.

## Issue

- Related issue(s): #

## Scope

- In scope:
- Out of scope:

## Docs Impact

- Files updated:
- If none, reason:

---

<details>

- [ ] Purpose is clear and linked to public context when possible.
- [ ] Scope is bounded.
- [ ] Compatibility with vLLM v0.26.0 is considered.
- [ ] No changes are made to the vLLM source checkout.
- [ ] Plugin-owned classes or explicit dotted class paths are preferred over monkey patches.
- [ ] Any compat shim or monkey patch is isolated, idempotent, version-guarded, documented, and tested.
- [ ] Imports remain CPU-safe; CUDA-heavy work is delayed or GPU-gated.
- [ ] Validation evidence is included, including skipped GPU tests when applicable.

</details>

### DIFF
--- a/afd_plugin/compat/npu/ops.py
+++ b/afd_plugin/compat/npu/ops.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+import ctypes
 import os
 from functools import lru_cache
 from pathlib import Path
@@ -16,6 +17,10 @@ CAM_DISPATCH_SEND = "async_dispatch_send"
 CAM_DISPATCH_RECV = "async_dispatch_recv"
 CAM_COMBINE_SEND = "async_combine_send"
 CAM_COMBINE_RECV = "async_combine_recv"
+CAM_CUST_OPAPI_ENV = "CAM_CUST_OPAPI_LIB_PATH"
+CAM_CUST_OPAPI_DEFAULT = Path(
+    "/usr/local/Ascend/cann-9.0.1/opp/vendors/CAM/op_api/lib/libcust_opapi.so",
+)
 
 
 def get_afd_cann_vendor_path() -> Path:
@@ -68,6 +73,25 @@ def _assert_cam_namespace_registered(torch: object) -> None:
     )
 
 
+def _preload_cam_cust_opapi() -> None:
+    """Make CAM's custom op-api symbols visible to the Python extension.
+
+    ``umdk_cam_op_lib`` resolves the aclnn CAM entry points by name while its
+    operators are first invoked.  CANN's stock ``libopapi.so`` does not export
+    those symbols; they live in CAM's ``libcust_opapi.so``.  Load that library
+    globally before importing the extension so the real CAM operators bind to
+    the vendor implementation instead of the stock op-api library.
+    """
+
+    library = Path(os.environ.get(CAM_CUST_OPAPI_ENV, CAM_CUST_OPAPI_DEFAULT))
+    if not library.is_file():
+        raise RuntimeError(
+            "CAMAsyncAFDConnector requires CAM libcust_opapi.so; expected "
+            f"{library}. Install the CAM operator package first.",
+        )
+    ctypes.CDLL(str(library), mode=ctypes.RTLD_GLOBAL)
+
+
 @lru_cache(maxsize=1)
 def ensure_cam_p2p_ops_available() -> None:
     """Import the custom operators used by ``CAMP2pAFDConnector``.
@@ -107,6 +131,7 @@ def ensure_cam_async_ops_available() -> None:
     """Ensure the runtime exposes the real CAM async operator namespace."""
 
     try:
+        _preload_cam_cust_opapi()
         import torch
         import torch_npu  # noqa: F401
         import umdk_cam_op_lib  # noqa: F401
--- a/afd_plugin/connectors/npu/async_cam.py
+++ b/afd_plugin/connectors/npu/async_cam.py
@@ -195,6 +195,8 @@ class AFDAsyncTopology:
 
     role: str
     role_rank: int
+    dp_group_index: int
+    num_dp_groups: int
     world_rank: int
     attn_size: int
     ffn_size: int
@@ -256,11 +258,13 @@ class CAMAsyncAFDConnector(AFDConnectorBase):
             afd_config,
             role_rank,
             num_routed_experts=self.num_routed_experts,
+            attn_ranks_per_dp=self.tp_size,
         )
         self.world_rank = self.topology.world_rank
         self.attn_size = self.topology.attn_size
         self.ffn_size = self.topology.ffn_size
         self.expert_per_rank = self.topology.expert_per_rank
+        self._process_group_name = _cam_process_group_name(self.topology)
         self.comm_args: Tensor | None = None
         self._placeholder: Tensor | None = None
         self._pending_attention_payloads: dict[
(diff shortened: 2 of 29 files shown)


afd_plugin/compat/npu/ops.py
```diff
@@ -16,6 +17,10 @@
 CAM_DISPATCH_RECV = "async_dispatch_recv"
 CAM_COMBINE_SEND = "async_combine_send"
 CAM_COMBINE_RECV = "async_combine_recv"
+CAM_CUST_OPAPI_ENV = "CAM_CUST_OPAPI_LIB_PATH"
+CAM_CUST_OPAPI_DEFAULT = Path(
+    "/usr/local/Ascend/cann-9.0.1/opp/vendors/CAM/op_api/lib/libcust_opapi.so",
```
ShwStone, line 22: I guess this is a bug of the CAM operator and maybe this workaround is not permanent

afd_plugin/connectors/npu/async_cam.py
```diff
@@ -349,7 +353,23 @@ def recv_ffn_work_item(
...
+        # A zero in the header's leading count is valid for an FFN rank that
+        # received no routed tokens in this dispatch.  Such ranks must still
+        # enter combine-send (the zero-token fallback below supplies its
+        # placeholder) so every participant completes the CAM collective.
+        # CAM 209 currently returns zero in the documented layer-index slot
+        # for every dispatch-recv result.  The connector-driven FFN runner
+        # already has the authoritative decoder-layer order, so consuming this
+        # stale field makes every DSV4 work item execute FFN layer 0.  CAM
+        # combine-send also uses this header to match the completion to its
+        # original dispatch, so merely fixing the local model call is not
```
ShwStone, line 532: This model-specific workaround appears to break the existing DeepSeek V2 two-stage async MoE ubatching path.

`_ffn_forward_connector_driven()` still receives only `num_moe_layers` work items per invocation. With two ubatches, however, Attention sends two work items for each MoE layer in this order:

```text
L0/S0, L0/S1, L1/S0, L1/S1,
L2/S0, L2/S1, L3/S0, L3/S1
```

The outer FFN worker loop repeatedly invokes `_ffn_forward_connector_driven()`, so for four MoE layers the first invocation consumes the first four messages and the second invocation consumes the remaining four. Before this change, the loop variable was intentionally unused and the actual layer was taken from the received CAM metadata.

With `expected_layer_idx` restarted on every invocation, the same messages are now interpreted as:

```text
received    forced layer
L0/S0       L0
L0/S1       L1  # wrong
L1/S0       L2  # wrong
L1/S1       L3  # wrong

L2/S0       L0  # wrong after the next invocation restarts
L2/S1       L1  # wrong
L3/S0       L2  # wrong
L3/S1       L3
```

This also rewrites `token_nums_rankid_layeridx[2]`, which is later passed back to `async_combine_send`, so the mismatch affects both the local FFN model selection and CAM completion matching.

I understand that CAM 209 currently reports an unusable layer field for the DSV4 baseline, but that workaround should not be applied to the shared V2 ubatch path. Could we either scope it to DSV4 without ubatching, or make the FFN receive schedule explicitly track `(layer_idx, stage_idx)`? Please also add a regression test using the two-stage message order above.

afd_plugin/v1/worker/npu/attention_model_runner.py
```diff
@@ -1496,22 +1485,33 @@ def _send_dp_metadata(
...
+                    .to(device="cpu", dtype=torch.int64)
+                    .flatten()
+                    .tolist()
+                )
         is_warmup = bool(self._is_warmup)
         is_graph_capturing = bool(self._afd_is_graph_capturing)
         payload = AFDControlPayload(
             dp_metadata_list=dp_metadata_list,
             is_graph_capturing=is_graph_capturing,
             is_warmup=is_warmup,
```
ShwStone, line 1524: This adds a device-to-host synchronization, Python list materialization, and JSON serialization to every control-plane forward, including DeepSeek V2 paths that never consume `input_ids` on the FFN side. That is a significant hot-path cost, particularly for large prefill batches, and may erase part of the disaggregation benefit we are trying to measure.

Could we avoid making token-ID transport part of the generic DP metadata protocol?

Some possible designs, in decreasing order of preference:

1. For Async CAM with `compute_gate_on_attention=true`, keep hash routing on Attention and do not transport input IDs at all.
2. Introduce an explicit connector/model capability such as `requires_ffn_input_ids`, so V2 and non-hash DSV4 paths pay no additional cost.
3. If CAMP2P DSV4 must route on FFN, send one flat `int32` tensor plus stage offsets through an NPU/HCCL data path instead of converting it to CPU lists and JSON.
4. If a device data path is not currently possible, at minimum restrict the existing CPU serialization to DSV4 hash-routing steps and avoid duplicating IDs across stages or TP ranks.

Please include a prefill/decode latency comparison with and without this transport. The current implementation changes the performance characteristics of existing V2 workloads even though the data is only needed by a new DSV4 routing mode.

afd_plugin/v1/worker/npu/ffn_model_runner.py
```diff
@@ -330,6 +338,13 @@ def _ffn_forward_connector_driven(
                     work_item,
                     rank_ffn_output,
                 )
+                # ``async_combine_send`` returns after enqueueing its NPU work.
+                # Without an acknowledgement/credit path, immediately calling
+                # dispatch-recv for all following MoE layers can reuse CAM's
+                # bounded buffers while the prior combine is still in flight.
+                # The target baseline is deliberately serialized per layer;
+                # async MoE ubatching will introduce explicit stage credits.
+                torch.npu.synchronize()
```
ShwStone, line 347: Could you document and provide evidence for why a device-wide synchronization is required here?

This PR adds `torch.npu.synchronize()` after:

- `async_dispatch_send`;
- `async_combine_recv`;
- every FFN-side `async_combine_send`.

Only the FFN-side point mentions bounded CAM buffer reuse. The other two have no explanation, and normal same-stream custom-op ordering should already prevent consumers from observing an unfinished output unless CAM launches work on an external stream or has a non-standard buffer-lifetime contract.

For each synchronization point, could you clarify:

- which CAM operation continues asynchronously after returning;
- which input/output or internal buffer may be reused too early;
- why current-stream ordering is insufficient;
- the affected CAM/CANN version;
- the concrete failure observed after removing the synchronization;
- why an event, stream wait, completion handle, or explicit credit mechanism cannot be used instead.

The FFN worker already synchronizes after each connector-driven step. The new FFN synchronization is additionally performed after every layer, so it serializes the complete device rather than only protecting the relevant CAM buffer. This is especially concerning for the existing V2 overlap/ubatch path.

Please provide correctness reproduction and profiler results with and without each synchronization. If these are temporary CAM 209 workarounds needed only by the serialized DSV4 baseline, they should be scoped accordingly instead of changing the shared V2 connector behavior.

jiangkuaixue123 (commented):
Thanks for working on this. I have two requests regarding the scope and code organization:

1. Please split the decode-stage P2P adaptation out of this PR. It should be implemented and reviewed independently from the CAM async connector support so that each change has a clear scope and validation boundary.

2. Please separate the NPU DeepSeek V4 implementation from the existing GPU DeepSeek V4 model implementation. Before proceeding, please discuss the directory structure with Li Runxin and agree on a layout that keeps the backend-specific implementations clearly isolated.

hsliuustc0106:
fix conflicts and update PR desc

PR #260, opened 2026-08-20
fix(npu): unblock async FFN teardown

### LINKED ISSUE #251: NPU async E2E teardown always fails pytest
## Problem

NPU async E2E scenarios can launch and complete the evaluation workload, but pytest still fails during teardown. The current process cleanup path cannot terminate the async NPU workers cleanly and reports the forced termination as a cleanup failure.

## Expected behavior

After evaluation completes, the runner should terminate all Attention and FFN processes without turning a successful E2E run into a pytest failure.

ShwStone:
## Summary

- add a full one-token CAM dispatch/combine shutdown handshake, using an out-of-range layer index as an unambiguous wire marker
- keep async FFN ranks receiving until they acknowledge the shutdown sentinel, then close HCCL only after the daemon exits
- preserve close-before-join for control-plane connectors and defer unsafe cleanup when the async handshake cannot complete
- add focused connector, model-runner, and worker lifecycle regression coverage

## Testing

- [x] NPU async E2E: `afd-eager-async-cam`
- [x] NPU async E2E: `afd-async-ubatch`
- [x] CPU-only CI: lint
- [x] DCO

Closes #251

### DIFF
--- a/afd_plugin/connectors/npu/async_cam.py
+++ b/afd_plugin/connectors/npu/async_cam.py
@@ -247,6 +247,10 @@ class CAMAsyncAFDConnector(AFDConnectorBase):
         self.hidden_size = hf_config.hidden_size
         self.topk = hf_config.num_experts_per_tok
         self.num_routed_experts = hf_config.n_routed_experts
+        # Normal layer indices occupy [0, num_hidden_layers). Using the first
+        # out-of-range value gives shutdown an unambiguous wire marker without
+        # depending on a model having a dense layer 0.
+        self.shutdown_sentinel_layer_idx = int(hf_config.num_hidden_layers)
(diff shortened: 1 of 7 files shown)


jiangkuaixue123 (commented):
Overall, this PR introduces a significant amount of shutdown-specific logic into the runtime, including a custom sentinel protocol, additional state handling, and different cleanup paths for different connector types. The resulting lifecycle is fairly complex and may be difficult for future maintainers to understand and reason about.

Could we consider addressing this in the test runner or E2E test cleanup instead—for example, by tolerating or handling forced termination during teardown—rather than adding a new shutdown protocol to the production runtime? That may provide a simpler and more maintainable solution to the original test failure.

ShwStone:
Superseded by #279.

PR #264, opened 2026-08-24
[NPU]Add DSV4 NPU ubatch async connector

bjf-frz:
PLEASE FILL IN THE PR DESCRIPTION AND MAKE SURE THE CHECKLIST ITEMS HAVE BEEN CONSIDERED.

## Issue

- Related issue(s): #

## Scope

- In scope:
- Out of scope:

## Docs Impact

- Files updated:
- If none, reason:

---

<details>

- [ ] Purpose is clear and linked to public context when possible.
- [ ] Scope is bounded.
- [ ] Compatibility with vLLM v0.26.0 is considered.
- [ ] No changes are made to the vLLM source checkout.
- [ ] Plugin-owned classes or explicit dotted class paths are preferred over monkey patches.
- [ ] Any compat shim or monkey patch is isolated, idempotent, version-guarded, documented, and tested.
- [ ] Imports remain CPU-safe; CUDA-heavy work is delayed or GPU-gated.
- [ ] Validation evidence is included, including skipped GPU tests when applicable.

</details>


afd_plugin/__init__.py
```diff
@@ -45,6 +47,51 @@ def __getattr__(name: str):
 _logger = logging.getLogger(__name__)
 _registered = False
 
+
```
jiangkuaixue123, line 50: Why do we need this process-wide multiprocessing override here? It does not appear necessary for the AFD plugin itself, and mutating Python's global multiprocessing context (including the private `_concrete_contexts` registry) is quite intrusive. Please remove this unless there is a concrete AFD-specific requirement and reproducer that cannot be addressed in the affected NPU runtime path.

afd_plugin/v1/worker/npu/npu_ubatch_wrapper.py
```diff
@@ -157,7 +157,15 @@ def __init__(
         self.vllm_config = vllm_config
         self.compilation_config = vllm_config.compilation_config
         self.comm_stream = torch.npu.Stream(device=device)
-        assert self.vllm_config.parallel_config.num_ubatches == AFD_NPU_NUM_UBATCHES
+        # vLLM native ubatching/DBO is rejected for CAMAsyncAFDConnector by
+        # fail_if_unsupported_npu_afd_features; the AFD async MoE ubatching
+        # path drives stage counts through connector_extra_config
+        # (async_moe_num_ubatches), which is validated there as well.
+        if self.vllm_config.parallel_config.use_ubatching:
```
jiangkuaixue123: This check does not look correct. Rejecting `use_ubatching` in the wrapper disables native ubatching for ordinary decode as well, rather than only restricting the CAM async path. Please remove this change so normal decode ubatching can still be enabled.

afd_plugin/model_executor/models/deepseek_v4.py
```diff
@@ -1,510 +1,16 @@
...
-        return super().load_weights(
-            _iter_role_weights(weights, role=self.afd_role),
-        )
-
+"""Platform-selected AFD wrapper for DeepSeek-V4."""
+
+from vllm.platforms import current_platform
+
+if current_platform.device_type == "npu":
+    from afd_plugin.model_executor.models.deepseek_v4_npu import (
```
jiangkuaixue123: Could we place the GPU and NPU implementations under dedicated `gpu/` and `npu/` directories, respectively, and keep the two backend implementations isolated? Please avoid sharing/reusing backend-specific DeepSeek-V4 code across GPU and NPU.

afd_plugin/compat/patches/npu/deepseek_v4_kv_cache.py
```diff
@@ -0,0 +1,164 @@
+# SPDX-License-Identifier: Apache-2.0
```
jiangkuaixue123: I do not understand why this KV-cache patch is necessary. What concrete upstream issue does it address, and why must it be patched in AFD rather than fixed in vLLM-Ascend? Please provide a minimal reproducer and explain the necessity of this patch; otherwise, please remove it.

afd_plugin/compat/patches/npu/modelslim_dsv4.py
```diff
@@ -0,0 +1,119 @@
+# SPDX-License-Identifier: Apache-2.0
```
jiangkuaixue123: I do not understand why this ModelSlim patch is necessary. What exact incompatibility requires AFD to override the upstream prefix mapping, and why cannot it be handled in vLLM-Ascend? Please provide a concrete reproducer and justify the patch; otherwise, please remove it.

jiangkuaixue123 (commented):
Please address the inline comments.

jiangkuaixue123:
Please fix conflict.

ShwStone:
It seems that there are a lot of documents and operation manuals in the repository about specific test environments. Please make sure that the documents you submit are general guidance and avoid submitting irrelevant documents and startup scripts.

afd_plugin/model_executor/models/npu/deepseek_v2_async_cam_forward.py
```diff
@@ -168,6 +168,13 @@ def run_attention_gate_afd_forward(
             llama_4_scaling,
         )
 
+        # CAM allocates persistent operator workspace on its first dispatch.
+        # Keep EngineCore's synthetic KV-cache profile local so that workspace
+        # is not mistaken for reclaimable KV capacity.  The first real request
+        # performs the normal CAM dispatch/receive sequence below.
+        if forward_context.in_profile_run:
```
ShwStone: Could you explain why the startup profile must skip CAM dispatch/combine here?

PR #248 intentionally removed an equivalent guard to fix #247. Matched profile traffic ensures that every Attention DP rank enters the CAM operators before live serving; skipping it appears to defer lazy CAM/HCCL initialization to the first real request, potentially reintroducing the DP>1 first-request deadlock.

Has the CAM 209 initialization contract changed? If so, how does this PR initialize all Attention ranks before serving, and has the DP2 single-cold-request scenario from #247 been validated?

This also appears to conflict with the existing `test_async_cam_profile_forward_runs_matched_connector_io` regression test added by #248.

ShwStone:
Could you clarify the motivation for replacing the previous global-DP topology with multiple independent DP1 Attention services?

The previous setup already supported a cross-node DP3TP8 + EP8 topology (for example, local DP2 on one node and DP1 on another), while vLLM handled request routing across the Attention DP replicas automatically. With this change, request routing must be moved to an external load balancer, and the FFN side additionally needs multiple CAM/HCCL communicators plus a TCPStore scheduler.

What concrete limitation or failure in the previous global CAM world requires this higher-complexity design? Is this mandated by a DSV4/CAM 209 contract, or does the old topology exhibit a reproducible correctness or performance issue under concurrent DP traffic?

It would also be helpful to explain why the existing cross-node DP mechanism cannot satisfy the intended deployment and what benefit justifies giving up vLLM built-in DP routing.

.buildkite/cuda/test-merge.yml
```diff
@@ -12,10 +12,18 @@ steps:
   - group: ":card_index_dividers: E2E Test"
     depends_on: upload-merge-pipeline
     steps:
-      - label: "E2E · DeepSeekV2-Lite"
+      - label: "E2E · DeepSeekV2-Lite · Legacy"
```
jiangkuaixue123: This file change is not needed.

.buildkite/cuda/test-ready.yml
```diff
@@ -12,10 +12,18 @@ steps:
   - group: ":card_index_dividers: E2E Test"
     depends_on: upload-ready-pipeline
     steps:
-      - label: "E2E · DeepSeekV2-Lite"
+      - label: "E2E · DeepSeekV2-Lite · Legacy"
```
jiangkuaixue123: This file change is not needed.

afd_plugin/compat/patches/engine_core.py
```diff
@@ -42,6 +42,14 @@ def __init__(
     executor_fail_callback: Callable | None = None,
     include_finished_set: bool = False,
 ):
+    # Install DSV4's Ascend KV grouping compatibility before an EngineCore can
+    # ask vLLM to construct its cache configuration.
+    from afd_plugin.compat.patches.npu.deepseek_v4_kv_cache import (
```
jiangkuaixue123: Remove these two patches if there is no need.

afd_plugin/model_executor/models/npu/deepseek_v2_async_cam_forward.py
```diff
@@ -168,6 +168,13 @@ def run_attention_gate_afd_forward(
             llama_4_scaling,
         )
 
+        # CAM allocates persistent operator workspace on its first dispatch.
```
jiangkuaixue123: Please remove this.

afd_plugin/model_executor/models/deepseek_v2.py
```diff
@@ -128,6 +128,14 @@ def _send_and_receive(
         if afd_metadata is None:
             raise RuntimeError("RemoteFFNProxy requires AFD forward metadata")
         forward_context = get_forward_context()
+        # vLLM's startup memory profile executes a synthetic, role-local
+        # forward.  There is no matching FFN request stream at that point, so
+        # issuing CAM dispatch/combine from the Attention profile can pair its
+        # small dummy batch with an FFN receive-capacity buffer and leave the
+        # external communicator in a broken state.  Keep the profile local;
+        # live forwards below retain the real remote-FFN exchange.
+        if bool(getattr(forward_context, "in_profile_run", False)):
```
jiangkuaixue123: There is no need. Please remove.

afd_plugin/v1/worker/npu/ffn_model_runner.py
```diff
@@ -330,6 +338,13 @@ def _ffn_forward_connector_driven(
                     work_item,
                     rank_ffn_output,
                 )
+                # ``async_combine_send`` returns after enqueueing its NPU work.
+                # Without an acknowledgement/credit path, immediately calling
+                # dispatch-recv for all following MoE layers can reuse CAM's
+                # bounded buffers while the prior combine is still in flight.
+                # The target baseline is deliberately serialized per layer;
+                # async MoE ubatching will introduce explicit stage credits.
```
jiangkuaixue123, line 345: Need remove.

afd_plugin/v1/worker/npu/attention_model_runner.py
```diff
@@ -1555,6 +1571,11 @@ def load_model(self) -> None:
         if not self.connector.is_initialized:
             self.connector.init_afd_connector()
 
+    def _is_dsv4_model(self) -> bool:
```
jiangkuaixue123: 这个函数多余

afd_plugin/connectors/npu/async_cam.py
```diff
@@ -820,43 +747,13 @@ def _require_async_transfer_state(
     return states
 
 
-_CAM_LOG_SKIPPED_ARGS = frozenset({"comm_args", "comm_id", "group_name"})
-_CAM_OP_IO_LOG_ENV = "AFD_CAM_OP_IO_LOG"
-
```
ShwStone, line 825: _log_cam_op_values ​​is useful for debugging, especially for async cases. I recommend reconsidering removing this function

ShwStone: However, the current implementation does produce excessive output pollution logs, and perhaps a better implementation could be considered.

ShwStone:
I think the overall direction should be reconsidered. PR #203 implements ubatching as an explicit model-side, two-stage CAM pipeline with well-defined `send/recv` ordering and backpressure. PR #264 instead passes the DSV4 stages to `AscendUBatchWrapper`, which runs two full model forwards concurrently. These execution contracts are not equivalent.

The current approach has two main concerns:

- It lacks explicit stage credits and relies on `torch.npu.synchronize()` after each FFN work item to protect CAM buffers, which substantially reduces the intended overlap.
- The concurrent full forwards share mutable DSV4 state, including `_mtp_hidden_buffer` and potentially IndexCache buffers, without a demonstrated concurrency-safety contract.

I suggest keeping the existing stage planning, attention metadata, and CAM layout logic, but implementing DSV4 ubatching as a model-side pipeline similar to PR #203. Each stage should maintain its own mHC state, Attention/CAM/FFN execution should be explicitly scheduled, and the stages should be merged before updating the MTP buffer and running `hc_head` and the final norm. `AscendUBatchWrapper` should remain limited to native DBO rather than serving as the executor for Async CAM MoE ubatching.

afd_plugin/distributed/afd_process_group.py
```diff
@@ -42,12 +42,13 @@ def __exit__(self, exc_type: object, exc_value: object, tb: object) -> None:
 def init_afd_process_group(
     *,
     backend: str,
-    init_method: str,
+    init_method: str | None = None,
```
jiangkuaixue123, line 45: This file does not need to change.

afd_plugin/model_executor/models/model_utils.py
```diff
@@ -19,6 +19,8 @@ def get_afd_model_config(
     device_type: Literal["cuda", "npu"],
 ) -> ModelConfig:
     """Return a model config that resolves to an AFD model implementation."""
+    from afd_plugin import register_afd
+    register_afd()
```
jiangkuaixue123, line 23: This change is no need.

afd_plugin/v1/worker/npu/attention_model_runner.py
```diff
@@ -1438,10 +1426,19 @@ def _install_afd_metadata_on_forward_context(
             return
         dp_metadata = forward_context.dp_metadata
         ubatch_slices = forward_context.ubatch_slices
+        if self._afd_async_moe_ubatch_metadata is not None:
```
jiangkuaixue123: Why this is need?

afd_plugin/v1/worker/npu/npu_ubatch_wrapper.py
```diff
@@ -153,11 +153,11 @@ def __init__(
         enable_enpu: bool = False,
     ):
         assert not enable_enpu, "AscendUBatchWrapper does not support ENPU"
+        assert vllm_config.parallel_config.num_ubatches == 2
```
jiangkuaixue123: This change seems useless.

afd_plugin/v1/worker/npu/attention_model_runner.py
```diff
@@ -428,6 +413,10 @@ def _build_attention_metadata_with_async_moe_ubatches(
             use_sequence_parallel=use_sequence_parallel,
             parent_input_tokens=num_tokens_padded,
         )
+        self._afd_pending_metadata = self._build_afd_metadata(
```
ShwStone: This second assignment looks redundant for the model-owned Async CAM ubatching path.

`_afd_pending_metadata` has already been created as full-batch metadata at L362-L365. The physical stage partition is then fully represented by `AsyncMoeUbatchMetadata` above, while each CAM operation carries its actual `stage_idx`, `layer_idx`, and `seq_len` through `AFDTransferMetadata`.

Expanding `_afd_pending_metadata` again here introduces two competing representations of the same stage layout. It also allocates a second transaction ID and leaves both stage-local `ForwardContext`s sharing an `AFDForwardContextMetadata` whose `stage_idx` remains 0, even when `forward_context.ubatch_idx == 1`.

Could we remove this assignment and keep `_afd_pending_metadata` as the full-batch/request-level metadata created above? The stage slices should remain owned by `AsyncMoeUbatchMetadata`.

The related reconstruction at L1429-L1433 also appears unnecessary for CAMAsync because its connector has no control plane. If model-owned ubatching must support a control-plane connector later, that path should explicitly derive its control payload from `AsyncMoeUbatchMetadata`, without changing or sharing the parent `AFDForwardContextMetadata`.

jiangkuaixue123: I agree with it.

afd_plugin/model_executor/models/npu/deepseek_v4.py
```diff
@@ -0,0 +1,853 @@
...
+        hidden_states: torch.Tensor,
+        layer_idx: int,
+        router_logits: torch.Tensor,
+    ) -> torch.Tensor:
+        raise NotImplementedError(
+            "DSV4 AFD does not support compute_gate_on_attention yet"
+        )
+
+
+def _pad_stage_input_ids(
```
jiangkuaixue123, line 509: Why does async need pad input_ids

afd_plugin/connectors/npu/async_cam.py
```diff
@@ -349,7 +349,24 @@ def recv_ffn_work_item(
...
+        # already has the authoritative decoder-layer order, so consuming this
+        # stale field makes every DSV4 work item execute FFN layer 0.  CAM
+        # combine-send also uses this header to match the completion to its
+        # original dispatch, so merely fixing the local model call is not
+        # sufficient: all later layers would still be combined as layer 0.
+        # CAM returns the true decoder-layer index in the dispatch-recv header
+        # (field 2 of token_nums_rankid_layeridx).  Use it for both the local
+        # FFN layer computation and the combine-send header so the completion
+        # matches the attention-side combine-recv.
+        layer_idx = max(0, int(token_nums_rankid_layeridx[2].item()))
```
jiangkuaixue123: This change should be removed. The previous code already reads the decoder layer from `token_nums_rankid_layeridx[2]`; if CAM 209 actually returns `0` in that slot, `max(0, ...)` still produces layer 0 and does not fix the stated issue. The cloned header is never modified, so it cannot correct the metadata passed to `async_combine_send`. For a negative value, the local layer would be clamped to 0 while the cloned header would still carry the negative value, creating exactly the local/combine mismatch the comment says must be avoided. The connector-driven runner also currently discards its loop layer index (`for _ in _ffn_layer_indices(self)`), so the claim that it already has and uses an authoritative decoder-layer order is not reflected in the implementation. The updated tests only replace the fake list with a Tensor so that `clone()` can run; their positive layer values already passed with the old implementation. Please delete this added block and retain the original `layer_idx = int(token_nums_rankid_layeridx[2].item())`. If there is a real CAM header bug, it should be supported by the CAM field contract plus a reproducer, and the eventual fix must update both the local layer selection and the header consumed by combine-send consistently.

PR #275, opened 2026-08-27
[Feat]: Add NPU ModelRunnerV2 DBO with Eager and FULL_DECODE_ONLY ACL Graph Support

### LINKED ISSUE #265: [Feature]: Add DBO support for Ascend NPU ModelRunnerV2
### Feature, motivation, and pitch

Ascend NPU ModelRunnerV2 support landed in #257, but the current path explicitly rejects `enable_dbo=true` and `use_ubatching=true`. This leaves DBO workloads on ModelRunnerV1 and prevents the Ascend ModelRunnerV2 path from reaching the DBO milestone described in Phase 3 of #240.

Add native DBO support to `AFDNPUAttentionModelRunnerV2` for the synchronous `CAMP2pAFDConnector` path. The implementation should preserve native vLLM/vLLM-Ascend ownership of ModelRunnerV2 request state, input preparation, per-uBatch Attention metadata, output merging, NPU streams/events, and ACL graph lifecycle.

Related work:

- #240 — Ascend NPU ModelRunnerV2 RFC, Phase 3
- #257 — Ascend ModelRunnerV2 implementation without DBO

### Proposed change

- Enable `enable_dbo` / native ubatching for the Ascend ModelRunnerV2 Attention path.
- Prefer an `AscendUBatchRunner` owned by vLLM-Ascend, exposed through a narrow factory or extension hook.
- Integrate AFD transaction, stage, token-shape, DP, and connector-control metadata per uBatch without copying the ModelRunnerV2 execution loop.
- Initially support exactly two uBatches with synchronous `CAMP2pAFDConnector`.
- Deliver eager DBO first; qualify `FULL_DECODE_ONLY` DBO only after eager correctness is established.
- Keep ModelRunnerV1 DBO and ModelRunnerV2 no-DBO behavior unchanged.
- Fail closed for unsupported connectors, graph modes, topologies, or uBatch counts.

### vLLM compatibility and extension points

Preferred extension point:

- A vLLM-Ascend-owned `AscendUBatchRunner` and an overridable uBatch-runner construction hook in the native NPU ModelRunnerV2 path.
- Narrow AFD composition/hooks for connector lifecycle, control publication, AFD forward-context metadata, profiling, and cleanup.

Compat shim needed:

- A version-pinned shim may be used if the pinned vLLM-Ascend runtime does not yet expose the required factory or lifecycle hooks.

Monkey patch needed:

- Avoid if possible. If unavoidable, keep it exact-version, minimal, and documented with an upstream/removal plan.
- Do not copy native `execute_model`, input preparation, Attention metadata construction, output merging, or ACL graph-manager lifecycle into afd-plugin.

The implementation targets the repository's pinned vLLM/vLLM-Ascend runtime pair and must not modify either upstream source tree from this repository.

### Alternatives considered

- Keep DBO on ModelRunnerV1 only: lowest implementation risk, but leaves Ascend MRV2 feature parity incomplete.
- Port the existing NPU V1 DBO loop into ModelRunnerV2: rejected because it duplicates upstream execution ownership and creates a large upgrade surface.
- Copy the GPU `UBatchRunner` into afd-plugin: rejected because CUDA-specific stream/graph assumptions do not define the Ascend runtime contract.
- Implement eager and ACL graph DBO together: increases the debugging surface; eager-first provides a smaller correctness gate.

### Validation plan

Focused tests:

- Configuration accepts the supported two-uBatch MRV2 DBO setup and rejects unsupported combinations before connector rendezvous or weight loading.
- Per-uBatch AFD metadata preserves real and padded token lengths, stage identity, transaction identity, and DP rank agreement.
- Connector control is published before each uBatch data transfer and outside ACL graph bodies.
- Failure paths clean up sibling execution, streams/events, forward contexts, graph state, and connector transactions.

Ascend hardware E2E:

- DeepSeek-V2-Lite, synchronous `CAMP2pAFDConnector`, 1A1F, DP1/TP1, eager DBO.
- Decode, prefill, mixed batches, uneven splits, padding-only tails, repeated requests, profiling, and clean shutdown.
- Deterministic output comparison against approved native/no-DBO or ModelRunnerV1 references.
- DP2/TP1 and TP2/DP1 coverage before broad topology claims.
- Full non-accuracy NPU E2E gate, then GSM8K-7 accuracy.
- After eager acceptance, `FULL_DECODE_ONLY` capture/replay, graph-key separation by uBatch count/shape, graph misses, and shape changes.

Acceptance criteria:

- Exactly two uBatches execute correctly through native Ascend ModelRunnerV2 DBO ownership.
- No copy of the native MRV2 execution loop or Ascend Attention/graph internals is introduced.
- AFD connector payloads and control metadata are correct for every uBatch.
- Eager correctness, cleanup, and accuracy gates pass on Ascend hardware.
- Graph DBO is not claimed until its separate capture/replay gate passes.
- Existing Ascend MRV2 no-DBO and ModelRunnerV1 DBO tests remain green.

### Before submitting

- [x] I identified whether this belongs in plugin-owned code, compat helpers, or compat patches.

lirx-pd:
## Purpose

This PR adds temporary Dual Batch Overlap (DBO) support to the AFD NPU
ModelRunnerV2 path. The pinned vLLM `v0.26.0` ABI does not yet include the
ModelRunnerV2 DBO implementation it needs.

It adapts ModelRunnerV2 DBO behavior from `specture724/vllm` branch
`feat/v2/dbo-fullcg` at commit `626fee7831` to vLLM ABI `568afb3a13`. The
missing upstream pieces live in an isolated compatibility backport. When the
pinned vLLM release gains native ModelRunnerV2 DBO support, this backport can
be removed.

The AFD NPU ModelRunnerV2 path supports:

- eager execution without DBO;
- eager execution with two DBO microbatches;
- ACL graph execution without DBO using `FULL` or `FULL_DECODE_ONLY`;
- DBO ACL graph execution using `FULL_DECODE_ONLY`.

The change does not modify the vLLM or vLLM-Ascend source trees. NPU
ModelRunnerV1 keeps its existing execution path.

## Architecture diagram

```mermaid
flowchart TB
    Config["AFD NPU V2 configuration"]
    Validation["Scoped v0.26 validation shim"]
    Runner["AFDNPUAttentionModelRunnerV2"]
    Dispatch["DP synchronization and DBO dispatch"]
    Single["Native single-batch V2 path"]
    Split["Two request-boundary uBatches"]
    Eager["AFDAscendUBatchRunnerV2<br/>eager execution"]
    Graph["AFDModelAclGraphManagerV2<br/>DBO graph capture/replay"]
    Yield["Per-layer DBO handoff"]
    Connector["CAMP2pAFDConnector"]
    FFN["Existing AFD NPU FFN runner"]
    Output["Merged model output"]
    Guard["Upstream capability guard<br/>remove backport when native DBO lands"]

    Config --> Validation --> Runner
    Runner --> Dispatch
    Dispatch -->|threshold not met| Single --> Connector
    Dispatch -->|eager DBO| Split --> Eager --> Yield
    Dispatch -->|captured DBO shape| Split --> Graph --> Yield
    Yield --> Connector --> FFN --> Connector --> Output
    Guard -.-> Validation

    classDef new fill:#fff7ed,stroke:#f59e0b,color:#92400e,stroke-width:2px;
    classDef reused fill:#ecfdf5,stroke:#16a34a,color:#166534,stroke-width:2px;
    classDef upstream fill:#eff6ff,stroke:#3b82f6,color:#1e40af,stroke-width:2px;

    class Validation,Dispatch,Split,Eager,Graph,Guard new;
    class Runner,Yield,Connector,FFN,Output reused;
    class Single upstream;
```

## Issue

- Related issue(s): #265 
- Closing keyword, only if fully resolved: #265 

### In scope

- Backport the ModelRunnerV2 DBO execution pieces missing from the pinned vLLM
  `v0.26.0` ABI into `afd-plugin`.
- Enable two-microbatch eager DBO for the AFD NPU ModelRunnerV2 Attention path.
- Enable two-microbatch `FULL_DECODE_ONLY` ACL graph capture and replay.
- Synchronize token counts, graph mode, uniform decode state, and DBO selection
  across DP ranks before execution.
- Apply the configured decode/prefill DBO thresholds using the minimum real
  token count across DP ranks.
- Pad DBO execution to the maximum selected DP shape while falling back to a
  single batch when graph padding would leave the second microbatch empty.
- Split ModelRunnerV2 inputs at request boundaries and preserve Ascend-specific
  input, attention, block-table, slot-mapping, and forward-context state for
  each microbatch.
- Run the two eager microbatches through isolated forward contexts and merge
  tensor, auxiliary-hidden-state, and intermediate outputs in request order.
- Capture plugin-owned twin DBO graphs separately from native single-batch ACL
  graphs and publish matching AFD metadata for warmup, capture, and replay.
- Reuse the upstream FIA workspace while maintaining per-microbatch MLA graph
  parameters and updating the captured task-group handles during replay.
- Allow the existing connector-driven FFN runner to replay the matching graph
  during repeated Attention-side capture events.
- Narrowly bypass vLLM's native DBO/all-to-all validation only for an active AFD
  NPU ModelRunnerV2 configuration.
- Preserve native execution when DBO is disabled or the runtime threshold is
  not met.
- Preserve the existing NPU ModelRunnerV1 DBO and graph paths.

### Out of scope

- Native or generally reusable vLLM ModelRunnerV2 DBO support.
- More than two microbatches.
- DBO with DP size one.
- DBO ACL graph modes other than `FULL_DECODE_ONLY`.
- Speculative decoding, LoRA, multimodal models, or encoder-decoder models with
  NPU ModelRunnerV2 DBO.
- PP, PCP, or DCP greater than one, elastic expert parallelism, EPLB, sequence
  parallel MoE, or other topologies already unsupported by AFD NPU
  ModelRunnerV2.
- Asynchronous CAMP or asynchronous AFD connector lifecycles.
- Performance tuning or a claim that DBO improves every model or workload.

### Temporary vLLM v0.26 ABI backport

`afd_plugin/compat/backports/vllm_v026_mrv2_dbo` contains the copied and
adapted behavior missing from the pinned upstream ABI:

- an AFD batch descriptor carrying `num_ubatches`;
- DP synchronization and DBO/graph dispatch;
- request-boundary microbatch slicing;
- ModelRunnerV2 input and model-input slicing;
- per-microbatch attention metadata preparation;
- output merging for tensors and Ascend auxiliary output structures;
- the temporary ModelRunnerV2 execute path.

At startup, the backport checks the upstream `BatchExecutionDescriptor` shape.
If upstream already provides `num_ubatches`, initialization fails with an
instruction to remove the compatibility layer rather than shadowing the native
implementation.

### Eager DBO execution

`AFDNPUAttentionModelRunnerV2.execute_model()` uses the backported execute path
only when vLLM reports `use_ubatching`; otherwise it uses the native
vLLM-Ascend runner.

For a selected DBO batch, it:

1. prepares the native ModelRunnerV2 input batch and model state;
2. splits the padded batch into exactly two request-boundary slices;
3. builds independent Ascend attention metadata, slot mappings, DP metadata,
   and forward contexts for both stages;
4. runs both stages under the existing DBO handoff used by the AFD model proxy;
5. merges the ordered outputs before native sampling and request-state update.

ModelRunnerV2 stores stage-specific Ascend state in
`ForwardContext.additional_kwargs`. ModelRunnerV1 continues to use its existing
attributes, so this does not remove or redirect V1 state.

### DP dispatch and fallback

All DP ranks participate in one CPU-group reduction for token count,
uniform-token state, local DBO eligibility, and requested graph mode. This
keeps the eager or graph and single- or dual-batch decision consistent across
ranks.

DBO is used only when every rank allows it, the configured threshold is met,
and both microbatches still contain real work after final graph padding.
Otherwise, the step uses the native single-batch descriptor. If a rank cannot
use a graph, all ranks run the two microbatches eagerly for that step.

### Full ACL graph DBO

During ModelRunnerV2 initialization, a scoped wrapper replaces only the graph
manager factory used by the current AFD runner. `finally` blocks restore the
original vLLM and vLLM-Ascend symbols.

`AFDModelAclGraphManagerV2` keeps the native single-batch graph descriptors and
owns separate two-microbatch twins for eligible, evenly divisible capture
shapes. Each DBO graph includes:

- a warmup execution and a formal capture execution;
- two stage-specific forward contexts and metadata builders;
- per-stage MLA graph parameters backed by the native aggregate FIA workspace;
- staged replay state containing the real runtime token counts;
- AFD control metadata published before Attention and FFN capture/replay;
- the same FULL replay ordering fence required by vLLM-Ascend.

An uncaptured DBO shape uses eager DBO. A captured single-batch shape continues
through the native graph manager.

### Compatibility and isolation

- The validation relaxation applies only to active AFD NPU ModelRunnerV2 DBO
  configurations and restores the original config values after upstream
  validation.
- Graph-manager substitution is limited to one runner initialization scope and
  is always restored.
- Full-graph replay hooks are installed on one manager instance and restored
  after each execution scope.
- Non-AFD configurations preserve upstream validation and runner identity.
- DBO-disabled AFD ModelRunnerV2 execution continues to use the native runner.
- ModelRunnerV1 keeps its existing execution, forward-context, DBO, and ACL
  graph ownership.
- Plugin-owned DBO graphs and staged replay state are released during cleanup.

## Configuration boundary

NPU ModelRunnerV2 DBO requires:

- vLLM `v0.26.0` without native ModelRunnerV2 DBO descriptors;
- the synchronous `CAMP2pAFDConnector`;
- `compute_gate_on_attention=false`;
- DP size greater than one;
- exactly two microbatches;
- matching configured AFD ranks and `DP * TP`;
- PP, PCP, and DCP equal to one;
- static expert parallelism;
- a registered AFD model;
- eager execution, or ACL graph mode `FULL_DECODE_ONLY`.

Unsupported combinations fail during validation before model execution.

### Unit and compatibility coverage

- DBO threshold selection and uniform DP padding.
- Single-batch fallback when a rank is below threshold.
- Single-batch fallback when graph padding would create an empty second stage.
- Eager DBO fallback when DP ranks disagree on graph availability.
- Compatibility with the pinned upstream graph-manager dispatch signature.
- Request-boundary slicing, all-padding trailing slices, metadata-builder
  selection, and structured output merging.
- AFD-only configuration-validation relaxation and restoration of original
  values.
- Rejection of unsupported DBO topology, graph mode, and model features.
- ModelRunnerV2 forward-context storage without changing ModelRunnerV1 state.
- Graph-manager/replay-hook scoping, restoration, failure cleanup, and
  non-AFD/native-runner isolation.
- FFN graph replay for repeated capture keys.

### Ascend E2E coverage

The hardware E2E matrix includes:

- ModelRunnerV2 eager DBO off versus on;
- ModelRunnerV2 `FULL_DECODE_ONLY` DBO off versus on;
- ModelRunnerV1 DBO regression coverage;
- deterministic output consistency for the corresponding DBO on/off pairs.

## Test result

- The Ascend E2E consistency suite passed for NPU ModelRunnerV2 DBO, including
  eager and `FULL_DECODE_ONLY` graph execution.
- The NPU ModelRunnerV1 DBO regression check passed on its existing
  implementation path.
- DeepSeek-V2-Lite performance validation showed no DBO performance benefit.

## Limitations and follow-up

- This compatibility backport is not the long-term home for generic
  ModelRunnerV2 DBO behavior.
- Current NPU support is restricted to two microbatches and the validated AFD
  topology above.
- Once the pinned vLLM release provides native ModelRunnerV2 DBO descriptors,
  dispatch, slicing, and execution, remove the compatibility package and scoped
  validation and graph-manager patches. Retain only NPU-specific execution and
  ACL graph integration that is still needed.

---

<details>
<summary>Essential PR checklist</summary>

- [x] Eager and `FULL_DECODE_ONLY` DBO behavior is documented.
- [x] The temporary vLLM v0.26 backport and its removal condition are explicit.
- [x] DP synchronization, fallback, slicing, and graph ownership are described.
- [x] Unsupported topology and model-feature boundaries are stated.
- [x] ModelRunnerV1 and non-AFD isolation are documented.
- [x] Ascend consistency and performance results are stated.
- [x] No upstream vLLM or vLLM-Ascend files are modified.

</details>

### DIFF
--- a/afd_plugin/compat/backports/vllm_v026_mrv2_dbo/__init__.py
+++ b/afd_plugin/compat/backports/vllm_v026_mrv2_dbo/__init__.py
@@ -1,0 +1,27 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
+"""Temporary vLLM 0.26 ModelRunnerV2 eager DBO backport."""
+
+from .runtime import (
+    AFDBatchExecutionDescriptor,
+    assert_backport_required,
+    create_ubatch_slices,
+    dispatch_afd_dbo_and_sync_dp,
+    prepare_attn_for_ubatch,
+    share_metadata_builder_workspaces,
+    slice_input_batch,
+    slice_model_inputs,
+    use_two_metadata_builders,
+)
+
+__all__ = [
+    "AFDBatchExecutionDescriptor",
+    "assert_backport_required",
+    "create_ubatch_slices",
+    "dispatch_afd_dbo_and_sync_dp",
+    "prepare_attn_for_ubatch",
+    "share_metadata_builder_workspaces",
+    "slice_input_batch",
+    "slice_model_inputs",
+    "use_two_metadata_builders",
+]
--- a/afd_plugin/compat/backports/vllm_v026_mrv2_dbo/__init__.py
+++ b/afd_plugin/compat/backports/vllm_v026_mrv2_dbo/__init__.py
@@ -4,11 +4,9 @@
 
 from .runtime import (
     AFDBatchExecutionDescriptor,
-    assert_backport_required,
     create_ubatch_slices,
     dispatch_afd_dbo_and_sync_dp,
     prepare_attn_for_ubatch,
-    share_metadata_builder_workspaces,
     slice_input_batch,
     slice_model_inputs,
     use_two_metadata_builders,
@@ -16,11 +14,9 @@ from .runtime import (
 
 __all__ = [
     "AFDBatchExecutionDescriptor",
-    "assert_backport_required",
     "create_ubatch_slices",
     "dispatch_afd_dbo_and_sync_dp",
     "prepare_attn_for_ubatch",
-    "share_metadata_builder_workspaces",
     "slice_input_batch",
     "slice_model_inputs",
     "use_two_metadata_builders",
--- a/afd_plugin/compat/backports/vllm_v026_mrv2_dbo/__init__.py
+++ b/afd_plugin/compat/backports/vllm_v026_mrv2_dbo/__init__.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
-"""Temporary vLLM 0.26 ModelRunnerV2 eager DBO backport."""
+"""Temporary vLLM 0.26 ModelRunnerV2 DBO backport."""
 
 from .runtime import (
     AFDBatchExecutionDescriptor,
--- a/afd_plugin/compat/backports/vllm_v026_mrv2_dbo/execute.py
+++ b/afd_plugin/compat/backports/vllm_v026_mrv2_dbo/execute.py
@@ -1,0 +1,213 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
+"""vLLM 0.26 ModelRunnerV2 execute path with the eager DBO seams added."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import torch
+from vllm.config import CUDAGraphMode
+from vllm.forward_context import (
+    BatchDescriptor,
+    get_forward_context,
+    set_forward_context,
+)
+from vllm.sequence import IntermediateTensors
+from vllm.v1.worker.gpu.attn_utils import build_slot_mappings_by_layer
+from vllm.v1.worker.gpu.cudagraph_utils import get_uniform_token_count
+from vllm.v1.worker.gpu.input_batch import InputBatch
+from vllm.v1.worker.gpu.model_runner import ExecuteModelState
+from vllm_ascend.worker.v2.input_batch import AscendInputBatch
+
+from .runtime import (
+    AFDBatchExecutionDescriptor,
+    create_ubatch_slices,
+    dispatch_afd_dbo_and_sync_dp,
+)
+
+if TYPE_CHECKING:
+    from vllm.v1.core.sched.output import SchedulerOutput
+    from vllm.v1.outputs import ModelRunnerOutput
+
+
+def execute_model_v026_eager_dbo(
+    runner: Any,
+    scheduler_output: SchedulerOutput,
+    intermediate_tensors: IntermediateTensors | None = None,
+    *,
+    dummy_run: bool = False,
+    skip_attn_for_dummy_run: bool = False,
+    is_profile: bool = False,
+) -> ModelRunnerOutput | IntermediateTensors | None:
+    """Execute the supported plain-decoder subset with eager DBO."""
+
+    if not dummy_run:
+        runner.update_pp_decode_requests()
+        runner.finish_requests(scheduler_output)
+        runner.free_states(scheduler_output)
+        runner.add_requests(scheduler_output)
+        runner.update_requests(scheduler_output)
+        runner.block_tables.apply_staged_writes()
+        if scheduler_output.total_num_scheduled_tokens == 0:
+            return runner.kv_connector.no_forward(scheduler_output)
+
+    num_reqs = len(scheduler_output.num_scheduled_tokens)
+    num_tokens = int(scheduler_output.total_num_scheduled_tokens)
+    max_query_len = max(scheduler_output.num_scheduled_tokens.values())
+    uniform_token_count = get_uniform_token_count(
+        num_reqs,
+        num_tokens,
+        max_query_len,
+    )
+    batch_desc, num_tokens_across_dp = dispatch_afd_dbo_and_sync_dp(
+        num_reqs=num_reqs,
+        num_tokens=num_tokens,
+        uniform_token_count=uniform_token_count,
+        dp_size=runner.dp_size,
+        dp_rank=runner.dp_rank,
+        parallel_config=runner.parallel_config,
+        decode_query_len=runner.decode_query_len,
+        allow_ubatching=not skip_attn_for_dummy_run,
+    )
+    if batch_desc.num_tokens == 0:
+        return runner.kv_connector.no_forward(scheduler_output)
+
+    num_ubatches = (
+        batch_desc.num_ubatches
+        if isinstance(batch_desc, AFDBatchExecutionDescriptor)
+        else 1
+    )
+    if not dummy_run:
+        runner.input_buffers.is_padding[:num_tokens].fill_(False)
+        runner.input_buffers.is_padding[
+            num_tokens : batch_desc.num_tokens
+        ].fill_(True)
+        input_batch = runner.prepare_inputs(scheduler_output, batch_desc)
+        block_tables, slot_mappings = runner.prepare_attn(input_batch)
+        runner.model_state.preprocess_state(
+            input_batch,
+            block_tables,
+            runner.kv_cache_config,
+            runner.req_states.num_computed_tokens.gpu,
+        )
+    else:
+        dummy_batch_cls = AscendInputBatch if num_ubatches > 1 else InputBatch
+        input_batch = dummy_batch_cls.make_dummy(
+            batch_desc.num_reqs or num_reqs,
+            batch_desc.num_tokens,
+            runner.input_buffers,
+        )
+        if not skip_attn_for_dummy_run:
+            block_tables, slot_mappings = runner.prepare_dummy_attn(input_batch)
+        else:
+            block_tables = None
+            slot_mappings = None
+
+    attn_metadata = None
+    slot_mappings_by_layer = None
+    ubatch_slices = None
+    if num_ubatches > 1:
+        assert runner.ubatch_runner is not None
+        assert block_tables is not None and slot_mappings is not None
+        ubatch_slices = create_ubatch_slices(input_batch, num_ubatches)
+    elif not (dummy_run and skip_attn_for_dummy_run):
+        assert block_tables is not None and slot_mappings is not None
+        slot_mappings_by_layer = build_slot_mappings_by_layer(
+            slot_mappings,
+            runner.kv_cache_config,
+        )
+        attn_metadata = runner.model_state.prepare_attn(
+            input_batch,
+            CUDAGraphMode.NONE,
+            block_tables,
+            slot_mappings,
+            runner.attn_groups,
+            runner.kv_cache_config,
+        )
+
+    model_inputs = {
+        "input_ids": input_batch.input_ids,
+        "positions": input_batch.positions,
+        "inputs_embeds": None,
+        "intermediate_tensors": None,
+        **runner.model_state.prepare_inputs(input_batch, runner.req_states),
+    }
+    runner.eplb.prepare_forward(
+        runner.model_config,
+        input_batch.num_tokens,
+        ubatch_slices,
+    )
+
+    if ubatch_slices is not None:
+        batch_descriptor = BatchDescriptor(
+            num_tokens=input_batch.num_tokens_after_padding,
+            has_lora=False,
+            num_active_loras=0,
+        )
+        with set_forward_context(
+            None,
+            runner.vllm_config,
+            num_tokens=input_batch.num_tokens_after_padding,
+            cudagraph_runtime_mode=CUDAGraphMode.NONE,
+            num_tokens_across_dp=num_tokens_across_dp,
+            batch_descriptor=batch_descriptor,
+            ubatch_slices=ubatch_slices,
+            is_padding=input_batch.is_padding,
+        ):
+            parent_context = get_forward_context()
+            afd_metadata = parent_context.additional_kwargs["afd_metadata"]
+            afd_metadata.tokens_unpadded_lens = [
+                max(
+                    0,
+                    min(int(stage.token_slice.stop), int(input_batch.num_tokens))
+                    - int(stage.token_slice.start),
+                )
+                for stage in ubatch_slices
+            ]
+            ubatch_state = runner.ubatch_runner.prepare(
+                input_batch,
+                block_tables,
+                slot_mappings,
+                ubatch_slices,
+                parent_context,
+            )
+            runner.kv_connector.pre_forward(scheduler_output)
+            model_output = runner.ubatch_runner.run(
+                runner.model,
+                model_inputs,
+                ubatch_state,
+            )
+    else:
+        batch_descriptor = BatchDescriptor(
+            num_tokens=input_batch.num_tokens_after_padding,
+            has_lora=False,
+            num_active_loras=0,
+        )
+        with set_forward_context(
+            attn_metadata,
+            runner.vllm_config,
+            num_tokens=input_batch.num_tokens_after_padding,
+            cudagraph_runtime_mode=CUDAGraphMode.NONE,
+            num_tokens_across_dp=num_tokens_across_dp,
+            batch_descriptor=batch_descriptor,
+            slot_mapping=slot_mappings_by_layer,
+            is_padding=input_batch.is_padding,
+        ):
+            runner.kv_connector.pre_forward(scheduler_output)
+            model_output = runner.model(**model_inputs)
+
+    assert runner.is_last_pp_rank
+    assert isinstance(model_output, torch.Tensor)
+    runner.execute_model_state = ExecuteModelState(
+        input_batch=input_batch,
+        attn_metadata=attn_metadata,
+        slot_mappings_by_layer=slot_mappings_by_layer,
+        hidden_states=model_output,
+        aux_hidden_states=None,
+        finished_req_ids=scheduler_output.finished_req_ids,
+    )
+    return None
+
+
+__all__ = ["execute_model_v026_eager_dbo"]
--- a/afd_plugin/compat/backports/vllm_v026_mrv2_dbo/execute.py
+++ b/afd_plugin/compat/backports/vllm_v026_mrv2_dbo/execute.py
@@ -16,10 +16,10 @@ from vllm.forward_context import (
 from vllm.sequence import IntermediateTensors
 from vllm.v1.worker.gpu.attn_utils import build_slot_mappings_by_layer
 from vllm.v1.worker.gpu.cudagraph_utils import get_uniform_token_count
-from vllm.v1.worker.gpu.input_batch import InputBatch
 from vllm.v1.worker.gpu.model_runner import ExecuteModelState
 from vllm_ascend.worker.v2.input_batch import AscendInputBatch
 
+from . import runtime as dbo_runtime
 from .runtime import (
     AFDBatchExecutionDescriptor,
     create_ubatch_slices,
@@ -29,6 +29,17 @@ from .runtime import (
 if TYPE_CHECKING:
     from vllm.v1.core.sched.output import SchedulerOutput
     from vllm.v1.outputs import ModelRunnerOutput
+
+
+_EXPECTED_RUNTIME_ABI = 3
+_loaded_runtime_abi = getattr(dbo_runtime, "AFD_MRV2_DBO_RUNTIME_ABI", 1)
+if _loaded_runtime_abi != _EXPECTED_RUNTIME_ABI:
+    raise ImportError(
+        "AFD ModelRunnerV2 DBO backport modules are out of sync: "
+        f"execute expects runtime ABI {_EXPECTED_RUNTIME_ABI}, but loaded "
+        f"ABI {_loaded_runtime_abi}. Reinstall afd-plugin from one checkout "
+        "and restart every worker process."
+    )
 
 
 def execute_model_v026_eager_dbo(
@@ -92,8 +103,7 @@ def execute_model_v026_eager_dbo(
             runner.req_states.num_computed_tokens.gpu,
         )
     else:
-        dummy_batch_cls = AscendInputBatch if num_ubatches > 1 else InputBatch
-        input_batch = dummy_batch_cls.make_dummy(
+        input_batch = AscendInputBatch.make_dummy(
             batch_desc.num_reqs or num_reqs,
             batch_desc.num_tokens,
             runner.input_buffers,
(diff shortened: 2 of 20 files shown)


afd_plugin/compat/backports/vllm_v026_mrv2_dbo/execute.py
```diff
@@ -0,0 +1,240 @@
...
+if _loaded_runtime_abi != _EXPECTED_RUNTIME_ABI:
+    raise ImportError(
+        "AFD ModelRunnerV2 DBO backport modules are out of sync: "
+        f"execute expects runtime ABI {_EXPECTED_RUNTIME_ABI}, but loaded "
+        f"ABI {_loaded_runtime_abi}. Reinstall afd-plugin from one checkout "
+        "and restart every worker process."
+    )
+
+
+def execute_model_v026_eager_dbo(
```
jiangkuaixue123, line 48: This function is a copied/adapted upstream execution path, but the AFD-specific differences are not marked. The repository guidelines require copied or wrapped upstream functions to have the patch reason/functionality/signature comments immediately above them and to surround only the AFD-specific deltas with `# ### PATCH START: ...` / `# ### PATCH END: ...`. Please re-copy from the exact pinned source and mark the adaptations so future vLLM upgrades can mechanically compare and reapply this backport. The same applies to the copied helpers in `runtime.py`.

jiangkuaixue123 (changes requested):
Thanks for the detailed design and test coverage. I found two blockers before this can merge:

1. The temporary backport copies/adapts upstream execution code but does not follow this repository's patch-marking requirements; see the inline comment.
2. The current head fails the `pre-commit` workflow. The failing job reports mypy errors in the changed source/test files and the SPDX hook modifies four touched test files. Please run the full pre-commit checks and make the workflow green.

afd_plugin/compat/backports/vllm_v026_mrv2_dbo/execute.py
```diff
@@ -0,0 +1,240 @@
...
+    dispatch_afd_dbo_and_sync_dp,
+)
+
+if TYPE_CHECKING:
+    from vllm.v1.core.sched.output import SchedulerOutput
+    from vllm.v1.outputs import ModelRunnerOutput
+
+
+_EXPECTED_RUNTIME_ABI = 3
+_loaded_runtime_abi = getattr(dbo_runtime, "AFD_MRV2_DBO_RUNTIME_ABI", 1)
```
jiangkuaixue123: This module-to-module ABI handshake looks over-defensive. These modules are shipped from the same package/checkout, and there is already a second proactive guard in `assert_backport_required()`. The `getattr(..., 1)` fallback also masks the actual missing-symbol failure. Please remove this private ABI protocol (and the single-use descriptor-field sentry) and let the pinned vLLM contract plus normal import/static-test failures expose drift.

afd_plugin/compat/backports/vllm_v026_mrv2_dbo/runtime.py
```diff
@@ -0,0 +1,441 @@
...
+            return AFDBatchExecutionDescriptor(
+                cg_mode=CUDAGraphMode.NONE,
+                num_tokens=tokens,
+                num_reqs=num_reqs,
+                uniform_token_count=uniform_tokens,
+                num_active_loras=base.num_active_loras,
+                num_ubatches=ubatches,
+            )
+
+        dispatch_ubatches = getattr(cudagraph_manager, "dispatch_ubatches", None)
```
jiangkuaixue123: Please avoid probing this contract with `getattr` and replacing the original failure with a custom `RuntimeError`. The DBO initialization path installs `AFDModelAclGraphManagerV2`, which defines `dispatch_ubatches`; call `cudagraph_manager.dispatch_ubatches(...)` directly (and preferably give the manager a concrete protocol/type). If that contract changes, the original attribute/type failure should remain visible, per the repository's upstream-compatibility guidance.

afd_plugin/compat/backports/vllm_v026_mrv2_dbo/runtime.py
```diff
@@ -0,0 +1,441 @@
...
+def share_metadata_builder_workspaces(
+    attn_groups: list[list[AttentionGroup]],
+) -> None:
+    """Share the backend workspace while retaining independent builders."""
+
+    workspace = None
+    for groups in attn_groups:
+        for group in groups:
+            for builder in group.metadata_builders:
+                if workspace is None and hasattr(builder, "_get_workspace_buffer"):
```
jiangkuaixue123: These `hasattr` branches silently turn an ABI mismatch into partially initialized graph state. Because this backport targets pinned vLLM/vLLM-Ascend versions, access `_get_workspace_buffer()` / `set_workspace_buffer()` directly with the expected builder type and let an upstream incompatibility fail at its source. This removes defensive branching and makes static checking useful.

jiangkuaixue123 (commented):
A follow-up pass focused specifically on abstraction and defensive programming. Most of the larger boundaries (`AFDAscendUBatchRunnerV2`, graph-entry state, and the scoped graph-manager context managers) are justified because they isolate substantial lifecycle/state. The overengineering is concentrated in compatibility sentries and optional attribute probing. `assert_backport_required()` is also a single-use helper that duplicates the module ABI sentry; I suggest removing both and relying on the pinned ABI plus direct accesses/tests.

afd_plugin/compat/backports/vllm_v026_mrv2_dbo/execute.py
lirx-pd: All the related files have been refactored

hsliuustc0106 (commented):
Four actionable findings from review of the current head.

afd_plugin/compat/backports/vllm_v026_mrv2_dbo/__init__.py
```diff
@@ -0,0 +1,23 @@
+# SPDX-License-Identifier: Apache-2.0
```
hsliuustc0106, line 1: **[P1] Package the new backport modules**

Package discovery sets `namespaces = false`, but `afd_plugin/compat/backports` has no `__init__.py`. I verified that `setuptools.find_packages()` excludes `afd_plugin.compat.backports.vllm_v026_mrv2_dbo`; checkout tests pass through implicit namespace imports, while an installed wheel will omit these modules and fail the new imports. Please add the parent package marker or enable namespace discovery, plus a wheel-import smoke test.

afd_plugin/v1/worker/npu/attention_model_runner_v2.py
```diff
@@ -81,6 +95,10 @@ def run_fullgraph(
         self: v2_cudagraph_utils.ModelCudaGraphManager,
         desc: v2_cudagraph_utils.BatchExecutionDescriptor,
     ) -> torch.Tensor | tuple[torch.Tensor, list[torch.Tensor]] | IntermediateTensors:
+        # ### PATCH START: bypass native replay metadata for AFD DBO graphs.
+        if isinstance(desc, AFDBatchExecutionDescriptor):
+            return original_run_fullgraph(desc)
```
hsliuustc0106: **[P1] Mark DBO graph replay in control metadata**

This early return bypasses the code that sets `_afd_is_graph_replaying = True`. `AFDModelAclGraphManagerV2.run_fullgraph()` subsequently calls `send_dp_metadata()`, so the FFN receives `is_graph_replaying=False` and selects eager execution rather than its matching captured graph. Please wrap this call with the same replay-flag save/set/restore logic, or set the flag in the plugin manager before it sends control.

afd_plugin/v1/worker/npu/ubatch_runner_v2.py
```diff
@@ -0,0 +1,363 @@
...
+            try:
+                # ### PATCH START: Ascend worker device
+                torch.npu.set_device(self.device)
+                # ### PATCH END: Ascend worker device
+                with context:
+                    outputs[context.id] = model(**inputs)
+            except BaseException as error:  # noqa: BLE001
+                errors[context.id] = error
+                # ### PATCH START: Failed worker release
+                self.ready_barrier.abort()
```
hsliuustc0106: **[P1] Cancel event waits after a stage failure**

Aborting `ready_barrier` only releases threads still entering their contexts. Once execution begins, stages wait on `cpu_wait_event`; if one stage fails, its sibling can resume, reach the next DBO yield, and wait forever because the failed stage can no longer signal it. `close_execution()` then blocks indefinitely in `thread.join()`. Please add shared cancellation checked by the event-wait loop and wake every stage on failure, with a regression test that fails one stage mid-forward.

afd_plugin/v1/worker/npu/attention_model_runner_v2.py
```diff
@@ -415,6 +471,13 @@ def shutdown(self) -> None:
             stop_afd_npu_profiler(self.prof)
         finally:
             try:
+                # ### PATCH START: AFD MRV2 DBO graph cleanup
+                if isinstance(
+                    self.cudagraph_manager,
+                    AFDModelAclGraphManagerV2,
+                ):
+                    self.cudagraph_manager.clear_afd_graphs()
```
hsliuustc0106: **[P2] Release ubatch ownership before native shutdown**

`clear_afd_graphs()` only clears graph entries. Both this runner and the graph manager still retain `ubatch_runner`, which retains `model_state`; `ModelState` owns the model itself. Therefore `super().shutdown()` deleting `self.model_state` and `self.model` does not release the weights, defeating same-process memory reclamation. Please break these plugin-owned references before native shutdown.

afd_plugin/v1/worker/npu/attention_model_runner_v2.py
hsliuustc0106: [P1] Preserve replay state for AFD DBO descriptors. execute_model resets _afd_is_graph_replaying to false, and this branch delegates before setting it true. AFDModelAclGraphManagerV2.run_fullgraph then sends the control payload with that false value, while the FFN graph policy selects REPLAY only when is_graph_replaying is true. As a result, the Attention DBO graph replays but the matching FFN graph runs eagerly. Please set and restore the replay flag around this delegation and add a cross-role payload/replay regression test.

afd_plugin/validation.py
```diff
@@ -147,8 +147,31 @@ def validate_npu_model_runner_v2_config(
...
+            raise RuntimeError(
+                "AFD NPU ModelRunnerV2 DBO requires DP > 1 and exactly two ubatches",
+            )
+        if (
+            not vllm_config.model_config.enforce_eager
+            and cudagraph_mode_name(vllm_config) != "FULL_DECODE_ONLY"
+        ):
+            raise RuntimeError(
+                "AFD NPU ModelRunnerV2 DBO ACL graph requires FULL_DECODE_ONLY",
+            )
```
hsliuustc0106: [P1] Reject unsupported non-MLA DBO graph configurations. A registered Qwen3MoeForCausalLM configuration with use_mla=false, DP=2, two ubatches, and FULL_DECODE_ONLY passes this validator. The new AFDModelAclGraphManagerV2 capture path then unconditionally requires an MLA/FIA workspace and raises. Please either require MLA here for DBO graph mode or provide a backend-generic capture path, with a non-MLA regression test.

hsliuustc0106 (commented):
Independent review found two blocking correctness gaps in the NPU ModelRunnerV2 DBO graph path.

afd_plugin/compat/backports/vllm_v026_mrv2_dbo/__init__.py
lirx-pd: Fix this problem

afd_plugin/v1/worker/npu/attention_model_runner_v2.py
lirx-pd: Fix this problem

afd_plugin/v1/worker/npu/ubatch_runner_v2.py
lirx-pd: Fix this problem

afd_plugin/v1/worker/npu/attention_model_runner_v2.py
lirx-pd: Fix this problem

afd_plugin/v1/worker/npu/attention_model_runner_v2.py
lirx-pd: Fix this problem

afd_plugin/validation.py
lirx-pd: Fix this problem

hsliuustc0106 (commented):
Delta re-review of `d209945` against the last reviewed head `884ac64` (one commit, +319/-31; static review; pre-commit/DCO green).

All six prior blocking findings are verified fixed in-code, each with a regression test: backport modules now packaged in the wheel (new `compat/backports/__init__.py` + `test_package.py` smoke test); replay flag set on the AFD-descriptor graph path (`attention_model_runner_v2.py:107` → payload → FFN, with the cross-role test asserting `is_graph_replaying` and `replay_count == 1`); sibling event-wait deadlock closed via the shared `cancellation_event` (`ubatching.py`) and `cancel_execution()` waking every `cpu_wait_event` (5s-bounded regression test); shutdown releases the ubatch runner before native teardown; non-MLA DBO graph configs rejected at validation (`validation.py:162`, correctly scoped so eager non-MLA DBO still passes); earlier patch-marking, private-ABI, and `getattr`/`hasattr` items all gone. Note jiangkuaixue123's reviewed head `616d97b` was orphaned by a rebase; those findings were re-verified directly at this head.

Remaining items on this delta:

[P1] Provide verifiable NPU evidence for the blocking claims — PR body, "Test result" section
The MRV2 DBO eager + FULL_DECODE_ONLY E2E matrix and the DeepSeek-V2-Lite no-benefit perf claim are prose-only: no logs, commands, model/config details, dates, or numbers anywhere in the body or replies, and docs/npu/TESTING.md:12 states a CPU run does not replace the Ascend check. The in-tree NPU-gated regression tests cover the fixed findings, but merge needs the run artifacts (or an explicit maintainer waiver recording that in-repo coverage was accepted instead).

[P2] Mark the wheel smoke test `vllm_runtime` or inspect the wheel instead of importing it — tests/unit/package/test_package.py:24
The subprocess imports `afd_plugin.compat.backports`, whose `__init__` pulls torch/vLLM, so the documented CPU-only suite (`pytest -m "not gpu and not vllm_runtime"`) fails on macOS/CPU-only dev environments where vLLM is intentionally optional.

[P2] The body documents a startup capability guard that no longer exists — afd_plugin/compat/backports/vllm_v026_mrv2_dbo/__init__.py:1
The `BatchExecutionDescriptor`/`num_ubatches` fail-fast was removed with the re-export refactor, so the PR body (and its guard node in the diagram) no longer matches the code; on the next pin upgrade nothing fails loudly about native DBO shadowing the backport. Restore a one-line field check, or update the body/diagram — the body is the removal-condition record for this temporary backport, so it should be reconciled before merge.

[P2] Keep `shutdown()` idempotent — afd_plugin/v1/worker/npu/attention_model_runner_v2.py:481
`del manager.ubatch_runner` / `del self.ubatch_runner` raise `AttributeError` on a second shutdown call (error-path retry / double teardown), surfacing after `connector.close()` has already run; the V1 twin is idempotent.

Also: the PR currently has merge conflicts with main, so a rebase is needed regardless of review outcome.

PR #279, opened 2026-08-27
fix(e2e): clean up NPU async FFN process trees

### LINKED ISSUE #251: NPU async E2E teardown always fails pytest

ShwStone:
## Summary

- tag NPU async E2E process trees by run and role
- kill every matching FFN process through procfs during teardown
- keep one bounded 120-second cleanup window
- retain focused procfs, process-identity, and role-isolation coverage

## Testing

- 149 related unit tests
- pre-commit
- real procfs cleanup smoke test

Fixes #251

### DIFF
--- a/tests/e2e/process_utils.py
+++ b/tests/e2e/process_utils.py
@@ -8,7 +8,28 @@ import os
 import signal
 import subprocess
 import time
-from collections.abc import Sequence
+from collections.abc import Collection, Mapping, Sequence
+from contextlib import suppress
+from dataclasses import dataclass
+from pathlib import Path
+
+PROC_ROOT = Path("/proc")
+
+
+@dataclass(frozen=True)
+class ProcessIdentity:
+    """A PID plus a handle that remains bound to the same Linux process."""
+
+    pid: int
+    pidfd: int
+
+
+def close_process_identities(processes: Sequence[ProcessIdentity]) -> None:
+    """Close pidfds returned by ``find_processes_matching_environment``."""
+    for process in processes:
+        # Closing an owned pidfd is best effort during process teardown.
+        with suppress(OSError):
+            os.close(process.pidfd)
 
 
 def terminate_process_groups(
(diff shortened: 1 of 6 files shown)


tests/e2e/process_utils.py
```diff
@@ -127,3 +139,80 @@ def terminate_process_groups(
...
+    reported_signal_failures: set[int] = set()
+    deadline = time.monotonic() + timeout_s
+
+    while matching_pids := find_processes_matching_environment(
+        required_environment,
+        proc_root=proc_root,
+    ):
+        for pid in matching_pids:
+            try:
+                os.kill(pid, signal.SIGKILL)
```
jiangkuaixue123: [P2] Please avoid signaling a recycled PID. `find_processes_matching_environment()` closes the procfs files and returns bare numeric PIDs; while this loop is running, a matched worker can exit and its PID can be reused before `os.kill()`. In that case teardown can SIGKILL an unrelated process that never carried these markers—the preceding group SIGKILL makes this race especially relevant. Please keep a stable process identity across matching and signaling (for example, open a pidfd before reading `/proc/<pid>/environ` and signal that pidfd, or verify the process start time immediately before signaling), and add a test covering the exit/reuse window.

jiangkuaixue123 (commented):
The run/role tagging and bounded procfs cleanup are well scoped, and the related 148 unit tests pass locally. I found one process-identity race that should be addressed before merge.

tests/e2e/process_utils.py
ShwStone: Fixed in 45c4ccf: the scanner opens a pidfd before reading environ and SIGKILL is sent through that stable handle. Added a regression test for exit and immediate numeric-PID reuse.

ShwStone:
Given #281, I tested this PR against vLLM-Ascend `e0a5027c` (the parent of `f1180d906`); the NPU async E2E test passed.

PR #284, opened 2026-08-28
refactor(npu): isolate DSV4 Async CAM forward

ShwStone:
## Purpose

Follow up on #264 by separating the model-owned DSV4 Async CAM ubatch orchestration from the role-aware model definitions.

## Scope

- Move the existing token-ID staging and Async CAM forward pipeline into a dedicated DSV4 module.
- Keep the native unsplit forward path and the two-stage CAM send/receive order unchanged.
- Replace local magic values and `Any` schedule callbacks with named constants and concrete types.
- Do not change common runners, connectors, metadata, layouts, or tests.

## Validation

- `python3 -m py_compile afd_plugin/model_executor/models/npu/deepseek_v4.py afd_plugin/model_executor/models/npu/deepseek_v4_async_cam_forward.py`
- `git diff --check`
- Unit and NPU E2E tests were not run for this refactor-only change.

## Docs Impact

None; runtime behavior and user configuration are unchanged.

### DIFF
--- a/afd_plugin/model_executor/models/npu/deepseek_v4_async_cam_forward.py
+++ b/afd_plugin/model_executor/models/npu/deepseek_v4_async_cam_forward.py
@@ -1,0 +1,311 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
+"""DeepSeek V4 model-owned Async CAM ubatch execution."""
+
+from __future__ import annotations
+
+from copy import copy
+from itertools import islice
+from typing import Final
+
+import torch
+from vllm.distributed import get_pp_group
+from vllm.distributed.parallel_state import get_tp_group
+from vllm.forward_context import (
+    ForwardContext,
+    get_forward_context,
+    override_forward_context,
+)
+from vllm.sequence import IntermediateTensors
+from vllm_ascend.models import deepseek_v4 as native
+
+from afd_plugin.connectors import AFDTransferContext, AFDTransferMetadata
+from afd_plugin.model_executor.models import get_afd_metadata_from_forward_context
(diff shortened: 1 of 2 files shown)


afd_plugin/model_executor/models/npu/deepseek_v4_async_cam_forward.py
```diff
@@ -0,0 +1,308 @@
...
+            hidden_states = inputs_embeds
+        else:
+            hidden_states = model.embed_input_ids(input_ids)
+        hidden_states = hidden_states.unsqueeze(1).repeat(1, model.hc_mult, 1)
+    else:
+        if intermediate_tensors is None:
+            raise ValueError("DSV4 pipeline stage requires intermediate tensors")
+        hidden_states = intermediate_tensors["hidden_states"]
+
+    parent_context = get_forward_context()
```
jiangkuaixue123, line 111: [P1] Import `get_forward_context` before calling it. This symbol was available in the original module, but the extracted module imports only `ForwardContext` and `override_forward_context`. Every DSV4 Async CAM ubatch request reaches this line and raises `NameError`, so this path is currently unusable. `ruff check` also reports F821 here; please add the missing import and a regression test that enters `run_async_moe_ubatch_forward()`.

jiangkuaixue123 (changes requested):
The extracted Async CAM path currently has a blocking runtime failure.

afd_plugin/model_executor/models/npu/deepseek_v4_async_cam_forward.py
ShwStone: Fixed in `db4a33f` by importing `get_forward_context` in the extracted module. `uv run ruff check` now passes for both affected DSV4 files. Per the current refactor scope, no new test was added.

PR #289, opened 2026-08-31
fix: restore request-count DPLB for async Attention DP

### LINKED ISSUE #278: [Prefill-only] Implement vLLM-aware DPLB for AFD, then evaluate DP-wave
## Summary

The immediate priority is to make **vLLM's existing request-count-aware DPLB
work correctly for asynchronous AFD Attention DPs**. DP-wave should be treated
as a separate, second-stage optimization only after the native DPLB path is
working, observable, tested, and benchmarked on hardware.

> [!IMPORTANT]
> The performance numbers in this issue come from an experimental
> discrete-event simulator using an analytic msModeling profile. They are
> **not hardware measurements**. The simulator is useful for prioritizing the
> work, but it does not prove production performance.

The simulator indicates that functioning vLLM-aware DPLB already captures a
large part of the available benefit. At QPS 4, it reduces simulated P99 TTFT by
57.8% versus round-robin. The proposed DP-wave policy improves further, but
that incremental opportunity should not delay the simpler and more urgent
native DPLB work.

## Priority 1: implement and validate vLLM-aware DPLB for AFD

vLLM's DP frontend routes requests using per-DP `waiting` and `running`
request counts. For this policy to work, every asynchronous AFD Attention DP
must publish fresh counts and the frontend must consume them while routing new
requests.

AFD Attention DPs must remain asynchronous. Restoring request-count reporting
must **not** reintroduce synchronous DP execution, a cross-DP wave barrier,
global unfinished-request collectives, or `FIRST_REQ` wave wakeups.

The first implementation should stay deliberately narrow:

1. Publish each AFD Attention DP's current `(waiting, running)` counts at the
   cadence expected by the vLLM DP client/coordinator path.
2. Verify that the frontend receives those updates and uses the native vLLM
   request-count score when selecting an Attention DP.
3. Preserve the frontend's optimistic waiting-count increment so bursts that
   arrive between periodic publications do not all select the same DP.
4. Add metrics or debug logging that make the reported and frontend-observed
   counts inspectable during a benchmark.
5. Keep the first patch free of prompt-length scoring, predicted waves, or
   other experimental scheduling logic.

### Phase-1 acceptance criteria

- AFD asynchronous Attention DPs publish correct waiting/running counts while
  requests enter, run, and finish.
- The frontend demonstrably makes non-round-robin choices when queue counts
  differ.
- A burst arriving inside one stats-publication interval is distributed using
  the frontend's optimistic local counts.
- No synchronous DP-wave coordination or cross-DP execution barrier is
  enabled.
- Unit/integration tests cover stale updates, ties, bursts, and request
  completion.
- A hardware benchmark compares round-robin with functioning vLLM-aware DPLB
  on the same fixed length and arrival trace.

## Simulator evidence used for prioritization

The experiment compares three **simulated** policies on the same AFD topology:

1. **AFD round-robin baseline**.
2. **vLLM count-aware baseline with functioning stats**, including a modeled
   100 ms publication cadence and frontend optimistic waiting-count updates.
3. **Proposed DP-wave token-sum**, which trial-places an arriving request on
   every DP and estimates its completion from future FIFO prefill waves.

Simulation configuration:

- Scope: prefill-only; no decode traffic.
- Model: DeepSeek-V4-Flash with an analytic msModeling profile.
- AFD topology: Attention `DP6×TP4` plus FFN `DP8×TP1×EP8`.
- Physical budget: 32 dies; Attention/FFN die ratio is `24/8 = 3`.
- Equal-die merged topology from the original control-grid selection:
  `DP8×TP4×EP32`. It is topology context only; it is not one of the three
  policies compared below.
- Workload: one fixed 512-request MoonConv V4 Flash `formal_0` trace cycle.
- Input lengths: minimum 893, P50 6,529, P95 32,238, maximum 63,007,
  mean 10,412, coefficient of variation 1.078.
- Arrival process: identical trace gaps scaled to offered QPS
  `0.5, 1, 2, 3, 4, 6`.
- `max_num_batched_tokens = chunk_size = 65,536`; `max_num_seqs = 64`.
- AFD token uBatch split: one request per uBatch.

| Offered QPS | Simulated policy | P50 TTFT (ms) | P99 TTFT (ms) | Requests/s | Input ktok/s | Assigned-token CV |
| ---: | --- | ---: | ---: | ---: | ---: | ---: |
| 3 | AFD round-robin | 2,691 | 12,169 | 2.873 | 29.91 | 8.99% |
| 3 | vLLM count-aware | 2,517 | 9,609 | 2.878 | 29.96 | 10.07% |
| 3 | Proposed DP-wave token-sum | 2,365 | 8,416 | 2.891 | 30.10 | 4.24% |
| 4 | AFD round-robin | 4,457 | 28,528 | 3.549 | 36.95 | 8.99% |
| 4 | vLLM count-aware | 3,601 | 12,026 | 3.716 | 38.70 | 4.91% |
| 4 | Proposed DP-wave token-sum | 3,130 | 8,546 | 3.786 | 39.42 | 2.41% |
| 6 | AFD round-robin | 17,631 | 54,027 | 3.804 | 39.61 | 8.99% |
| 6 | vLLM count-aware | 19,056 | 39,307 | 4.166 | 43.38 | 3.90% |
| 6 | Proposed DP-wave token-sum | 19,172 | 31,347 | 4.385 | 45.66 | 1.42% |

At QPS 4, functioning vLLM-aware DPLB provides the majority of the simulated
gain over round-robin:

- P99 TTFT: `28,528 ms → 12,026 ms` (`-57.8%`);
- P50 TTFT: `4,457 ms → 3,601 ms` (`-19.2%`);
- request throughput: `3.549 → 3.716 requests/s` (`+4.7%`); and
- assigned-token CV: `8.99% → 4.91%`.

DP-wave then provides an incremental improvement over functioning vLLM-aware
DPLB at the same QPS:

- P99 TTFT: `12,026 ms → 8,546 ms` (`-28.9%`); and
- request throughput: `3.716 → 3.786 requests/s` (`+1.9%`).

At QPS 6, functioning vLLM-aware DPLB improves throughput over round-robin by
9.5% and reduces P99 by 27.2%. DP-wave adds another 5.3% throughput and reduces
P99 by a further 20.3% relative to vLLM-aware DPLB. This point is overloaded,
and neither simulated DPLB policy improves P50.

At QPS 0.5 and 1, functioning count-aware DPLB is identical to round-robin in
the simulation because there is usually no concurrent count difference to
observe. This is another reason to implement the native policy first and
measure its real operating region instead of assuming universal benefits.

## Priority 2: evaluate token-aware DP-wave

After Phase 1 is complete and hardware results show meaningful residual
length-driven imbalance, we can evaluate a prefill-only DP-wave prototype.

The proposed online policy asks:

> If this arriving request is appended to a candidate DP's FIFO queue under
> the real max-batch and chunk constraints, in which wave will it finish, and
> which candidate DP predicts the earliest completion?

It does not reorder already queued requests and does not require Attention DPs
to execute synchronously. A detailed Chinese explanation of the simulator's
Wave Sum and Square-Sum designs is available here:

[Prefill-only 下的 Wave Sum / Square-Sum DP 调度](https://github.com/ShwStone/afd-plugin/blob/61b84064b63f79072833303286686abe2128c094/simulator/docs/prefill-wave-dplb.md)

That document describes an experimental simulator design in a personal fork;
it is background material, not a production patch proposed for merging
unchanged.

## Open questions

- What is the smallest patch that lets asynchronous AFD Attention cores publish
  native request counts without inheriting synchronous DP-wave behavior?
- Which metrics should expose engine-reported counts, frontend-observed counts,
  and routing decisions during Phase-1 validation?
- After vLLM-aware DPLB is working, how much imbalance remains on hardware and
  is it explained by prompt length?
- If Phase 2 is justified, should its score use token-sum, square-sum, or a
  calibrated profile-derived completion estimate?
- How should shared FFN pressure enter a future score when FFN is not idle or
  fully hidden by the AFD pipeline?

ShwStone:
## Purpose

Restore vLLM's native request-count-aware DPLB for asynchronous AFD Attention DPs, as proposed in #278.

Async Attention engines currently use `EngineCoreProc` to avoid MoE DP-wave synchronization, but vLLM 0.26.0 only publishes `(running, waiting)` scheduler counts from `DPEngineCoreProc`. As a result, the frontend DPLB client receives no fresh load information.

## Issue

- Related issue(s): #278
- This draft does not close #278 yet because hardware benchmark evidence is still pending.

## Scope

- In scope:
  - Publish changed Attention scheduler request counts before and after each independent engine step.
  - Preserve the DP coordinator's native 100 ms frontend publication cadence.
  - Backport non-lockstep coordinator stats handling from upstream vLLM #49204.
  - Keep native `waiting * 4 + running` routing and optimistic waiting-count increments.
  - Add debug visibility and focused lifecycle/routing regression coverage.
- Out of scope:
  - DP-wave, prompt-length, token-sum, KV-pressure, or predicted-completion scoring.
  - Hardware performance claims or benchmark results.
  - Synchronous cross-DP barriers, unfinished-request collectives, or `FIRST_REQ` wakeups for AFD async DP.

## Implementation Notes

- `EngineCoreProc.run_busy_loop` remains the independent AFD async Attention loop. It now emits changed-only `SchedulerStats` after input processing and after an engine step, making waiting, running, and completion transitions visible.
- `DPCoordinatorProc.process_input_socket` is copied from pinned vLLM 0.26.0. AFD-specific differences are marked and limit step/wave snapshots plus the 50 ms lockstep aggregation wait to `enable_wave_coordination=True`.
- AFD async launch continues to set `enable_wave_coordination=False` and skip client `FIRST_REQ`; native lockstep/non-AFD paths retain their original behavior.
- Engine-reported counts are logged at DEBUG level. The existing frontend `Received counts` DEBUG log remains unchanged.
- The compatibility patch remains version-guarded and idempotent. The coordinator backport can be removed once the pinned vLLM contains upstream #49204.

## Test Plan

- CPU-safe fake-runtime tests for request-count lifecycle, independent/stale step counters, lockstep preservation, AFD `FIRST_REQ` suppression, and non-AFD wakeup preservation.
- Pinned-vLLM runtime contract tests for non-round-robin routing, tied bursts with optimistic counts, and request completion cleanup.
- Hardware benchmark comparing round-robin and native request-count DPLB on the same fixed trace before marking the PR ready.

## Test Result

From a clean worktree based on `upstream/main`:

```text
$ python -m pytest -q tests/unit/compat/patches/test_async_dp_engine.py \
    tests/unit/compat/patches/test_async_dp_dplb_runtime.py \
    tests/unit/compat/patches/test_engine_core.py
................                                                         [100%]
```

The vLLM runtime contract module is skipped locally because vLLM is not installed in the CPU-only environment.

```text
$ python -m ruff check <changed files>
All checks passed!

$ python -m ruff format --check <changed files>
5 files already formatted

$ git diff --check
# clean
```

Hardware benchmark: not run yet; required before moving this PR out of draft.

## Docs Impact

- Files updated: none.
- Reason: Phase 1 restores the documented/native request-count DPLB behavior without adding a user-facing configuration option. Hardware results will be recorded separately.

---

<details>

- [x] Scope is bounded.
- [x] Plugin-owned classes or explicit dotted class paths are preferred over monkey patches where available.
- [x] Validation evidence is included, including skipped GPU/hardware tests.

</details>


afd_plugin/compat/patches/async_dp_engine.py
```diff
@@ -405,6 +688,7 @@ def apply_async_dp_engine_patch() -> bool:
     EngineCoreProc.run_engine_core = staticmethod(run_engine_core)
     engine_utils_module.launch_core_engines = launch_core_engines
     core_client_module.launch_core_engines = launch_core_engines
+    coordinator_module.DPCoordinatorProc.process_input_socket = process_input_socket
```
jiangkuaixue123: [P1] Ensure this patch reaches spawn-started coordinator processes

This assignment only mutates DPCoordinatorProc in the parent process, while the coordinator process target remains vllm.v1.engine.coordinator.DPCoordinatorProc.run_coordinator. With the multiprocessing spawn method, the child imports the upstream coordinator module afresh and resolves its original process_input_socket, so the non-lockstep handling added here is not guaranteed to run in the coordinator child. Please use a plugin-owned process target (or patch run_coordinator itself) and add a real spawn regression test.

jiangkuaixue123: [P2] Do not overwrite newer dev Coordinator implementations

The shared compatibility gate accepts every version string containing dev. Adding this global assignment therefore replaces the Coordinator implementation in newer vLLM development checkouts as well as in 0.26.0. Current upstream already contains the non-lockstep fix and carries KV-cache usage as a third load-balancing value, whereas this copied 0.26 implementation only updates waiting and running. That can silently regress load balancing for non-AFD DP configurations whenever the plugin is loaded. Please give this Coordinator backport its own exact-version or capability guard and leave an already-fixed upstream implementation intact.

afd_plugin/compat/patches/async_dp_engine.py
```diff
@@ -332,6 +335,286 @@ def launch_core_engines(
...
+    zmq_addr_pipe=None,
+):
+    time = coordinator_module.time
+    msgspec = coordinator_module.msgspec
+    zmq = coordinator_module.zmq
+    logger = coordinator_module.logger
+    make_zmq_socket = coordinator_module.make_zmq_socket
+    MsgpackDecoder = coordinator_module.MsgpackDecoder
+    EngineCoreOutputs = engine_core_module.EngineCoreOutputs
+    EngineState = coordinator_module.EngineState
```
jiangkuaixue123, line 401: [P2] Mark this non-upstream scaffolding as an AFD delta

These local aliases are not present in the copied upstream process_input_socket body, so they are AFD-specific patch scaffolding. The repository patching rules require every difference from the pinned upstream function to be enclosed by PATCH START/PATCH END markers. Please mark this block, or import the upstream symbols at module scope so that the copied function body remains directly comparable with upstream.

jiangkuaixue123 (changes requested):
Requesting changes for two runtime-boundary issues and one patch-maintenance violation. The pinned vLLM 0.26.0 + fork path otherwise looks sound.

ShwStone:
Hardware benchmark update (preliminary; keeping this PR in Draft):

- Topology: single-node Attention `DP2 x TP4` + FFN `EP8`
- Workload: fixed 512-request dataset
- Comparison: restored request-count DPLB baseline vs. opt-in prefill-token policy
- Peak throughput: `15,870 -> 16,211` (`+2.1%`)
- Tail drain time: `31.5 s -> 24.5 s` (`-7.0 s`, `-22.2%`)

The result shows a modest peak-throughput improvement and a clearer reduction in tail drain time on this workload. This is one single-node measurement, so the PR remains Draft while the outstanding review items and broader validation are handled separately.

ShwStone:
A carefully constructed worst-case request test shows that DPLB works as expected.

PR #234, opened 2026-08-13
Derive CAM HCCL buffer sizes per role

ShwStone:
## Summary

- derive independent Attention and FFN CAM HCCL buffer sizes from the model, scheduler, topology, and quantization configuration
- apply 1.1x headroom through connector-scoped HCCL process-group options for both CAM Async and CAMP2P
- log each role's selected buffer at startup
- warn when the memory outside `gpu_memory_utilization` is less than 2.5x the role-local CAM buffer, without changing the configured utilization
- preserve the process-wide `HCCL_BUFFSIZE` setting for unrelated HCCL groups
- update the CAM Async and CAMP2P guides, launch recipes, and CPU-safe coverage

## Motivation

`HCCL_BUFFSIZE` is process-wide, so increasing it for CAM also enlarges unrelated TP, DP, PCP, or EP communication buffers. The derived size is now scoped to connector-owned HCCL process groups. `HCCL_BUFFSIZE` is neither overwritten nor unset, so users can still tune other HCCL groups intentionally.

The derived requirements are:

```text
tokens_per_npu = ceil(max_num_batched_tokens / num_npus_per_dp_group)
attention_bytes = 2 * hidden_size * tokens_per_npu * (topk + 1)
ffn_bytes = num_attention_ranks * (6176 if dynamic quantization else 12288) * tokens_per_npu
role_buffer_mb = ceil(1.1 * role_bytes / 1_MiB)
```

The FFN formula intentionally has no `topk + 1` multiplier. `num_npus_per_dp_group` describes the total NPU count within one Attention DP group; the currently supported topology realizes that dimension with TP.

## Validation

- `uv run --locked ruff check .`
- `uv run --locked ruff format --check .`
- `uv run --locked pytest -q tests/unit -m 'not gpu and not vllm_runtime'`
- `python3 -m compileall -q afd_plugin tests`
- `bash -n recipe/npu/CAMAsyncAFDConnector/deepseek_v3_2/v0_26_accuracy/attention_dp2tp8.sh recipe/npu/CAMAsyncAFDConnector/deepseek_v3_2/v0_26_accuracy/ffn_ep16.sh recipe/npu/CAMP2pAFDConnector/deepseek_v3_2/afd_attention.sh recipe/npu/CAMP2pAFDConnector/deepseek_v3_2/afd_ffn.sh`
- `git diff --check`

NPU end-to-end validation is still pending.

PR #291, opened 2026-08-31
Add connector-scoped HCCL buffer configuration

ShwStone:
## Summary

- add optional `connector_extra_config.hccl_buffer_size` support for CAM Async and CAMP2P
- apply fresh `ProcessGroupHCCL.Options` only to connector-owned HCCL domains
- preserve the existing `HCCL_BUFFSIZE` and torch-npu default fallback when the option is unset
- keep the connector call sites ready for the future #234 automatic sizing work through `self.hccl_buffer_size_mb`
- keep unit coverage minimal: one helper test plus updates to the two existing connector initialization tests

## Motivation

`HCCL_BUFFSIZE` is process-wide, so increasing it for CAM also enlarges unrelated TP, DP, and EP communication buffers. The optional setting scopes the override to CAM-owned process groups.

This restarts the work from #207 while leaving the formula work in #234 out of scope. The retained v0.19 recipe is unchanged because its recorded validation predates this option.

## Validation

- `git ls-files -z '*.py' | xargs -0 .venv/bin/ruff check`
- `git ls-files -z '*.py' | xargs -0 .venv/bin/ruff format --check`
- `python3 -m compileall -q afd_plugin tests`
- `git diff --check upstream/main...HEAD`
- fake-module smoke coverage for HCCL options fallback, configured value, and fresh instances
- async CAM E2E environment/command construction smoke coverage
- final architectural review: Merge verdict OK

The focused pytest suite could not execute in the local environment: upstream's current `uv.lock` is rejected as stale by uv 0.12.5, and the existing `--no-sync` environment does not contain torch/vLLM. NPU hardware validation is pending.

Replaces closed PR #207, which GitHub cannot reopen after its head branch was force-pushed during the rebase.

### DIFF
--- a/afd_plugin/connectors/npu/async_cam.py
+++ b/afd_plugin/connectors/npu/async_cam.py
@@ -31,7 +31,7 @@ import os
 from collections.abc import Mapping
 from dataclasses import dataclass
 from datetime import timedelta
-from typing import TYPE_CHECKING, Any, Final
+from typing import TYPE_CHECKING, Any, Final, cast
 
 import torch
 from torch import Tensor
@@ -44,6 +44,7 @@ from afd_plugin.config_utils import (
     coerce_extra_int,
     coerce_extra_positive_int,
     coerce_extra_str,
+    coerce_optional_extra_positive_int,
 )
 from afd_plugin.connectors.base import (
     AFDConnectorBase,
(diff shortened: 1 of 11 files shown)


jiangkuaixue123:
Fix pre-commit

PR #307, opened 2026-09-04
perf(npu): fuse CAM shared expert activation

jiangkuaixue123:
Keeps CAM dynamicQuant shared inputs in the INT8 path through W13, fused dequant-SwiGLU-quant, and W2. Adds contract coverage for the fused scale handoff.\n\nValidation: python3 -m compileall and git diff --check. Pytest could not run because the local .venv pytest launcher points to a missing interpreter.

### DIFF
--- a/afd_plugin/model_executor/models/npu/deepseek_v2_attention_gate.py
+++ b/afd_plugin/model_executor/models/npu/deepseek_v2_attention_gate.py
@@ -184,6 +184,7 @@ def compute_attention_gate_moe_ffn(
                     experts._shared_experts,
                     shared_input,
                     shared_scales,
+                    swiglu_limit=layer.mlp.swiglu_limit,
                     output_dtype=torch.bfloat16,
                 )
             else:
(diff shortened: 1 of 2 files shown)


afd_plugin/model_executor/models/npu/deepseek_v2_attention_gate.py
```diff
@@ -184,6 +184,7 @@ def compute_attention_gate_moe_ffn(
                     experts._shared_experts,
                     shared_input,
                     shared_scales,
+                    swiglu_limit=layer.mlp.swiglu_limit,
```
ShwStone: `layer.mlp.swiglu_limit` raises `AttributeError` on the DSV2/V3.2 gate-on-Attention path. The FFN role constructs `vllm.model_executor.models.deepseek_v2.DeepseekV2MoE` (the adapter imports vLLM's module, and pinned vLLM-Ascend has no NPU `deepseek_v2` model module), and that class never sets a `swiglu_limit` attribute; only the vLLM-Ascend `DeepseekV4MoE` does. Any DSV2-family CAM Async run with W8A8 + `dynamicQuant=1` fails here.

Could we use the same sentinel as the native shared-expert executor?

```python
swiglu_limit=getattr(layer.mlp, "swiglu_limit", None),
```

The helper already maps `None` to `clamp_limit=0.0`, matching vLLM's unclamped `SiluAndMul` and the native `0.0 if ... is None` convention, while DSV4 keeps its real limit.

ShwStone (changes requested):
This fused path matches the native Ascend W8A8 shared-expert arithmetic (W13 INT32 accumulator -> `npu_dequant_swiglu_quant` -> W2 with `activation_scale` as `pertoken_scale`), and the scale handoff looks correct. One blocker below; two non-blocking notes:

1. Native gates `glu_alpha`/`glu_bias` behind `HardwareCapability.FUSED_SWIGLU_TUNING_ARGS`; this patch passes them unconditionally. Fine for current CAM Async hardware, but worth a comment for future profiles.
2. The source-string test asserts an attribute the real model class lacks (the fake MLP sets `swiglu_limit=7.0` manually), which masks the runtime failure described in the thread.

jiangkuaixue123:
Addressed the review in 38ae22d. CAM shared experts now use getattr(layer.mlp, "swiglu_limit", None), so DSV2/V3.2 retain the native unclamped SiluAndMul behavior; the unit fixture now omits the attribute to cover that path. I also documented the current CAM hardware contract for fused SwiGLU tuning arguments.

PR #325, opened 2026-09-09
Upgrade the GPU backend to vLLM v0.28.0

yujuancao07:
## Purpose

Rebases the AFD GPU backend from vLLM `v0.26.0` (`568afb3a1`) to vLLM
`v0.28.0` (`2cf0a6915`, the `releases/v0.28.0` tag) and re-validates the
full GPU E2E matrix on the new runtime. 

 **NPU execution is unsupported on this release until the NPU upgrade lands**; 
the NPU docs are kept as the pending-upgrade baseline.

## Upstream changes that drove the adaptations

| vLLM 0.26.0 → 0.28.0 change | AFD adaptation |
| --- | --- |
| `FusedMoE` replaced by `FusedMoEFactory`/`MoERunner` (#44941); gate computation moved into the runner; `is_internal_router` removed | FFN external routing keeps `experts.gate = None` (the runner now holds the shared gate); external-routing asserts use the runner-gate attribute; NPU factory refresh rebinds `FusedMoEFactory` |
| `DPCoordinatorProc` gates lockstep timeouts and step/wave ordering on `enable_wave_coordination` (#49204); LB counts widened to `[waiting, running, kv_cache_usage]` | **Removed** the copied `run_coordinator`/`process_input_socket` patches and their exact-release gate — upstream absorbs all AFD deltas once launch passes `enable_wave_coordination=False` |
| `_maybe_publish_request_counts` hoisted to the base busy loop | **Removed** the AFD async-Attention publication loop (upstream publishes the same counts at the same points, plus `kv_cache_usage`) |
| `launch_core_engines` yields `CoreEngineLaunch` (no `num_api_servers`); `run_engine_core` uses `reconfigure_for_independent_dp_rank()` | Rebuilt from the 0.28.0 skeletons, AFD markers replayed |
| `EngineCore.__init__` gained `_weight_version` and the `ec_connector` aggregator; warmup split into `Executor.compile_or_warm_up_model()`; busy loops wrapped in `@fault_tolerant_wrapper` | `engine_core.py` rebuilt from the target skeletons; FFN divert and late-loaded noop scheduler preserved |
| V1 `_build_attention_metadata` gained mid-signature `max_num_sampled_tokens`; `_dummy_run` gained `randomize_inputs`; warmup synchronizes before capture | Signatures matched, keyword forwarding, sync barrier replayed |
| V2 `prepare_inputs_to_capture` dropped `skip_attn` for required `full_cudagraph` + `max_query_len`; `execute_model` gained `context_len` | Capture-input wrapper rewritten; `context_len` forwarded |
| qwen3_moe/qwen3_5 ctor kwargs; `ParallelLMHead.tie_weights()`; qwen3_next SP MoE rework; DSV4 `eager_scratch_pool` + layout-derived `use_sequence_parallel` | Role-aware construction adapted; DSV4 pins `use_sequence_parallel=False` (native derivation would enable SP for AFD DP2TP2) |

## Validation

Runtime: vLLM 0.28.0 wheel, torch 2.13.0+cu130, Python 3.12, 8× NVIDIA H200, 
DeepSeek-V2-Lite / Qwen3-30B-A3B / Qwen3.6-35B-A3B.

## Docs

README, connector/recipe READMEs, design-module docs, GPU guide, and
issue/PR templates refreshed to the 0.28.0 identity; NPU statements
downgraded to the pending-upgrade baseline; historical v0.19/v0.26 NPU
records preserved as historical.


afd_plugin/model_executor/models/deepseek_v2.py
```diff
@@ -243,8 +243,11 @@ class AFDDeepseekV2RemoteExpertsMoE(native.DeepseekV2MoE):
     # constructing only the gate owned by Attention and a parameter-free proxy.
     # Signature: AFD-owned; adds layer_idx and compute_gate_on_attention and omits
     # quant_config because no local expert kernel is constructed.
-    # Upstream: vLLM v0.26.0, vllm/model_executor/models/deepseek_v2.py
-    # Commit: 568afb3a13806beb53bb2e6bd518269357b237c0
+    # Upstream: vLLM v0.28.0, vllm/model_executor/models/deepseek_v2.py
+    # Commit: 2cf0a6915ce544dc493a0990f2ea38d81601128a
+    # In v0.28.0 the native forward always delegates routing to the experts
```
jiangkuaixue123, line 248: [P1] Compute the Attention-side gate before forwarding router logits

With CUDA DeepSeek and `compute_gate_on_attention=True`, the inherited vLLM 0.28 `DeepseekV2MoE.forward()` always calls `self.experts(hidden_states=hidden_states, router_logits=hidden_states)` and no longer computes the outer gate. However, `AFDAttentionFusedMoE.forward()` only forwards its `router_logits` argument; it never invokes the gate retained on the Attention-side shell. Consequently, hidden states are sent as logits while the FFN runner's gate is cleared. A CPU comparison using the actual old/new upstream forward methods and the AFD proxy produced one gate call and a `[2, 64]` payload on 0.26, versus zero gate calls and a `[2, 2048]` payload on 0.28. The P2P receiver still allocates logits with width `n_routed_experts`, so this breaks the send/receive tensor-size contract for DeepSeek-V2-Lite. Please restore gate computation before transport and cover the `compute_gate_on_attention=True` forward path. Could you confirm this behavior?

afd_plugin/compat/patches/engine_core.py
```diff
@@ -450,74 +467,38 @@ def run_busy_loop(self):
...
     connector = None
+    ec_connector = None
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         pass
 
     def get_kv_connector(self) -> None:
         return None
 
+    def get_kv_event_publisher_config(self) -> None:
```
jiangkuaixue123, line 496: [P1] Apply the new FFN handshake contract to early-loaded initialization too

Adding this accessor to `_AFDFFNNoopScheduler` fixes the late-loaded path, but `_initialize_ffn_engine_core()` still assigns `self.scheduler = None`. When the patch is already installed before FFN EngineCore construction (for example, a fork inheriting the parent's installed patches), that early-return initialization path is selected instead. The native 0.28 `_make_ready_response()` now unconditionally calls `self.scheduler.get_kv_event_publisher_config()`, so the input thread fails before completing the READY handshake. Calling the actual extracted FFN initializer followed by the target upstream ready-response method reproduces `AttributeError: 'NoneType' object has no attribute 'get_kv_event_publisher_config'` on CPU. Please make the early-loaded path satisfy the same handshake contract and test both plugin-loading orders. Could you confirm whether this startup path is supported and reproduce the failure?

jiangkuaixue123 (commented):
@yujuancao07 Could you please confirm the two findings below against this revision, including whether the affected configurations/startup paths are intended to remain supported?

Both findings come from comparing the pinned vLLM 0.26.0 and 0.28.0 sources and running CPU contract reproductions using extracted source methods; they have not been reproduced on GPUs. Please check the Attention-side gate path and both early-loaded and late-loaded FFN initialization paths, and add regression coverage if confirmed.

For clarity, I am not treating the initial Legacy E2E CI failure as a code regression: it was a disk-space error and the retry passed. The remaining NPU unit failures have not been established as regressions introduced by this PR.

afd_plugin/model_executor/models/deepseek_v2.py
yujuancao07: Confirmed and fixed[[2cef4f9](https://github.com/vllm-project/afd-plugin/pull/325/commits/2cef4f955375cf1363cfa503b93c79e01100ca63)]
The 0.28 native forward always delegates routing to the runner, so the shell's gate was never invoked; the shell now overrides forward and computes gate before transport when it owns a gate. Regression tests cover both the gated transport (gate called once, [2, n_routed_experts] logits) and the gate-less native delegation. compute_gate_on_attention=True on CUDA DeepSeek remains a supported configuration.

afd_plugin/compat/patches/engine_core.py
yujuancao07: Confirmed and fixed[[8a72056](https://github.com/vllm-project/afd-plugin/pull/325/commits/8a72056c7bb00198cb6992dfbe86cfa9a1a7d0d7)]
The early-loaded path now installs the same noop scheduler the late-loaded path uses, satisfying get_kv_event_publisher_config() in _make_ready_response. Regression tests now cover both plugin-loading orders (pre-installed patches → early divert; mid-init install → late path) and enumerate the handshake surface. Both startup paths remain supported.

afd_plugin/compat/patches/engine_core.py
```diff
@@ -450,74 +467,38 @@ def run_busy_loop(self):
...
+        # Publish request counts before and after GPU step to ensure freshness.
+        self._maybe_publish_request_counts()
         # 2) Step the engine core and return the outputs.
         self._process_engine_step()
+        self._maybe_publish_request_counts()
 
     raise SystemExit
 
 
-def _run_async_attention_busy_loop(self) -> None:
```
jiangkuaixue123, line 459: Could you explain why `_run_async_attention_busy_loop` can be removed here? The updated comments say that vLLM 0.28 now publishes load-balancer stats from the base busy loop. Could you point to the corresponding upstream change and clarify how it preserves the behavior this helper provided: independent Attention stepping, request-count publication before and after each step, and no DP-wave coordination or cross-DP collectives? Please also point to the regression coverage that verifies this replacement for AFD async-DP.

Issue #327, opened 2026-09-09
[Bug]: FFN worker initialization fails when FFN ranks span multiple hosts


swjeong9:
### Current environment

```text
Python: 3.12.13
OS: Ubuntu 22.04.5 LTS in the container / Amazon Linux 2023 on the host (kernel 6.12.100-125.179.amzn2023.x86_64)
afd-plugin commit: 117ebf0 (unmodified)
vLLM version/checkout: v0.26.0
CUDA/GPU: NVIDIA L4, one per process. Driver 580.178.04 (open kernel module), CUDA 13.0, NCCL 2.28.9+cuda13.0
Install command: an image built on vllm/vllm-openai:v0.26.0, with the vLLM v0.26.0
                 checkout and afd-plugin each installed through `pip install -e`
Deployment: AWS EKS 1.33, hostNetwork, one process per pod
Hosts: one g6.12xlarge (4x L4) and two g6.xlarge (1x L4 each), all in us-west-2a
Network: TCP over ENA. No EFA is attached, so NCCL falls back to the socket transport
Run: 2A2F, TP 1, DP 2 per role, expert parallel enabled, eager, DBO disabled
```

### Reproduction

We deployed `2A2F` across three hosts.

| Placement   | g6.12xlarge | g6.xlarge#1 | g6.xlarge#2 | Split role    |
| ----------- | ----------- | ----------- | ----------- | ------------- |
| **B** | F0, F1      | A0          | A1          | Attention     |
| **C** | A0, A1      | F0          | F1          | **FFN** |

- In placement B the Attention workers are scheduled onto separate hosts.
- In placement C the FFN workers are scheduled onto separate hosts.

The commands are below.

```bash
A=2; F=2
MODEL=deepseek-ai/DeepSeek-V2-Lite
COMMON=(--served-model-name deepseek-v2-lite --trust-remote-code --enable-expert-parallel
        --max-model-len 4096 --max-num-seqs 128 --max-num-batched-tokens 128
        --gpu-memory-utilization 0.9 --host 0.0.0.0 --data-parallel-size-local 1 --enforce-eager)
afd() { printf '{"afd":{"role":"%s","connector":"P2pNcclAFDConnector","host":"%s","port":6269,"num_attention_ranks":%d,"num_ffn_ranks":%d}}' "$1" "$FFN0_IP" "$A" "$F"; }

# FFN nodes, rank r = 0..F-1
vllm serve $MODEL "${COMMON[@]}" \
[...]

PR #328, opened 2026-09-09
fix(gpu): reuse the AFD world's store for subgroup rendezvous

swjeong9:
## Purpose

When the FFN ranks span multiple hosts, FFN worker initialization fails with `OSError: [Errno 99] Cannot assign requested address` (#327). Each subgroup brings up a store server of its own and takes `afd_config.host` as its bind address, so any FFN rank other than F0 attempts to bind an address that is not local to its host.

Following the design @specture724 outlined, a subgroup no longer brings up a store of its own; it reuses the store the AFD world has already rendezvoused on.

## Issue

- Related issue(s): #327

## Scope

In scope

- Subgroup rendezvous for the GPU P2P connector when the FFN ranks span multiple hosts.
- Removal of the derived subgroup port from the connector and its user guide.
- Unit coverage for per-subgroup store key namespacing.

Out of scope

- The rendezvous of the AFD world and the `p2p` control-plane group
- The NPU connector

## Implementation Notes

The subgroup store is the AFD world process group's store, obtained through `ProcessGroup.get_group_store()` and wrapped in a per-subgroup `PrefixStore`, so no additional rendezvous endpoint is created.

## Test Plan

- `2A2F` initialization on EKS. One g6.12xlarge (4x L4) and two g6.xlarge, running placement B (Attention split) and placement C (FFN split) against the images built before and after the change.
- `2A2F` startup and serving over EFA. Four g6.8xlarge with one rank each, followed by 256 requests.

## Test Result

- `pytest tests/unit`: 830 passed, 2 failed, 62 skipped. Both failures originate in `test_e2e_process_utils.py`, which requires the Linux-only `os.pidfd_open`; that attribute is unavailable on the macOS machine used for local CPU testing.
- `2A2F` on EKS across three hosts, run against the image built before the change and an image carrying this PR.

| Placement | Split role | 117ebf0 | This PR |
| --- | --- | --- | --- |
| B | Attention | succeeds | succeeds |
| **C** | **FFN** | **fails** | **succeeds** |

Before the change, FFN rank 1 terminates with `OSError: [Errno 99] Cannot assign requested address`; afterwards, initialization completes correctly.

The verification above was carried out over TCP (ENA). We repeated the check on EFA, with `2A2F` on four `g6.8xlarge`. Under the same condition as placement C, initialization succeeded on EFA as well, and serving proceeded correctly under `vllm bench`. The log below identifies the network interface that was selected.

```
NCCL INFO NCCL_NET_PLUGIN set by environment to /opt/amazon/ofi-nccl/lib/libnccl-net.so
NCCL INFO NET/OFI Initializing aws-ofi-nccl 1.21.1
NCCL INFO NET/OFI Using Libfabric version 2.6
NCCL INFO NET/OFI Selected provider is efa, fabric is efa (found 1 nics)
NCCL INFO NET/OFI Using transport protocol SENDRECV
NCCL INFO Using network Libfabric
```

- That said, g6 instances currently do not support GPUDirect RDMA, so we have not yet been able to reach the hang path @specture724 described.

## Docs Impact

- Files updated: three places in `docs/gpu/NCCL_P2P_CONNECTOR_USER_GUIDE.md`, together with the module and class docstrings in `afd_plugin/connectors/gpu/p2p.py`. The wording that required a derived port (`port + subgroup_index + 1`) has been removed.

### DIFF
--- a/afd_plugin/connectors/gpu/p2p.py
+++ b/afd_plugin/connectors/gpu/p2p.py
@@ -51,8 +51,9 @@ Requirements and limitations:
       share that rendezvous and do not need ports of their own.
     - AFD async mode (``async`` / ``async_dp``) is not supported; GPU DBO
       combined with CUDA graphs is limited to exactly two ubatches.
-    - Cross-node use is not established by the checked-in recipes and should
-      be treated as unverified.
+    - Cross-node placement is verified for DeepSeek-V2-Lite ``2A2F`` with the
+      FFN ranks on separate hosts, over TCP (ENA) and over EFA. Other
+      topologies and hardware should be treated as unverified.
 
 See ``docs/gpu/NCCL_P2P_CONNECTOR_USER_GUIDE.md`` for the configuration
 contract and launch examples, and ``recipe/gpu/p2p_nccl/`` for complete
@@ -275,13 +276,15 @@ class P2pNcclAFDConnector(AFDConnectorBase):
             # rendezvoused on. StatelessProcessGroup.create would stand up a
             # second store bound to ``host``, which is only correct while
             # every subgroup's rank 0 sits on that host.
+            subgroup_store = PrefixStore(
+                f"afd_subgroup_{self.mapping.subgroup_index}",
+                afd_pg.get_group_store(),
+            )
+            subgroup_store.set_timeout(timedelta(seconds=300))
             self.a2e_group = StatelessProcessGroup(
                 rank=self.mapping.rank_in_subgroup,
                 world_size=len(self.mapping.subgroup_ranks),
-                store=PrefixStore(
-                    f"afd_subgroup_{self.mapping.subgroup_index}",
-                    afd_pg.get_group_store(),
-                ),
+                store=subgroup_store,
             )
             self.e2a_group = self.a2e_group
             self.a2e_pynccl = PyNcclCommunicator(
--- a/afd_plugin/connectors/gpu/p2p.py
+++ b/afd_plugin/connectors/gpu/p2p.py
@@ -47,8 +47,8 @@ Requirements and limitations:
       size/dtype, and role-rank assignment. Initialization is collective:
       missing ranks, mismatched counts, or duplicate role ranks can cause
       initialization failure or timeout.
-    - The rendezvous base ``port`` and the derived subgroup ports
-      (``port + subgroup_index + 1``) must be free and reachable.
+    - The rendezvous ``port`` on ``host`` must be free and reachable. Subgroups
+      share that rendezvous and do not need ports of their own.
     - AFD async mode (``async`` / ``async_dp``) is not supported; GPU DBO
       combined with CUDA graphs is limited to exactly two ubatches.
     - Cross-node use is not established by the checked-in recipes and should
@@ -66,7 +66,11 @@ from datetime import timedelta
 from typing import TYPE_CHECKING, Any, NamedTuple
 
 import torch
-from torch.distributed.distributed_c10d import ProcessGroup, _get_default_group
+from torch.distributed.distributed_c10d import (
+    PrefixStore,
+    ProcessGroup,
+    _get_default_group,
+)
 from vllm.distributed.device_communicators.pynccl import PyNcclCommunicator
 from vllm.distributed.utils import StatelessProcessGroup
 from vllm.forward_context import DPMetadata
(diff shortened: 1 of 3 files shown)


hsliuustc0106 (commented):
Reviewed at `ea4b6dc`. No blocking findings from my review — this closes the #327 root cause with proportionate evidence.

Verified against the pinned stack (vLLM 0.26.0 / torch 2.11.0):

- `StatelessProcessGroup` is a dataclass in v0.26.0 and `create()` only builds the TCPStore before constructing it, so constructing the subgroup group over a provided store is behaviorally equivalent for the pin.
- `ProcessGroup.get_group_store()` exists on torch 2.11.0 (vLLM 0.26.0's pin). It is absent on older torch, so the `TARGET_VLLM_VERSION` gate is what makes the call safe — on an unsupported stack it fails loudly rather than silently.
- Per-subgroup `PrefixStore` prefixes keep the PyNccl unique-id bootstrap keys disjoint; the new unit test proves the separation through a real store, and follows the CPU-safe import convention.

The 3-host EKS A/B (placement C fails on the base image, passes here) plus the EFA repeat covers this regression better than the single-node gate would. Stating the GPUDirect RDMA path as untested rather than approximating it is the right call.

Non-blocking notes:

- No stale `port + subgroup_index` references remain — guide, recipes, and tests are all consistent with the new rendezvous scheme.
- At review time only DCO had reported; as a fork PR this likely will not get the standard single-node GPU E2E gate. The EKS evidence substitutes well for this fix, but a same-repo `test-ready` run post-approval wouldn't hurt.
- Follow-up worth an issue: the GPUDirect RDMA rendezvous hang path from the #327 design discussion is still unexercised. Tracking it separately keeps this PR mergeable without waiting on hardware with GPUDirect support.

docs/gpu/NCCL_P2P_CONNECTOR_USER_GUIDE.md
specture724: Please modify here.

afd_plugin/connectors/gpu/p2p.py
```diff
@@ -266,12 +270,18 @@ def init_afd_connector(self) -> None:
...
+            # rendezvoused on. StatelessProcessGroup.create would stand up a
+            # second store bound to ``host``, which is only correct while
+            # every subgroup's rank 0 sits on that host.
+            self.a2e_group = StatelessProcessGroup(
                 rank=self.mapping.rank_in_subgroup,
                 world_size=len(self.mapping.subgroup_ranks),
+                store=PrefixStore(
+                    f"afd_subgroup_{self.mapping.subgroup_index}",
+                    afd_pg.get_group_store(),
+                ),
```
specture724: explicitly set timeout here, since the previous behavior is 300s timeout. After the modification, the timeout is set to120s by default ( `init_afd_process_group` set the store timeout to `store.set_timeout(timedelta(minutes=2))`)

Please add:
```python
subgroup_store.set_timeout(timedelta(seconds=300))
```

tests/unit/connectors/test_p2p_connector.py
```diff
@@ -671,3 +684,58 @@ def test_p2p_recv_single_rank_requires_ref_tensor():
...
+    from torch.distributed import HashStore
+
+    module = importlib.import_module("afd_plugin.connectors.gpu.p2p")
+
+    root_store = HashStore()
+    afd_pg = SimpleNamespace(get_group_store=lambda: root_store)
+
+    monkeypatch.setattr(module, "init_afd_process_group", lambda **kwargs: afd_pg)
+    monkeypatch.setattr(module, "_get_default_group", lambda: None)
+    monkeypatch.setattr(module, "DefaultProcessGroupSwitcher", _NullSwitcher)
```
specture724:
```python
monkeypatch.setattr(module, "DefaultProcessGroupSwitcher", lambda *a, **k: nullcontext())
```
and remove `_NullSwitcher`

docs/gpu/NCCL_P2P_CONNECTOR_USER_GUIDE.md
specture724: Modify this line since the multi-node run is verified

specture724 (commented):
Thank @swjeong9 's work. I only have some nits

docs/gpu/NCCL_P2P_CONNECTOR_USER_GUIDE.md
swjeong9: Updated.

afd_plugin/connectors/gpu/p2p.py
swjeong9: Thank you for pointing this out — applied.

tests/unit/connectors/test_p2p_connector.py
swjeong9: Thank you for the better approach — applied as suggested.

docs/gpu/NCCL_P2P_CONNECTOR_USER_GUIDE.md
swjeong9: Updated.

specture724:
Thanks, LGTM

jiangkuaixue123:
Fix precommit

swjeong9:
Fixed. Pre-commit passes locally.

swjeong9:
@jiangkuaixue123 Thank you for re-running the pipeline. Unfortunately, it looks like the Buildkite E2E job hit a runner disk issue: `OSError: [Errno 28] No space left on device` on `/fsx/hf_cache`, before any code runs.

PR #332, opened 2026-09-10
[1/5] feat: NVSHMEM symmetric window for GPU AFD transport

specture724:
First of a five-PR stack that brings up `GpuAsyncAFDConnector` and runs
DeepSeek-V4 on it. **Review and merge in order**; each PR is based on the one
before it.

1. **this PR** — NVSHMEM symmetric window (transport)

## What this adds

A one-sided transport with no control plane: a rank writes a whole slot into a
peer's symmetric window — header, routing indices, weights, payload — and stamps
a flag last, on the same stream, so a peer that sees the flag sees the payload.

The flag protocol is what makes the window CUDA-graph-capturable, and that fixes
its shape. A captured stream wait compares against a value baked in at capture
time, so a sender cannot signal with a fresh sequence number every replay. It
signals with a constant marker instead, and the reader resets the flag in-band
once it has consumed the slot, which makes every replay identical.

`cuda_rt` and `nvshmem_rt` are thin ctypes bindings: stream memory ops for the
in-stream wait, and NVSHMEM symmetric allocation plus peer pointer lookup.

## Scope

**No AFD concepts appear in this layer at all** — it is a wire format and the
driver calls under it, and it imports nothing from the rest of the plugin. That
is why it is a separate PR: it can be reviewed by itself.

## Testing

`tests/unit/connectors/gpu/test_symm_window.py` pins the slot layout and the
header codec (field offsets, what a header does and does not carry, corrupt
magic, shutdown flag round trip). Runs on CPU, no GPU needed.

Full unit suite is unchanged against `main`: same 13 pre-existing failures,
none added. `pre-commit` clean over the PR range.

### DIFF
--- a/afd_plugin/connectors/gpu/cuda_rt.py
+++ b/afd_plugin/connectors/gpu/cuda_rt.py
@@ -1,0 +1,125 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
+"""Stream memory operations, for waiting on a peer's flag without the host.
+
+A receive is a wait on one 32-bit word that a peer writes. Doing that wait on
+the host -- copy the flag down, look at it, try again -- costs a synchronize per
+attempt and, far worse, stops the host from queueing anything behind the wait.
+A profile of the async connector showed 86% of kernel launches starting within
+5us of being issued: the GPU was idle waiting to be fed one kernel at a time,
+because every layer blocked the host twice.
+
+``cuStreamWaitValue32`` moves the wait onto the stream. The host enqueues "wait
+until this word reaches N" and keeps going, so the work behind the wait is
+already queued when the flag arrives. It is a driver API with no runtime API
+equivalent and no PyTorch binding, hence ctypes -- the same approach
+``nvshmem_rt`` already takes for the symmetric allocator.
+"""
+
+from __future__ import annotations
+
+import ctypes
+from typing import Final
+
+# CUstreamWaitValue_flags. GEQ is a cyclic comparison, so a monotonically
+# increasing sequence number keeps working across 32-bit wraparound.
+CU_STREAM_WAIT_VALUE_GEQ: Final[int] = 0x0
+CU_STREAM_WAIT_VALUE_EQ: Final[int] = 0x1
+# CUdevice_attribute: stream memory ops must be supported by the device.
+_CU_DEVICE_ATTRIBUTE_CAN_USE_STREAM_MEM_OPS: Final[int] = 74
+
+_lib: ctypes.CDLL | None = None
+_wait_value32 = None
+_checked_devices: set[int] = set()
+
+
+def _load() -> ctypes.CDLL:
+    global _lib, _wait_value32
+    if _lib is not None:
+        return _lib
+
+    lib = ctypes.CDLL("libcuda.so.1")
+    # CUDA 11.7 renamed the entry point; the unsuffixed symbol still exists on
+    # some builds, so take whichever this driver exports.
+    for symbol in ("cuStreamWaitValue32_v2", "cuStreamWaitValue32"):
+        fn = getattr(lib, symbol, None)
+        if fn is not None:
+            fn.argtypes = [
+                ctypes.c_void_p,  # CUstream
+                ctypes.c_ulonglong,  # CUdeviceptr
+                ctypes.c_uint32,  # value
+                ctypes.c_uint32,  # flags
+            ]
+            fn.restype = ctypes.c_int
+            _wait_value32 = fn
+            break
+    else:
+        raise RuntimeError(
+            "libcuda.so.1 exports no cuStreamWaitValue32; this driver cannot "
+            "wait on a flag from a stream",
+        )
+    lib.cuDeviceGetAttribute.argtypes = [
+        ctypes.POINTER(ctypes.c_int),
+        ctypes.c_int,
+        ctypes.c_int,
+    ]
+    lib.cuDeviceGetAttribute.restype = ctypes.c_int
+    _lib = lib
+    return lib
+
+
+def require_stream_mem_ops(device_index: int) -> None:
+    """Fail loudly at setup if the device cannot wait on memory from a stream."""
+    if device_index in _checked_devices:
+        return
+    lib = _load()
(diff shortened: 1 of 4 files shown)


hsliuustc0106 (commented):
Consolidated review of `6509abd` (static review; pre-commit/DCO green). No blockers; five P2s:

[P2] Validate the symmetric-heap layout across ranks — afd_plugin/connectors/gpu/symm_window.py:328
`__init__` never exchanges `total_bytes`/capacities over the group, and the receiver only checks magic/version, so config drift between role deployments produces unequal heaps (deep transport crash) or silently clamped partial reads (`_view(...)[: sizes[0]]` at :472 and :735-755) with no error. All-gather the capacity tuple in `__init__` and raise on mismatch, naming the differing parameter.

[P2] Fence the constructor zero-reset against earlier peers — afd_plugin/connectors/gpu/symm_window.py:329
A peer returning from its own collective `malloc` can issue its first fabric write while this rank's `zero_()` is still queued behind stream work; the zero erases the arriving flag and the receiver spins forever. End `__init__` with a `group.barrier()` after the synchronize, or state the required post-construction collective as a class-level contract that #333's handshake must provably provide.

[P2] 1,191 of 1,336 added lines are unreachable at this head — afd_plugin/connectors/gpu/symm_window.py:303
Only the test file imports `symm_window`; the factory registry and `connectors/gpu/__init__.py` are untouched. Acceptable as part 1/5 if #333/#334 land immediately, but if the stack stalls, main carries dead transport whose ctypes ABI assumptions (struct sizes, `(1<<16)+sizeof` versioning, `libnvshmem_host.so.3` discovery) have zero executable evidence. The series should not be called complete before a GPU-gated test or E2E exercises the real window.

[P2] Justify the module-level mutable globals — afd_plugin/connectors/gpu/cuda_rt.py:31
`_lib`/`_wait_value32`/`_checked_devices` are new process-wide mutable state without the justification AGENTS.md requires; `nvshmem_rt.py:110` already models the expected comment.

[P2] Update the connector design page — docs/design/module/connector_contracts.md
`afd_plugin/connectors/**/*.py` is a primary path of this page: the NVSHMEM substrate needs an evidence-table row (window/rt files ↔ `test_symm_window.py`) plus a limitations line for the NVSHMEM host library, NVLink-reachable PEs, and stream-mem-ops device gate. Backend-mode/topology rows can ride with #333.

Validation gaps: no Buildkite test-ready run in the rollup; no GPU-marked test anywhere in the series yet; the docstring perf numbers (hot-spin vs backoff TTFT, launch-latency percentages) need linked artifacts when #334's DSV4 runs land.

specture724:
Thanks — all five addressed in `50c3e04`. The stack was rebased end to end, so 333–336 moved with it.

**Symmetric-heap layout across ranks.** `__init__` now all-gathers the geometry before allocating and raises naming the parameter that differs (`num_regions`, `ring_depth`, `header_words`, `partial_cap`, `token_cap`, `shared_cap`, `hidden_size`, `payload_itemsize`, `slot_bytes`, `total_bytes`). Erroring at construction beats either failure mode you named.

**Constructor zero-reset fence.** You were right that the collective malloc does not order this: it orders the allocation, and a peer that returns first can land a dispatch while the `zero_()` is still queued, erasing the arriving flag. `__init__` now ends with `zero_()` → `torch.cuda.synchronize()` → `dist.barrier(group)` → `torch.cuda.synchronize()`, so nobody writes until every rank has finished resetting. I took the barrier rather than the class-level contract, since it costs one collective at startup.

**Module-level mutable globals.** Documented, in the shape `nvshmem_rt.py:112` already models — the driver library is process-global and resolving the handle per call would repeat a dlopen on the layer path, which is the host cost the module exists to remove.

**Design page.** `connector_contracts.md` gains an evidence row (window/rt files ↔ `test_symm_window.py`, noted CPU-only) and a limitations paragraph covering the NVSHMEM host library, the NVLink-reachability requirement behind `nvshmem_ptr`, the stream-mem-ops device gate, and the fact that the ctypes ABI assumptions have no executable evidence until a GPU-gated test exists.

**Unreachable at this head.** No change — this is inherent to a 1/5 PR, and `git grep` confirms your count: `test_symm_window.py` is the only importer. I'd rather keep the layer reviewable on its own than merge it with its first consumer. Your condition stands: the series should not be called complete before a GPU-gated test or E2E exercises a real window, and I've recorded that in the limitations paragraph rather than leaving it implicit.

Validation gaps noted — the GPU-marked test and the artifacts behind the docstring's perf numbers are still outstanding.

hsliuustc0106 (commented):
Delta re-review of `50c3e040` against the previously reviewed `6509abd` (one commit, +82/−1).

Status of the five findings from the last review: (1) cross-rank geometry validation — fixed-verified (`_validate_uniform_geometry` at symm_window.py:336, all-gathers a 10-field tuple before `nvshmem_rt.init`, raises `ValueError` naming the differing parameter and per-rank values); (2) zero-reset fence — fixed-verified (order at symm_window.py:346-355 is `zero_()` → synchronize → `dist.barrier` → synchronize; no rank can issue fabric writes before every peer has reset); (3) unreachable-at-merge scope — still open by agreed condition (no production caller yet; fine as long as #333/#334 land and a GPU-gated test or E2E exercises the window before the series is called complete); (4) mutable-globals justification — fixed-verified (cuda_rt.py:31-36); (5) design-page row — fixed-verified (connector_contracts.md evidence row, honestly qualified CPU-only).

One new item in the delta:

[P2] Include the payload dtype itself in the geometry check — afd_plugin/connectors/gpu/symm_window.py:412
The gathered tuple carries `payload_itemsize` but not dtype identity, so fp16 vs bf16 (both 2 bytes) passes every check; the sender then writes bf16 bit patterns the receiver decodes as fp16 — silent wrong output, the same "invisible at the wire" class this validation exists to close. Add a dtype identity (e.g. `str(self.payload_dtype)` via `all_gather_object`, or a small dtype→int code) to the tuple.

Also: the new validation path has no test coverage (`test_symm_window.py` still exercises layout/codec only), and the constructor now requires a GPU-capable process group for the new collectives — worth stating in the class contract for #333's wiring.

Gate: DCO is ACTION_REQUIRED on this push — the new commit is missing its `Signed-off-by` trailer; please amend (`git commit --amend -s`) and re-push to clear it.

PR #333, opened 2026-09-10
[2/5] feat: the async GPU connector on the symmetric window

specture724:
Second of a five-PR stack. **Based on #332** — the diff shown here is only this
PR's own change; review #332 first.

2. **this PR** — the async GPU connector

## What this adds

`GpuAsyncAFDConnector` drives AFD's Attention-to-FFN traffic over the symmetric
window with **no control plane at all**. The Attention side routes tokens itself
and one-sided-writes the already-grouped rows into every FFN peer's slot; the
FFN side learns the layer index and the row counts from the arriving payload, so
nothing has to be broadcast ahead of a step and the two roles never rendezvous
per layer.

`plan_dispatch` produces one row per `(token, expert slot)`, sorted by global
expert, so a destination's partials are one contiguous run already grouped for
its grouped GEMM. **Nothing about the routing is ever read back to the host**:
the per-destination counts go into the slot header on the device, which is what
takes the last synchronize off the layer path.

The dispatch sequence number lives on the device for the same reason the
window's flag protocol uses a constant marker — a replay runs no Python, so a
host-side counter would freeze at whatever was live at capture.

## Shared with the NPU async connector

`AFDAsyncTopology` and `build_async_topology` move out of the NPU CAM connector
into `connectors/async_topology.py` so both async connectors share one
role-rank-to-world-rank mapping (logic unchanged — pure hoist). With more than
one async connector now, `AFD_ASYNC_CONNECTOR` becomes
`AFD_ASYNC_NPU_CONNECTOR`.

The shared Attention-gate forward gains an opaque-op dispatch path so Dynamo
splits at the MoE round trip instead of tracing into the connector. **That path
is gated on the connector type**: CAM keeps its direct calls, because the ops
carry neither router logits nor FlashComm1 token sharding and the CAM protocol
needs both.

## Testing

- `tests/unit/connectors/test_async_gpu_connector.py` — 33 tests pinning the
  protocol: routing invariants, the flag protocol, dispatch header assembly,
  ring allocation across stages.
- `tests/e2e/async_gpu_connector_e2e.py` (2 GPUs) — full round trip against a
  reference, eagerly and with the Attention half captured and replayed.
- `tests/e2e/async_gpu_moe_equivalence.py` (1 GPU) — the dispatch/compute/
  combine chain against a naive per-token MoE.

added. `pre-commit` clean over the PR range.

### DIFF
--- a/afd_plugin/compat/npu/feature_validation.py
+++ b/afd_plugin/compat/npu/feature_validation.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from afd_plugin.config import (
-    AFD_ASYNC_CONNECTOR,
+    AFD_ASYNC_NPU_CONNECTOR,
     AFDConfig,
     is_afd_async_dp,
     parse_afd_config,
@@ -38,7 +38,7 @@ def fail_if_unsupported_npu_afd_features(
     if is_dsv4:
         _fail_if_unsupported_dsv4_connector(afd_config)
 
-    if afd_config.connector == AFD_ASYNC_CONNECTOR:
+    if afd_config.connector == AFD_ASYNC_NPU_CONNECTOR:
         _fail_if_unsupported_npu_afd_async_features(
             vllm_config,
             afd_config,
@@ -130,7 +130,7 @@ def _fail_if_unsupported_dsv4_async_features(
 
 
 def _fail_if_unsupported_dsv4_connector(afd_config: AFDConfig) -> None:
-    if afd_config.connector != AFD_ASYNC_CONNECTOR:
+    if afd_config.connector != AFD_ASYNC_NPU_CONNECTOR:
         raise RuntimeError("DSV4 NPU AFD supports only CAMAsyncAFDConnector")
 
 
--- a/afd_plugin/config.py
+++ b/afd_plugin/config.py
@@ -16,14 +16,22 @@ if TYPE_CHECKING:
     from vllm.config import VllmConfig
 
 AFD_ADDITIONAL_CONFIG_KEY: Final[str] = "afd"
-AFD_ASYNC_CONNECTOR: Final[str] = "CAMAsyncAFDConnector"
+AFD_ASYNC_NPU_CONNECTOR: Final[str] = "CAMAsyncAFDConnector"
+AFD_ASYNC_GPU_CONNECTOR: Final[str] = "GpuAsyncAFDConnector"
+# Connectors that drive FFN work from the connector receive loop instead of a DP
+# metadata control plane, and therefore need the async-DP engine patches. The
+# patches are platform-neutral; only the connector below them differs.
+AFD_ASYNC_CONNECTORS: Final[frozenset[str]] = frozenset(
+    {AFD_ASYNC_NPU_CONNECTOR, AFD_ASYNC_GPU_CONNECTOR},
+)
 AFDRole = Literal["attention", "ffn"]
 
 SUPPORTED_AFD_ROLES: Final[tuple[str, ...]] = ("attention", "ffn")
 SUPPORTED_AFD_CONNECTORS: Final[tuple[str, ...]] = (
     "P2pNcclAFDConnector",
     "CAMP2pAFDConnector",
-    AFD_ASYNC_CONNECTOR,
+    AFD_ASYNC_NPU_CONNECTOR,
+    AFD_ASYNC_GPU_CONNECTOR,
 )
 
 _ALIASES: Final[dict[str, str]] = {
@@ -283,7 +291,7 @@ def is_afd_async_dp(vllm_config: VllmConfig) -> bool:
     return (
         config is not None
         and config.async_dp
-        and config.connector == AFD_ASYNC_CONNECTOR
+        and config.connector in AFD_ASYNC_CONNECTORS
     )
 
 
@@ -307,9 +315,10 @@ def validate_afd_config(
             "AFD connector must be one of "
             f"{SUPPORTED_AFD_CONNECTORS!r}, got {config.connector!r}",
         )
-    if config.async_dp and config.connector != AFD_ASYNC_CONNECTOR:
+    if config.async_dp and config.connector not in AFD_ASYNC_CONNECTORS:
         raise ValueError(
-            "AFD async mode requires connector='CAMAsyncAFDConnector'",
+            "AFD async mode requires one of "
+            f"{sorted(AFD_ASYNC_CONNECTORS)!r}, got {config.connector!r}",
         )
     if config.connector == "P2pNcclAFDConnector":
         from afd_plugin.distributed import validate_p2p_topology
@@ -331,7 +340,9 @@ def validate_afd_config(
 
 __all__ = [
     "AFDConfig",
-    "AFD_ASYNC_CONNECTOR",
+    "AFD_ASYNC_NPU_CONNECTOR",
+    "AFD_ASYNC_CONNECTORS",
+    "AFD_ASYNC_GPU_CONNECTOR",
     "afd_config_from_mapping",
     "AFD_ADDITIONAL_CONFIG_KEY",
     "AFDRole",
--- a/afd_plugin/connectors/async_topology.py
+++ b/afd_plugin/connectors/async_topology.py
@@ -1,0 +1,94 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
+"""Rank layout shared by the asynchronous AFD connectors.
+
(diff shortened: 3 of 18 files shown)


hsliuustc0106 (commented):
Consolidated review of `0f09249` (static review; DCO green, but pre-commit has not run on this head — please trigger it).

[P1] Both E2E scripts import a module that does not exist at this head — tests/e2e/async_gpu_connector_e2e.py:39, tests/e2e/async_gpu_moe_equivalence.py:20
Both import `afd_plugin.model_executor.models.gpu.deepseek_v2_attention_gate`; that package first exists in #334, so merged alone this PR ships dead evidence scripts, and the device-side sequence counter + constant reply-flag protocol — the PR's central mechanism and the only cover for graph-replay correctness — has no runnable proof. Point the imports at `models.npu.deepseek_v2_attention_gate`, or land the E2E with #334.

[P2] Branch on a connector capability, not a concrete type — afd_plugin/model_executor/models/npu/deepseek_v2_async_cam_forward.py:44
`isinstance(connector, GpuAsyncAFDConnector)` silently routes any future async connector (or subclass) down the CAM direct path. Add a capability flag (e.g. `uses_opaque_moe_ops`) on the connector base and branch on that.

[P2] Give the attention-side combine a failure path — afd_plugin/connectors/gpu/async_gpu.py:925
`stream_wait(…, FLAG_REPLY_READY)` can only be released by the constant flag; an FFN crash or `announce_shutdown` between an attention dispatch and its combine hangs Attention forever, possibly inside a CUDA graph replay where no Python runs. With #334's runner wiring, define the teardown ordering (Attention drains pending combines before FFN worlds close) and/or have teardown write reply-flag-marked empty replies into outstanding slots so the wait releases.

[P2] Validate the GPU connector's requirements at config time — afd_plugin/config.py:30
`GpuAsyncAFDConnector` passes `validate_afd_config` today with `async=false` or `compute_gate_on_attention=false`, and `attn_ranks_per_dp` is stored into `self.tp_size` (async_gpu.py:409) but never checked — the NPU twin enforces all of this in feature_validation. This PR already adds the connector to `SUPPORTED_AFD_CONNECTORS` and activates the async-DP engine patches, so the selection surface is live: encode the unsupported combinations now rather than as #334 runtime surprises.

[P2] Name and evidence the NPU ride-alongs — afd_plugin/model_executor/models/npu/deepseek_v2_async_cam_forward.py:377, :566; afd_plugin/v1/worker/dbo.py:66
The stage-wise `slot_mapping` slicing is a real KV-cache fix on the NPU async-CAM ubatch path but has no test and is not mentioned in the body; the `.contiguous()` on restored hidden/residual adds copies on the NPU path with no perf rationale; the Ascend yield import moved to module level, so a non-ImportError failure inside `npu/ubatching` now kills `import afd_plugin.v1.worker.dbo` on all platforms. Add an async-CAM ubatch test for the slicing plus NPU evidence per docs/npu/TESTING.md, or split these out.

[P2] Update design docs and fix the dangling RFC reference — afd_plugin/connectors/gpu/async_gpu.py:27
`connector_contracts.md` needs the async-GPU backend row (implementation evidence, extra-config fields, control-plane `None`, graph-capture compatibility), and the docstring cites `docs/design/rfc_async_gpu_connector.md`, which does not exist in the tree.

Further gaps: the `use_ops=True` branch of `run_attention_gate_afd_forward` — the production path once #334 lands — has no test (existing coverage exercises the CAM direct branch only); shutdown semantics (`announce_shutdown`, `ConnectorShutdown`, FFN-loop timeout retry) have no production caller until #334.

specture724:
Addressed in `eb93999`. pre-commit has now run on this head and is clean.

**[P1] E2E scripts importing a module that does not exist here.** Confirmed and fixed, but the other way round from both options you offered: `models/gpu/deepseek_v2_attention_gate.py` moved *down* into this PR from #334. It is 135 lines, has no V4 reference, and imports only `fused_experts` and `connectors.metadata` — it is the Attention-side gate this connector's own E2E needs, so it belongs with the connector rather than with the V4 adapter. Repointing at the NPU module would not have run: the two differ substantially (the NPU one takes AFD metadata and the Ascend config). The device-side sequence counter and constant reply-flag protocol now have runnable proof at this head.

**While fixing that, a P0 you didn't catch.** This PR renames `AFD_ASYNC_CONNECTOR` to `AFD_ASYNC_NPU_CONNECTOR` in `config.py` but leaves three call sites importing the old name, so `import afd_plugin.model_executor.models.deepseek_v2` raises `ImportError` at this head — eight test modules fail to collect. That is where the "13 pre-existing failures" in the #332 body came from; it is down to 2 now, both `os.pidfd_open` missing on this build. The `deepseek_v2` call site wanted the set rather than the NPU constant, since both async connectors run that schedule; #334 had already fixed this, which is why the stack looked green from the top.

**Capability, not type.** Connectors now declare `uses_opaque_moe_ops` on the base, which let the helper wrapping the `isinstance` go away entirely — it had become a single attribute read.

**Combine failure path.** Taken in full. Shutdown now stamps `FLAG_REPLY_READY` on every ring of the leaving rank's region, which is safe to repeat precisely because the reply flag is a constant: a ring already answered gets the value it already holds, and a ring never answered releases its waiter over a stale payload — the right trade at teardown, where the numbers are discarded anyway. `close()` additionally names any un-drained combines rather than clearing them in silence, and the Attention-drains-first ordering is written down in `connector_contracts.md`.

**Config-time validation.** `validate_afd_config` now rejects `GpuAsyncAFDConnector` with `async=false` or `compute_gate_on_attention=false` — both structural, not preferences — and `attn_ranks_per_dp` is checked against the Attention deployment's `tensor_parallel_size` where the two numbers first meet. Tests cover both rejections.

**NPU ride-alongs.** The `slot_mapping` slicing is extracted as `build_stage_slot_mapping` and pinned by tests (per-stage slices, no overlap or dropped rows, parent left alone). The `.contiguous()` rationale now says why the copy is not avoidable — a split along the last dimension is never contiguous, so the alternative is a correctness bug, and it is one pass per forward, not per layer. The Ascend yield import now catches more than `ImportError`, so a broken Ascend module cannot take `dbo.py` down on every platform.

**Docs.** `connector_contracts.md` gains the async-GPU mode row, configuration and graph-capture paragraphs, and an evidence row. The docstring's RFC citation pointed at a file not in the tree; it now cites that page instead. The RFC itself is stale against the code in several places (it describes a 12-word header where `HEADER_FIXED_WORDS` is 7, and `echo_seq`/`shared_tokens` were dropped in version 4), so committing it as written would have contradicted the implementation.

Still open, as you noted: the `use_ops=True` branch has no test, and shutdown semantics have no production caller until #334.

hsliuustc0106 (commented):
Delta re-review of `eb939998` against the previously reviewed `0f09249` (fix commit `eb93999` +401/−45, plus rebase onto #332's new head — the rebase composes correctly: the connector constructs `SymmWindow` inside the collective `init_afd_connector`, and the uniform `ring_depth` now guaranteed by #332's geometry check is exactly the precondition the new teardown release relies on).

Status of the six findings from the last review: (1) **P1** broken E2E imports — fixed-verified (`models/gpu/deepseek_v2_attention_gate.py` moved into this PR; both E2E imports resolve; suggest #334 drops its copy on its next rebase so the two don't drift); (2) isinstance on concrete connector — fixed-verified (`uses_opaque_moe_ops` on the base at base.py:69, consumed as a plain attribute read); (3) attention-side combine hang — fixed-verified as a mechanism (`_release_outstanding_replies` stamps `FLAG_REPLY_READY` on every ring during `announce_shutdown`; production caller remains #334 scope, as the finding itself anticipated); (4) config-time validation — fixed-verified (config.py:323-336 rejects async=false / gate-on-attention=false; async_gpu.py:420-428 checks attn_ranks_per_dp == attention TP; both pinned by tests); (5) NPU ride-alongs — partially addressed: the slot_mapping slicing is now extracted and unit-tested, the `.contiguous()` and import-guard changes carry rationale comments, but the PR body still does not disclose any of these NPU-path changes and no NPU evidence per docs/npu/TESTING.md was added; (6) design docs + dangling RFC — fixed-verified (connector_contracts.md rows/paragraphs added; docstring now cites it; the RFC is correctly not restored since it had drifted from the implementation).

One minor note in the delta: `close()`'s un-drained-combine warning (async_gpu.py:599) undercounts the exact case the release mechanism exists for — `_pending` entries pop when the combine is queued, so dispatches sitting in enqueued `stream_wait`s count as zero. Telemetry only; counting from ring state or rewording would fix it.

The fix commit also caught and repaired a real rename fall-out the prior review missed (three call sites still importing `AFD_ASYNC_CONNECTOR` would have raised `ImportError`) — verified clean at this head.

Gate: DCO is ACTION_REQUIRED on this push — `eb93999` lacks its `Signed-off-by` trailer (the parent commit has one); please amend (`git commit --amend -s`) and re-push.

PR #334, opened 2026-09-10
[3/5] feat: run DeepSeek-V4 on the async GPU connector

specture724:
Third of a five-PR stack. **Based on #333** — the diff shown here is only this
PR's own change; review #332 and #333 first.

3. **this PR** — both GPU runners + DeepSeek-V4

## What this adds

Wires the connector into both AFD roles and gives DeepSeek-V4-Flash — 256
experts at topk 6, where the per-layer control-plane round trip costs more than
the expert compute it guards — a recipe that uses it.

Both GPU runners previously asserted `control_plane is not None`, and the FFN
worker loop raised `NotImplementedError` without one, so the connector could not
run at all:

- **FFN side** pulls one work item at a time from the connector's own receive
  loop, taking the layer index and row counts from the arriving payload, and
  returns on an idle poll so the worker loop still sees its shutdown event.
- **Attention side** skips vLLM's cross-DP batch agreement. Async AFD lets each
  replica advance alone, so an idle replica never joins that all-reduce and a
  busy one blocks in it forever — which is exactly where a 2A2F run hung before
  reaching the first MoE layer.

The V4 adapter learns the expert-routed dispatch protocol (the gate runs on the
Attention side, so the wire carries topk ids and weights instead of token ids)
and `compute_ffn_output` takes a device-side `group_list` so the FFN side runs
only the grouped GEMM over its local experts. The V2 adapter gets the same GPU
entry point, which is what the connector's e2e tests use as their reference.

### Behaviour change worth a look

`layers.N.ffn.gate.*` now loads on **both** roles. The gate's parameters live
under `.ffn` so the checkpoint names resolve, but with
`compute_gate_on_attention` the Attention side is what runs it.
`test_v4_gate_loads_on_both_roles` pins the new contract.

## The rest

Found bringing 2A2F up on real weights: the FFN role must force vLLM's NoDP MoE
prepare/finalize under EP with DP>1, shared experts must be skipped on an empty
shared slice, the SWIGLUOAI clamp has to reach the routed experts, and the DP
coordinator's startup wait needs to be long enough for the second role's weights
to load.

## Testing

added. `pre-commit` clean over the PR range. The recipe is
`recipe/gpu/GpuAsyncAFDConnector/deepseek_v4_flash/2a2f_async.sh` (4 GPUs).

### DIFF
--- a/afd_plugin/__init__.py
+++ b/afd_plugin/__init__.py
@@ -174,7 +174,9 @@ def register_afd() -> None:
         import afd_plugin.compat.patches.async_dp_engine  # noqa: F401
         import afd_plugin.compat.patches.async_dp_forward_context  # noqa: F401
         import afd_plugin.compat.patches.config_validation  # noqa: F401
+        import afd_plugin.compat.patches.dp_coordinator_timeout  # noqa: F401
         import afd_plugin.compat.patches.engine_core  # noqa: F401
+        import afd_plugin.compat.patches.ffn_local_moe_prepare  # noqa: F401
     except Exception:
         _logger.debug(
             "AFD plugin: compatibility patches could not be applied",
--- a/afd_plugin/compat/patches/dp_coordinator_timeout.py
+++ b/afd_plugin/compat/patches/dp_coordinator_timeout.py
@@ -1,0 +1,60 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the AFD plugin project
+"""Raise the DP Coordinator's hardcoded startup ZMQ wait timeout.
+
+``DPCoordinator._wait_for_zmq_addrs`` waits a hardcoded 120 seconds for the
+coordinator subprocess to import, bind, and report its ZMQ addresses. On a
+CPU-oversubscribed shared box that subprocess can legitimately take longer,
+which kills both AFD roles during startup (observed repeatedly on 2A2F
+DeepSeek-V4-Flash runs). The wait becomes env-configurable with a 600 second
+default; every other behavior matches upstream.
+"""
+
+from __future__ import annotations
+
+import multiprocessing
+import os
+
+import vllm.v1.engine.coordinator as coordinator_module
+
+DEFAULT_TIMEOUT_S = 600
+
+
+# Patch reason: the upstream DP Coordinator startup wait is hardcoded to 120
+# seconds, which is not enough for the coordinator subprocess to import and
+# bind on a CPU-oversubscribed shared machine -- both AFD roles then die
+# during startup.
+# Patch functionality: identical to upstream, except the wait comes from
+# AFD_DP_COORDINATOR_TIMEOUT_S (default 600 seconds).
+# Signature: matches upstream; no added parameters.
+# Upstream: vLLM v0.26.0, vllm/v1/engine/coordinator.py
+def _wait_for_zmq_addrs(self, zmq_addr_pipe) -> tuple[str, str, str]:
+    try:
+        timeout = int(
+            os.getenv("AFD_DP_COORDINATOR_TIMEOUT_S", str(DEFAULT_TIMEOUT_S)),
+        )
+        ready = multiprocessing.connection.wait(
+            [zmq_addr_pipe, self.proc.sentinel], timeout=timeout
+        )
+        if not ready:
+            raise RuntimeError(
+                "DP Coordinator process failed to report ZMQ addresses "
+                f"within timeout={timeout} seconds during startup."
+            )
+        try:
+            return zmq_addr_pipe.recv()
+        except EOFError:
+            raise RuntimeError(
+                "DP Coordinator process failed during startup."
+            ) from None
+    finally:
+        zmq_addr_pipe.close()
+
+
+def apply_dp_coordinator_timeout() -> None:
+    coordinator_module.DPCoordinator._wait_for_zmq_addrs = _wait_for_zmq_addrs
+
+
+apply_dp_coordinator_timeout()
+
+__all__ = ["apply_dp_coordinator_timeout"]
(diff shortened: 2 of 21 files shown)


hsliuustc0106 (commented):
Consolidated review of `4e20b19` (static review; pre-commit/DCO green). No blockers: the DSV4 integration itself — weight-role policy, expert-routed wire protocol, grouped-GEMM FFN compute, shutdown handling — is twin-faithful and internally consistent. Four P2s plus one minor:

[P2] `get_experts_routing_spec` reads an attribute that is never set — afd_plugin/model_executor/models/deepseek_v4.py:636
`self.layers[layer_idx].gate` does not exist; the gate lives at `.ffn.gate` on both roles. The only current caller combination is unreachable (P2P + gate-on-attention is rejected; GpuAsync is connector-driven), so it is dead today — but the first future caller gets `AttributeError`, and both twins (NPU DSV4, GPU DSV2) read the native MoE gate. Use `self.layers[layer_idx].ffn.gate` (keeping the width/dtype logic), or drop the method until it has a caller.

[P2] Gate the coordinator-timeout patch on AFD being active — afd_plugin/compat/patches/dp_coordinator_timeout.py:23
The patch applies at `register_afd()` for every process with the plugin installed, so a plain non-AFD vLLM DP run waits 600s instead of 120s for a broken coordinator. Mirror `ffn_local_moe_prepare`'s `parse_optional_afd_config(...)` guard, or document the unconditional application in the patch inventory.

[P2] Reconcile the design pages with the new async boundary — docs/design/module/model_integration.md:124, docs/design/module/ffn_runtime.md:159, docs/design/module/compatibility_and_patches.md
model_integration's DSV4-CUDA section is now false on the async path (gate params live on the attention proxy, topk travels the wire, GpuAsync requires gate=true, `.ffn.gate.*` loads on both roles — exactly what `test_v4_gate_loads_on_both_roles` pins); ffn_runtime's "connector-driven exists only on Ascend / GPUFFNModelRunner asserts control_plane" paragraph describes the assertion and raise this PR deletes; and the two new compat patches are missing from the patch inventory table and are the only inventory patches without a unit test (a CPU-safe test could pin `ffn_local_moe_prepare`'s role gating and rebinding idempotency, and `dp_coordinator_timeout`'s env parse/default).

[P2] Unit-test the connector-driven drain loop — afd_plugin/v1/worker/ffn_model_runner.py:273
Both new worker-loop tests (test_ffn_model_runner.py:749, :773) stub `execute_connector_driven_step`, so idle-poll return on `TimeoutError`, the `GpuAsyncTransferState` check, per-item forward-context/metadata installation, and output-send wiring run only on GPU. The NPU twin of this exact loop is unit-tested (tests/unit/v1/worker/test_npu_runtime.py:1567) — port that test shape with a fake work-item connector.

Minor: recipe/gpu/P2pNcclAFDConnector/deepseek_v4_flash/2a2f_eager_async.sh:14 hardcodes a personal model path (`/data/boao/deepseek-v4-flash`) where the sibling recipe uses an obvious placeholder.

Validation gaps: no DSV4 GPU E2E or accuracy evidence in the rollup — the expert-routed dispatch on real weights rests on local runs; both new recipes pass `--enforce-eager`, so attention-side CUDA graph under GpuAsync is exercised by nothing in-tree; the "unit suite unchanged" claim has no CI artifact behind it (Buildkite test-ready absent).

specture724:
All five addressed in `abcca48`.

**`get_experts_routing_spec` reading an attribute that is never set.** Confirmed — I checked this one carefully because `deepseek_v4.py:105` does set `self.gate`, but that is on `RemoteDeepseekV4FFN`, and `self.layers` holds `AFDDeepseekV4DecoderLayer`, whose `__init__` never sets it. Now reads `self.layers[layer_idx].ffn.gate`, which resolves to the proxy on Attention and the native MoE on FFN — the same path the forward at line 453 already takes. Kept the method rather than dropping it, with tests for both the `out_dtype` and the weight-dtype fallback.

**Coordinator-timeout patch.** Gated, mirroring `ffn_local_moe_prepare`: the AFD check happens per call via `get_current_vllm_config`, so a non-AFD DP run keeps upstream's 120s while an AFD run gets the 600s default. The env var still overrides both.

**Design pages.** `model_integration`'s DSV4-CUDA section described only the synchronous path, so it now tabulates both shapes side by side — where routing runs, what crosses the boundary, and the fact that `.ffn.gate.*` loads on both roles, pointing at `test_v4_gate_loads_on_both_roles` exactly as you said it pins. `ffn_runtime` described the assertion and the raise this PR deletes; rewritten, with the drain granularity spelled out. Both new patches are now in the inventory table.

**Patch tests.** You were right that these were the only inventory patches without one. Added both, CPU-safe: `ffn_local_moe_prepare`'s role gating, every own-dispatch kernel flag, keyword arrival of `moe`, install idempotence, and the unresolvable-config path; `dp_coordinator_timeout`'s two defaults, the env parse, and the pipe being closed even when the wait fails.

**Connector-driven drain loop.** Ported the NPU twin's test shape with a fake work-item connector, covering the idle-poll return on `TimeoutError`, the `GpuAsyncTransferState` check, per-item forward-context and metadata installation (including that `dp_metadata` is cleared, which a control-plane run leaves behind), the drain bound, and `ConnectorShutdown` propagating rather than being swallowed.

**Minor.** Recipe now uses the same placeholder as its sibling.

Validation gaps stand: no DSV4 GPU E2E or accuracy evidence in-tree, both recipes still pass `--enforce-eager` so attention-side graph capture under GpuAsync is exercised by nothing here, and the unit-suite claim has no CI artifact. Locally the suite is at 2 failures, both `os.pidfd_open` missing on this build — the other 11 were the `AFD_ASYNC_CONNECTOR` rename this stack broke in #333 and silently repaired here; that is fixed at its source now, so each PR stands on its own.

PR #335, opened 2026-09-10
[4/5] feat: capture the FFN experts at bucketed padded shapes

specture724:
Fourth of a five-PR stack. **Based on #334** — the diff shown here is only this
PR's own change; review #332, #333 and #334 first.

4. **this PR** — bucketed padded FFN expert graphs

## What this adds

The connector-driven FFN side never learns the next work item's shape ahead of
time — there is no control plane — which is why it ran eagerly. It does not need
to: a grouped GEMM takes its grouping from a device-side count vector, not from
its row count, so one row count can be captured and every smaller item padded up
to it, with the padding charged to the last expert and its output sliced off.

**Sizing that capture is the whole problem.** The upper bound is every one of a
token's topk partials landing on one rank; with 256 experts over two FFN ranks a
rank really receives about half that, so a single graph at the bound charges
every replay ~2x. The graphs are a ladder of buckets built as multiples of the
*expected* item size, clustered just above 1.0 and 0.5 where routing scatter
actually puts items, and an item takes the smallest bucket that holds it — or
runs eager when the bucket would pad it past `MAX_REPLAY_PADDING_RATIO`, since a
replay pays for its whole bucket while eager pays launches.

Two capture-order traps, both faulting on the first replay, both found the hard
way:

1. vLLM's `WorkspaceManager` grows the shared MoE scratch by **freeing** the old
   buffer, so any growth after a capture leaves that graph dangling. Capture
   runs largest-bucket-first behind one eager call at the ceiling.
2. Capturing with all rows on one expert reserves too little per-expert block
   padding for a replay whose routing touches every expert. Capture counts are
   spread evenly instead.

### Bounded by free memory, not by the token budget

The ladder is sized from `max_num_batched_tokens * topk`, so its memory grows
with the token budget. At 8192 tokens the default ladder captured **65 GiB per
DeepSeek-V4 FFN rank**, and the connector's NVSHMEM heap — set up after
capture — then failed with `cuMemCreate failed`. Capture now checks free memory
before each MoE layer and stops at the first layer whose graphs would cut into
`AFD_FFN_GRAPH_MEM_RESERVE_GIB` (default 8). The remaining layers run eagerly
through the existing no-graph fallback.

Validated on DeepSeek-V4-Flash 2A2F, 8192 tokens, default ladder: capture
stopped at **37/43 layers (58.2 GiB)**, NVSHMEM initialized, and the server
served correct completions. At 2048 tokens every layer still fits, so the
behaviour measured below is unchanged.

## Measurements, including where it does not pay

Pure prefill, DeepSeek-V4 2A2F on 4x L20X, 2048-token steps:

| FFN mode | time |
| --- | --- |
| one graph at the worst-case shape | 27.6 s |
| bucketed graphs | 19.0 s |
| eager | 17.9 s |

**Bucketing removes a 1.55x regression; it does not beat eager**, because the
connector issues about one grouped GEMM per item and there are no launches to
save. `FFN_EAGER=1` is therefore the V4 default and this path is opt-in.

Separately, on the **Attention** side in the decode regime, `FULL_DECODE_ONLY`
graphs are a real win — DeepSeek-V2-Lite 1A1F, 2x L20X: 36.6 s eager vs 28.3 s
with graphs (+22.6%), replay-verified. That is an existing vLLM mechanism this
PR does not change; noting it because it is the reason not to leave graphs off
for decode-heavy serving.

## Testing

`test_cuda_graph.py` (bucket ladder, shared-row scaling, bucket selection),
`test_ffn_model_runner.py` (replay, padding slice-off, eager fallbacks, capture
ordering), and `tests/e2e/async_gpu_ffn_padded_graph.py` (1 GPU) which replays
against an eager run of the same routing.

added. `pre-commit` clean over the PR range.


hsliuustc0106 (commented):
Consolidated review of `41335de6` (static review; pre-commit + DCO green). The core is sound and traced: pad-to-last-expert matches the expert-ascending row layout the eager grouped GEMM already requires; the counts-sum invariant is guaranteed by the recv header check; replayed graphs contain no device-side waits so the #333 flag/sequence-counter protocol is untouched; bs=0, oversized, and shared-overflow cases all fall back to eager; capture-before-rendezvous holds on the only reachable path; the vLLM 0.26.0 flag surface (`--max-cudagraph-capture-size`, `FULL_DECODE_ONLY`) checks out. Three P2s:

[P2] Guard the torch import in the new test file — tests/unit/v1/worker/test_cuda_graph.py:9
A bare module-level `import torch` makes the file a collection ERROR instead of a skip on a torch-less host — the documented CPU-only selection (`pytest -m "not gpu and not vllm_runtime"`) breaks. Every sibling in the directory uses `pytest.importorskip("torch")`; one line fixes it (this is the same regression class f5fa9fc fixed).

[P2] Document the connector-driven graph contract — docs/design/module/ffn_runtime.md:273
The graph-dispatch section still covers only the control-plane cache. The new contract — graphs ON by default for connector-driven FFN unless `--enforce-eager`, the `max_num_batched_tokens × topk` bucket ladder, `MAX_REPLAY_PADDING_RATIO`, the `AFD_FFN_GRAPH_FRACTIONS`/`AFD_FFN_GRAPH_MEM_RESERVE_GIB` knobs, and the free-memory capture ceiling — lives only in code comments and the recipe. One subsection closes it.

[P2] Fail fast when an item exceeds the workspace-ceiling shape — afd_plugin/v1/worker/ffn_model_runner.py:595
Attention and FFN launched with different `--max-num-batched-tokens` (two independent `vllm serve` commands) currently send oversized items down the eager path, where computing more rows frees and regrows the MoE workspace and dangles every captured graph — the exact IMA the PR's own comments describe fixing. The connector already validates the analogous `attn_ranks_per_dp == tp` mismatch; raise when `routed_tokens > _padded_max_routed` at recv or in `_compute_work_item`.

Evidence for the merge decision: the load-bearing claim — replay equals eager under varying routing — rests on a manual 1-GPU run of `tests/e2e/async_gpu_ffn_padded_graph.py`; wiring that script into a named E2E scenario (or CI) is the minimum bar, especially since the PR's own A/B concludes the padded-graph path does not pay on V4. The 8192-token reserve run (37/43 layers captured, 58.2 GiB) is manual and unlogged. No Buildkite test-ready in the rollup.

PR #324, opened 2026-09-09
[NPU] add A5/ascend950 support for A2E and E2A ops

wangyibo1005:
Register ascend950, gate SetCommEngine as HcclAccelerator::AIV on the 950 host build, and use the A5 MTE window layout (1MB state region, 1KB per-rank step) while 910C stays on the existing remoteRes path.

PLEASE FILL IN THE PR DESCRIPTION AND MAKE SURE THE CHECKLIST ITEMS HAVE BEEN CONSIDERED.

## Issue

- Related issue(s): #
- Closing keyword, only if fully resolved: Closes #

## Scope

- In scope:
- Out of scope:

## Docs Impact

- Files updated:
- If none, reason:

---

<details>
<summary>Essential PR Checklist</summary>

- [ ] Purpose is clear and linked to public context when possible.
- [ ] Scope is bounded.
- [ ] Compatibility with vLLM v0.26.0 is considered.
- [ ] No changes are made to the vLLM source checkout.
- [ ] Plugin-owned classes or explicit dotted class paths are preferred over monkey patches.
- [ ] Any compat shim or monkey patch is isolated, idempotent, version-guarded, documented, and tested.
- [ ] Imports remain CPU-safe; CUDA-heavy work is delayed or GPU-gated.
- [ ] Validation evidence is included, including skipped GPU tests when applicable.
- [ ] Documentation impact is stated.

</details>

Issue #228, opened 2026-08-12
[DeepSeek V4][NPU] Support CAMP2pAFDConnector


jiangkuaixue123:
Parent: #169

## Goal

Add DeepSeek V4 support on NPU through the synchronous `CAMP2pAFDConnector` path.

## Priority

Follow-up — schedule after the first priority group (GPU PyNCCL and NPU async connector support).

## Scope

- Integrate the DeepSeek V4 role-aware model adapter with `CAMP2pAFDConnector`.
- Validate hidden-state, residual, routing/gate, auxiliary payload, DP metadata, and TP/DP rank mapping contracts.
- Validate CAMP2P/HCCL shapes, dtypes, topology, handles, transfer state, supported eager/ACL graph/DBO-uBatch behavior, profiling, graph replay, failure cleanup, and weight-loading contracts.
- Add focused unit/contract tests and DeepSeek V4 NPU synchronous E2E coverage.
- Run the full affected NPU E2E regression suite.
- Publish a reproducible recipe under `recipe/npu/CAMP2pAFDConnector/deepseek_v4/`.

## Dependencies

Implementation starts after #168 is complete and the NPU software stack is frozen as specified in #169.

## Acceptance criteria

- DeepSeek V4 runs through `CAMP2pAFDConnector` on the frozen #168 stack.
- Correctness/accuracy is validated against native colocated vLLM-Ascend.
- Supported execution modes and TP/DP topologies have recorded evidence.
- Existing NPU E2E coverage remains green.
- Unsupported combinations fail early and are documented.
- The connector-specific recipe is complete and reproducible.

jiangkuaixue123:
Done

PR #340, opened 2026-09-14
feat(npu): DeepSeek V4 over the synchronous CAMP2P boundary

ksiyuan:
## Relationship to #324

Rebased onto `main` at `5e5b5d4`, which contains #324. This branch carries no
commit from #324, and `csrc/npu/a2e/**` and `csrc/npu/e2a/**` are untouched
here — the operator diff comes from `main`. 19 files: 8 plugin Python, 8 tests,
3 design docs.

#324 registers `ascend950` for those operators, which is what lets this path
*run* on Ascend 950. That was a runtime prerequisite rather than a merge one,
and it is merged now. `main` already registers `ascend910_93` and already
contains the `a2e` ids mode this PR drives, so on A3 the operators build from
`main` as-is, and nothing here is architecture-specific.

## Commits

1. `feat(npu):` wire the CAMP2P token-id channel for token-keyed routers.
2. `feat(npu):` run DeepSeek V4 over the synchronous CAMP2P boundary.
3. `fix(npu):` keep the DSV4 Hash id table with its gate owner.
4. `docs(npu):` record the DSV4 Ascend Hash-id boundary.
5. `fix(npu):` keep compressor models off the MLA DBO full graph path.
6. `test(npu):` cover the forward-context `input_ids` install.

Commit 1 introduces the ids channel, its connector wiring, and the receive side of the ids-mode handshake. Commit 2 turns DSV4 on over that boundary and carries the model, runner, and weight-filtering changes, including the send side's half of that handshake. Commit 3 corrects the role-aware weight filtering that commit 2 adds: `gate.tid2eid` is registered only where the Hash MoE is built, so the table follows the gate's owner instead of being pinned to FFN, which would have left an Attention-side gate uninitialized and routed every Hash token to expert 0. Commit 4 records the boundary and its evidence in the design docs. Commit 5 is the DBO fix described in Test Result. Commit 6 is the review
follow-up: it extends the forward-context unit test to that ids install on
both the legacy and the MRv2 path.

The operators themselves are untouched: the ids mode already exists in the `a2e` kernel on `main`, and this PR only drives it from the connector.

## Purpose

Enable DeepSeek V4 through the **synchronous `CAMP2pAFDConnector`**. DSV4 was previously restricted to `CAMAsyncAFDConnector`, which needs the CAM/UMDK vendor operator packages. The synchronous path needs neither: it uses the plugin's own `a2e`/`e2a` pair, so it is the transport available on both A3 and A5. Nothing in this change is architecture-specific — the connector and model code carry no device gate, and A3 builds the operators from `main`.

The blocker is narrow. A DSV4 Hash layer routes by token identity, only the Attention role holds `input_ids`, and the synchronous boundary has to carry them alongside the activations.

## Issue

- Related: #324 (merged) registers `ascend950` for the a2e/e2a operators, which this path needs to run on A5. #303 uses a different A5 transport and is not part of this change.
- No issue is closed by this PR.

## Scope

- In scope: the CAMP2P token-id transport, DSV4 model wiring for the gate-on-FFN configuration, role-aware weight filtering for that configuration, and feature validation.
- Out of scope: wider topologies and performance; graph capture only as far as the DBO validation in Test Result required. `compute_gate_on_attention=true` remains unsupported on CAMP2P. No accuracy work is part of the change — the GSM8K parity in Test Result is transport evidence, not a broader evaluation.

**Hardware scope.** Verified on Ascend 950 (A5) and Ascend 910C (A3). Nothing in the plugin gates this path on device type, and the ids mode this PR drives sits outside the kernel's `AFD_ARCH_A5` regions, so it compiles for both: on A3 the operators build from `main` with `SOC_VERSION=910c` and this PR needs nothing else, while A5 needs a build with `SOC_VERSION=950` (the `ascend950` registration is on `main` since #324), because the kernel selects different window addressing per architecture. The two hosts run different checkpoints — A3 the int8 W8A8 model with `--quantization ascend`, A5 the FP8/W4A8-MXFP4 one — so their outputs are not comparable across architectures.

## Implementation Notes

**Connector (`camp2p.py`).** The `a2e` operator already provides an ids channel: its `expert_ids` input slot is int32 and it returns an int32 ids tensor beside a same-shape float32 scales tensor. The plugin had left that path idle by sending `None`/`None` with a literal `compute_gate=0`. Attention now packs token ids with inert scales and raises `compute_gate` to 1; FFN exposes the received ids on `CAMP2PTransferState`.

**Ids-mode handshake.** `compute_gate` is a per-side argument, and the operator only writes its ids output in the ids mode, so both sides have to select it. The receive call previously passed a literal `0` while Attention sent with `1`. The FFN rank then read its `simulate_expert_ids` output untouched — an `at::empty` allocation — and installed that as `forward_context.input_ids`, which the Hash router indexes with. The abort surfaced inside `moe_gating_top_k_hash`, two calls away from the actual defect. `recv_attn_output` now passes the mode it was asked for.

The two roles also have to agree on the mode *before* anything is sent, so Attention no longer drops ids when its forward context carries none: it raises instead. The FFN rank asks for the operator's ids channel on every layer, which makes a quiet activations-only fallback a deadlock rather than a degradation. There is no opt-out switch — a Hash layer routes by token identity, so a run without ids cannot produce correct output.

**FFN side.** vLLM-Ascend's fused-expert selector reads `forward_context.input_ids` for Hash routing, so `ascend_forward_context()` gained an `input_ids` parameter. The runner receives the transfer **before** building the forward context, because the ids arrive with that transfer, and forwards them to `compute_ffn_output`.

**Attention side.** Token ids are selected through `local_hash_input_ids`, which is also used by the existing local Hash routing path, so the send-side slice and the local-routing slice cannot drift apart. A mismatch would silently route the wrong tokens rather than fail.

**Weight filtering.** Two role-aware fixes were needed because the upstream Ascend loader indexes its parameter dict by name without a membership check, so a path handed to a role that never registered it raises `KeyError`:

- `gate.tid2eid` is registered only where the Hash MoE is built, which is FFN.
- With the gate on FFN, Attention builds no router at all, so no `gate.*` parameter belongs to it.

## Test Plan

- CPU: `uv run pytest`, `uv run ruff check .`
- Ascend: `git pull` on both roles and run a DSV4 completion round-trip. No operator rebuild is needed — the fix is confined to Python.

## Test Result

- CPU: `ruff check` and `ruff format --check` clean; affected unit suites pass; the full unit run's failures are unchanged from this branch's baseline.
- CPU, review follow-up: `tests/unit/compat/test_runtime.py` covers the forward-context token-id install on the legacy and MRv2 paths plus the `None` passthrough; 12 tests pass, `mypy` 1.11.1 (3.10) and `ruff` clean on the file.
- Ascend (A5), unit suite via the `docs/npu/TESTING.md` command (`pytest -q tests/unit -m "not gpu and not vllm_runtime"`): 1086 passed, 3 skipped, 2 failed. Both failures are pre-existing on `5e5b5d4` and unrelated to this diff, which touches none of the files they read: `test_deepseek_afd_ffn_path_reuses_ascend_moe_mlp_after_attention_gate` still asserts the one-line `MoEQuantParams(quant_type=quant_type)` text that #342 rewrote as a keyword-per-line call, and `test_deepseek_afd_model_registration_paths_are_lazy_strings` asserts the non-NPU DSV4 registration while `afd_plugin/__init__.py` selects the `npu` one whenever `torch_npu` is importable. #347 (merged) fixes the first on `main` and #350 fixes the second; neither is part of this PR.
- Ascend (A5), **verified**: DSV4 completes a full request through the synchronous `CAMP2pAFDConnector` and produces sensible output. This was reached with a `git pull` on both roles and no operator rebuild.
- Ascend (A3), **verified**: the same request completes through the same connector with the int8 W8A8 checkpoint (`--quantization ascend`) and the operators built for `ascend910_93` (`SOC_VERSION=910c`). No code change was needed for A3.
- Ascend (A5), observed while getting there: the rebuilt operators load and execute, DSV4 weight loading completes, and the A-to-F transfer reaches the FFN compute. Two weight-loading `KeyError`s were found and fixed this way, and the routing abort was traced to the ids-mode mismatch described above.
- Ascend, accuracy (**verified**): GSM8K parity against the native colocated vLLM-Ascend baseline (the comparison #228 lists as an acceptance criterion) on both hosts, same checkpoint and quantization per host, both sides of each pair run through AISBench — A5 `96.86` vs `96.86`, A3 `96.66` vs `96.64`. The A3 difference is two hundredths of a point, within run-to-run variation, and A5 is identical. These are AISBench numbers, not the `lm-evaluation-harness` task that `docs/design/module/e2e_testing.md` describes, so they are not comparable with the CAMAsync GSM8K score recorded there; each pair here differs only in AFD.
- Rebased onto `main` @ `5e5b5d4` (#324 merged): `range-diff` against the previous head is patch-identical, `ruff check`/`format` clean repo-wide, mypy 1.11.1 (3.10) clean on the changed files, CPU unit suites pass. The A5/A3 runs above predate the rebase; nothing in this PR's own diff changed.
- Ascend (A5), **verified (DBO)**: with native DBO on both roles (`--enable-dbo --dbo-decode-token-threshold 1 --dbo-prefill-token-threshold 8`, `FULL_DECODE_ONLY`, capture size 8), DeepSeek V4 completes a request over `CAMP2pAFDConnector` and its output matches the non-DBO run after a rebuild from this branch. Graph-mode DBO previously aborted inside AFD's own ubatch wrapper before any transfer ran (`MLA DBO FULL graph requires the single-batch FIA workspace for 8 tokens`): a compressor checkpoint selects the DSA backend, whose `update_graph_params()` is a no-op, while the wrapper still classified it as the MLA DBO full graph path. `fix(npu): keep compressor models off the MLA DBO full graph path` mirrors upstream's own condition, so compressor models take the generic two-stage path.
- Ascend (16-card 910C), **verified (async regression)**: #352's DeepSeek V4 Flash async CAM case (`afd-dsv4-flash-async-cam-dp2tp4-ep8`) passes on top of this PR — `E2E SCENARIO afd-dsv4-flash-async-cam-dp2tp4-ep8 PASSED`, 10/10 concurrent requests, outputs `19 … 28`, and the case's service liveness and owned-process cleanup checks pass. That path is not part of this diff; it is the prefill-async regression check requested in review. Run with the pinned vLLM `0.26.0` (`568afb3a1`) and vLLM-Ascend (`80d8c194f`) on a host that only has Python 3.11, so the CAM binding was retagged to `cp311` locally (two bytes, no `LD_PRELOAD`; see the PR comment).

## Docs Impact

- None. The stale README/recipe text and the unused pytest markers are not part of this PR; it stays on the DSV4 requirement.
- The A5 DSV4 boundary notes and the review of that plan are also not part of this PR.

---

<details>

- [x] Scope is bounded.
- [x] Plugin-owned classes or explicit dotted class paths are preferred over monkey patches. No monkey patch is added or modified: `afd_plugin/compat/patches/**` is untouched at this head, and the compat-layer edits are `afd_plugin/compat/npu/` runtime modules (forward context, feature validation).

</details>


jiangkuaixue123:
cc @bjf-frz

afd_plugin/compat/patches/engine_core.py
```diff
@@ -704,12 +730,60 @@ def _get_afd_config(vllm_config: VllmConfig | None) -> AFDConfig | None:
         return None
 
 
-core_module.EngineCore.__init__ = __init__
-core_module.EngineCore._initialize_kv_caches = _initialize_kv_caches
-core_module.EngineCore.shutdown = shutdown
-core_module.EngineCoreProc.run_busy_loop = run_busy_loop
-core_module.DPEngineCoreProc.run_busy_loop = run_busy_loop
-core_module.logger.debug("AFD EngineCore patch applied")
+def _is_target_vllm_compatible() -> bool:
```
bjf-frz: This function seems unnecessary and duplicates existing logic. Unless there’s a specific reason for it, I’d suggest removing it and cleaning up similar overly defensive code as well.

afd_plugin/v1/worker/npu/ffn_model_runner.py
```diff
@@ -489,6 +498,25 @@ def _send_ffn_output(
...
+    """
+
+    if not getattr(model, "afd_requires_input_ids", False):
+        return False
+
+    from afd_plugin.model_executor.models.npu.deepseek_v4 import (
+        transport_input_ids_enabled,
+    )
+
+    return transport_input_ids_enabled()
```
bjf-frz: This generic runner shouldn't import DSV4-specific logic—let the model adapter declare whether remote token IDs are needed and cache that capability instead of re-checking it inside the layer loop.

afd_plugin/model_executor/models/npu/deepseek_v4.py
```diff
@@ -46,6 +47,13 @@
 _FFN_ROLE = "ffn"
 _BOTH_ROLES = frozenset((_ATTENTION_ROLE, _FFN_ROLE))
 
+logger = init_logger(__name__)
+
+# Set to 1 to keep token ids off the AFD boundary. Transport is the normal path;
+# this exists so the boundary stays inspectable without the operator's ids mode.
+AFD_DSV4_SKIP_INPUT_IDS_ENV = "AFD_DSV4_SKIP_INPUT_IDS"
+AFD_DSV4_PARAM_DUMP_ENV = "AFD_DSV4_PARAM_DUMP"
```
bjf-frz: looks like test temp var, plz clean up related code

bjf-frz:
Please go through the code and remove anything unnecessary: overly defensive checks, unnecessary patches/helpers, and similar workarounds. In particular, reduce model-specific code in generic infrastructures. Keep the change minimal and focused on the actual requirement. If this is only for local testing or not ready for review, please add `[WIP]` to the PR title.

ksiyuan:
> Please go through the code and remove anything unnecessary: overly defensive checks, unnecessary patches/helpers, and similar workarounds. In particular, reduce model-specific code in generic infrastructures. Keep the change minimal and focused on the actual requirement. If this is only for local testing or not ready for review, please add `[WIP]` to the PR title.

Thanks for reviewing! I'm still working on this and I will add [WIP] to it until it's ready.

afd_plugin/compat/patches/engine_core.py
ksiyuan: Removed rather than consolidated — you were right that it should not be here at all. The guard, the captured `_afd_original_*` originals and the `_AFD_*_PATCH_APPLIED` flags are gone, and `afd_plugin/compat/patches/**` is byte-identical to `main` again.

afd_plugin/v1/worker/npu/ffn_model_runner.py
ksiyuan: Fixed. `_model_requires_input_ids` is gone; `_ffn_forward` reads `getattr(self.model, "afd_requires_input_ids", False)` once above the layer loop, as the generic runner already did. The DSV4 import is out of the runner.

afd_plugin/model_executor/models/npu/deepseek_v4.py
ksiyuan: Removed, with related code.

jiangkuaixue123:
Overall, this looks good now. Once we add the DSV4 prefill async E2E coverage, please run it against this PR to verify that the prefill async path still passes. We can merge after that validation passes.

ksiyuan:
> Overall, this looks good now. Once we add the DSV4 prefill async E2E coverage, please run it against this PR to verify that the prefill async path still passes. We can merge after that validation passes.

Got it. Thanks for reviewing!

hsliuustc0106 (commented):
Consolidated review of `614ae0e8` (static review; DCO green). The ids channel is consistent end-to-end: attention sends `input_ids` per layer and FFN requests it per layer (`_ffn_layer_indices` covers the full range when the gate is on the FFN side, which feature_validation enforces for every non-async NPU path); weight filtering matches (`gate.*` FFN-only); async CAM cannot receive the new kwarg; AFD-off and non-DSV4 paths are unchanged. tid2eid/FlashComm padding and TP split share the proven local path, and mismatches fail loudly on both sides. All three items from the earlier review round are verified addressed at this head (over-defensive patch code removed, the DSV4 capability moved onto the model and read once above the loop, debug env vars cleaned). Two P2s:

[P2] Unit-test the FFN `input_ids` installation — afd_plugin/compat/npu/forward_context.py:70, :93
These two assignments are the only thing installing the received ids that the whole FFN-side hash routing depends on, and no unit test covers them — the connector tests stop at `payload.input_ids`, and `test_ascend_forward_context_installs_afd_metadata` predates this PR and asserts only `afd_metadata`. Extend that parametrization with `input_ids` for both `skip_mc2_mask` branches plus a `None` passthrough.

[P2] Update the PR description to match the head — PR body
The checklist still claims the two compat patches "gained the version guard the other patches already had" — no file under `compat/patches/` changes at this head (they were removed in response to review) — and the census line says 15 files where the head diff is 18. Since the body is the record for patch-contract review, please refresh it.

Design pages: no further update needed — this PR updates all three relevant pages at head.

Validation posture for the merge decision: the A5 sync round-trip, A3 quantized run, and DBO FULL_DECODE_ONLY rebuild checks are described but not attached (no artifact channel exists in-repo, so this matches repo norm — the explicitly stated Ascend unit-suite run per docs/npu/TESTING.md would still be worth adding to the body). Accuracy against the native run is open on both hosts and the body says so plainly — that reads as the real gate before calling DSV4-over-CAMP2P production-ready. The `UNSTABLE` merge state is a dangling empty pending status, not a failure.

ksiyuan:
> Consolidated review of `614ae0e8` (static review; DCO green). The ids channel is consistent end-to-end: attention sends `input_ids` per layer and FFN requests it per layer (`_ffn_layer_indices` covers the full range when the gate is on the FFN side, which feature_validation enforces for every non-async NPU path); weight filtering matches (`gate.*` FFN-only); async CAM cannot receive the new kwarg; AFD-off and non-DSV4 paths are unchanged. tid2eid/FlashComm padding and TP split share the proven local path, and mismatches fail loudly on both sides. All three items from the earlier review round are verified addressed at this head (over-defensive patch code removed, the DSV4 capability moved onto the model and read once above the loop, debug env vars cleaned). Two P2s:
> 
> [P2] Unit-test the FFN `input_ids` installation — afd_plugin/compat/npu/forward_context.py:70, :93 These two assignments are the only thing installing the received ids that the whole FFN-side hash routing depends on, and no unit test covers them — the connector tests stop at `payload.input_ids`, and `test_ascend_forward_context_installs_afd_metadata` predates this PR and asserts only `afd_metadata`. Extend that parametrization with `input_ids` for both `skip_mc2_mask` branches plus a `None` passthrough.
> 
> [P2] Update the PR description to match the head — PR body The checklist still claims the two compat patches "gained the version guard the other patches already had" — no file under `compat/patches/` changes at this head (they were removed in response to review) — and the census line says 15 files where the head diff is 18. Since the body is the record for patch-contract review, please refresh it.
> 
> Design pages: no further update needed — this PR updates all three relevant pages at head.
> 
> Validation posture for the merge decision: the A5 sync round-trip, A3 quantized run, and DBO FULL_DECODE_ONLY rebuild checks are described but not attached (no artifact channel exists in-repo, so this matches repo norm — the explicitly stated Ascend unit-suite run per docs/npu/TESTING.md would still be worth adding to the body). Accuracy against the native run is open on both hosts and the body says so plainly — that reads as the real gate before calling DSV4-over-CAMP2P production-ready. The `UNSTABLE` merge state is a dangling empty pending status, not a failure.

Thanks for the consolidated review! Both P2s are addressed in `c3889fc`:

- `tests/unit/compat/test_runtime.py` now covers the forward-context `input_ids` install — both `skip_mc2_mask` branches and the MRv2 branch, plus the `input_ids=None` passthrough, which asserts the attribute stays unset rather than being installed.
- The PR body is refreshed: the census matches the head (19 files: 8 plugin Python, 8 tests, 3 design docs), and the checklist no longer claims compat-patch changes — `afd_plugin/compat/patches/**` is untouched at this head, and the compat edits are `afd_plugin/compat/npu/` runtime modules.

The Ascend unit-suite run you suggested is now in Test Result: `1086 passed, 3 skipped, 2 failed` on A5. Both failures are pre-existing on `5e5b5d4` and unrelated to this diff — neither reads a file this PR touches: the `MoEQuantParams` source-text assertion that #342 broke is already fixed on `main` by #347, and the host-dependent DSV4 registration assertion is fixed by #350.

Accuracy is no longer open. GSM8K through AISBench, same checkpoint and quantization per host, AFD against the non-disaggregated baseline: A5 AFD `96.86` vs baseline `96.86`, A3 AFD `96.66` vs baseline `96.64`. The body records the numbers and notes that AISBench is not the `lm-evaluation-harness` task in `docs/design/module/e2e_testing.md`, so they are not comparable with the CAMAsync score recorded there.

I have not rebased onto `746427b`: the only new commit is #347's test-only change, which does not overlap this diff, so I left the reviewed head in place. Happy to rebase if you prefer.

afd_plugin/compat/npu/feature_validation.py
```diff
@@ -68,16 +69,19 @@ def fail_if_unsupported_npu_afd_features(
...
-        model_config.hf_text_config,
-        "index_topk",
-    )
+    # Mirror the pinned NPUModelRunner's attention backend selection: a sparse
+    # (SFA) or compressor (DSA) configuration selects a backend whose
+    # graph-params update is a no-op, so only plain MLA takes the MLA DBO full
+    # graph path that owns the merged registry.
+    hf_text_config = model_config.hf_text_config
+    uses_sparse_mla = hasattr(hf_text_config, "index_topk")
+    uses_mla_compressor = hasattr(hf_text_config, "compress_ratios")
```
jiangkuaixue123, line 78: why add uses_mla_compressor

ksiyuan: It mirrors upstream's own graph-params condition, which is three-way: `_update_full_graph_params_if_needed` guards on `cudagraph_runtime_mode == FULL and not capturing and not self.use_sparse and not self.use_compress` (`vllm_ascend/worker/model_runner_v1.py`).

`use_compress` is a separate upstream attribute, not something `use_sparse` implies:

- `use_sparse = hasattr(hf_text_config, "index_topk") and not hasattr(hf_text_config, "compress_ratios")`
- `use_compress = hasattr(hf_config, "compress_ratios")`

and the backend map keys on all three — `(use_mla, use_sparse, use_compress) == (True, False, True)` selects `dsa_v1`, whose `update_graph_params()` is a documented no-op (`# dsa does not need to update graph params`). That no-op is exactly why the MLA DBO full graph path cannot own such a model.

`uses_sparse_mla` here is only `hasattr(hf_text_config, "index_topk")`, so it cannot express that cell: a configuration carrying `compress_ratios` without `index_topk` would still be classified as the MLA DBO full graph path while the runner sends it to `dsa_v1`, and validation would then demand `FULL_DECODE_ONLY` and reject speculative decoding for a model that never takes that path.

For DeepSeek V4 the term is a no-op — it declares both keys, so `uses_sparse_mla` is already true and the predicate is already false. What actually unblocked DBO is the runtime side (`attention_model_runner.py`, `mla_full_graph_enabled`); this line keeps the validation predicate on the same three-way condition so the two cannot drift apart again.

hsliuustc0106 (commented):
Delta re-review of `c3889fc2` against the previously reviewed `614ae0e8` (one commit, +14/−1, tests only).

Both findings from the last review are fixed-verified:

1. FFN `input_ids` installation is now unit-tested at both sites — the new `input_ids` parametrization covers the legacy install (`compat/npu/forward_context.py:93`, including the `None` passthrough pinned by the `not hasattr` assertion) and the extended MRv2 test covers the MRv2 site (`:70`); the fake's exact-dict equality also proves the ids are not forwarded into the upstream `set_ascend_forward_context` call.
2. The PR body is refreshed — the phantom version-guard claim is gone, the census now matches the 19-file diff exactly, and the Ascend unit-suite run is recorded (1086 passed / 3 skipped / 2 pre-existing failures, cross-referenced).

Also verified: jiangkuaixue123's `uses_mla_compressor` question is correctly resolved — the three-way predicate at `feature_validation.py:77-84` mirrors upstream's graph-params guard, and DSV4 declares both keys so it is unaffected.

No new findings in the delta — it is test-only and CPU-safe. Remaining open items from the prior round, unchanged: accuracy against the native run is still unmeasured on either host (the body states this plainly; that remains the gate before calling DSV4-over-CAMP2P production-ready), and the UNSTABLE merge state is just the dangling empty pending status, not a failure.

ksiyuan:
> Delta re-review of `c3889fc2` against the previously reviewed `614ae0e8` (one commit, +14/−1, tests only).
> 
> Both findings from the last review are fixed-verified:
> 
> 1. FFN `input_ids` installation is now unit-tested at both sites — the new `input_ids` parametrization covers the legacy install (`compat/npu/forward_context.py:93`, including the `None` passthrough pinned by the `not hasattr` assertion) and the extended MRv2 test covers the MRv2 site (`:70`); the fake's exact-dict equality also proves the ids are not forwarded into the upstream `set_ascend_forward_context` call.
> 2. The PR body is refreshed — the phantom version-guard claim is gone, the census now matches the 19-file diff exactly, and the Ascend unit-suite run is recorded (1086 passed / 3 skipped / 2 pre-existing failures, cross-referenced).
> 
> Also verified: jiangkuaixue123's `uses_mla_compressor` question is correctly resolved — the three-way predicate at `feature_validation.py:77-84` mirrors upstream's graph-params guard, and DSV4 declares both keys so it is unaffected.
> 
> No new findings in the delta — it is test-only and CPU-safe. Remaining open items from the prior round, unchanged: accuracy against the native run is still unmeasured on either host (the body states this plainly; that remains the gate before calling DSV4-over-CAMP2P production-ready), and the UNSTABLE merge state is just the dangling empty pending status, not a failure.

Thanks for the delta review. Accuracy is measured now — GSM8K on both hosts against the native colocated vLLM-Ascend baseline, A5 96.86 vs 96.86 and A3 96.66 vs 96.64, recorded in Test Result.

ksiyuan:
@jiangkuaixue123 — ran the DSV4 prefill async E2E coverage from #352 against this PR: **`afd-dsv4-flash-async-cam-dp2tp4-ep8` PASSED**.

- Tree: `main` (`8bb14be`, which already contains #347/#350/#352) plus this PR's six commits — i.e. #352's case unmodified.
- Result: `E2E SCENARIO afd-dsv4-flash-async-cam-dp2tp4-ep8 PASSED`, 10/10 concurrent chat requests, outputs `19 … 28`, same as the reference run recorded in #352; service liveness and owned-process cleanup checks passed too.
- Environment note: this host only has Python 3.11, where the CAM binding cannot be installed as shipped — the wheel is `cp312-cp312` and `comm_operator/pybind/pybind.cpp` carries a runtime version guard plus a `Py_mod_multiple_interpreters` slot. I installed a locally retagged copy with those two spots pointed at 3.11 (two bytes in the module; no `LD_PRELOAD`, no runtime behaviour change). The proper fix is a `cp311` build from UMDK; this was enough to run the gate here. vLLM/vLLM-Ascend are the pinned pair (`0.26.0` @ `568afb3a1`, vLLM-Ascend `80d8c194f`).
- Startup on this host is slow (~20 min; the weights come off shared storage) versus the 443 s reference on `jcz_afd1` — the request phase itself completes in seconds as in the reference.

That closes the async prefill regression check you asked for.

PR #342, opened 2026-09-14
fix(npu): support W4A8 expert weights in CAM MLP

bjf-frz:
CAM MLP needs the packed W4A8 weights, scales, and optional scale biases owned by the routed experts. Handle both static weights and dynamic EPLB weight lists, pass the per-channel quantization flag, and obtain the SwiGLU limit from the MLP wrapper.

Add 16 CPU contract tests covering dynamic EPLB, per-channel quantization, optional biases, and empty token batches. Update the existing quantization test mock for W4A8.

Validation:
- 16 CPU contract tests passed.
- Full configured pre-commit check passed in Linux CI, including Ruff and mypy.
- DCO passed with author and committer bjf-frz <frz123db@gmail.com>.

The extracted change has not been deployed against the current upstream revision.


afd_plugin/model_executor/models/npu/deepseek_v2_attention_gate.py
```diff
@@ -165,10 +165,36 @@ def compute_attention_gate_moe_ffn(
                 ],
                 w2_scale=[experts.get_eplb_parameter("w2_weight_scale")],
             )
+    # ### PATCH START: W4A8 CAM expert weights
```
jiangkuaixue123: LGTM. This is AFD-owned code, so there is no need for the PATCH START/END markers here.

bjf-frz: > LGTM. This is AFD-owned code, so there is no need for the PATCH START/END markers here.

Done, patch markers has been removed.

PR #344, opened 2026-09-16
Upgrade synchronous NPU runtime to vLLM v0.28.0

jiangkuaixue123:
## Purpose

Upgrade the synchronous AFD NPU runtime to vLLM `v0.28.0` (`2cf0a6915ce544dc493a0990f2ea38d81601128a`) and vLLM-Ascend `bd69bad88fc19e1aeeea585416d408df8bda8fef`, building on the GPU upgrade in #325.

The target Ascend config factory rejects AFD-owned settings, and backend normalization can make parent and child configuration hashes disagree during DP startup. This change preserves the plugin namespace through native validation and finalizes the collective backend before serialization. It also aligns the NPU runner and W8A8 compatibility patches with the pinned runtime.

## Scope

Hardware-tested scope: BF16 DeepSeek-V2-Lite, runner V1, synchronous `CAMP2pAFDConnector`, FFN-side gating, TP1, 2A2F and 2A1F eager/graph/DBO.

Async CAM and DSV4 are excluded from this experiment. W8A8 has focused unit coverage, not hardware qualification. NPU V2, additional topologies, PCP, speculative execution and Attention-side gating are not qualified here. CUDA hardware tests were not rerun for this NPU change.

## Implementation

- Rebase metadata/dummy execution and forward-context contracts; retain native model ownership of final sequence-parallel gathering.
- Apply opt-in W8A8 balancing to selected expert IDs while preserving native routing weights and shared-expert handling.
- Filter only the three AFD-owned keys at the strict Ascend factory, preserving caller mappings, cache ownership and validation of unrelated keys.
- Install namespace routing before scheduler construction and before parent/child configuration revalidation.
- Finalize the collective backend using the target raw FlashComm/DSA-CP selectors before serialization. Native DP hash checks remain intact.
- Pad eager DP token counts for CAMP2P fan-in. Previously, 2A1F with uneven Attention batches produced corrupt repeated text because kernels require equal chunks. The fix restores baseline output without changing the kernel protocol.
- No changes to either upstream source tree.

## Validation

Runtime fixes tested through `1b4ca7b`; subsequent commits update documentation and source comments only, with AST equivalence verified.

- **970 passed, 3 skipped, 1 deselected** in the scoped unit suite. Three DSV4 async modules were excluded. Skips: one opt-in two-rank operator test and two missing-extension error-path tests with the extension installed.
- Native TP1 eager request passed on the exact target pair and returned `Paris`.
- **All seven NPU synchronous basic cells passed**, each with GSM8K-7 strict/flexible exact match **2/7 (0.2857)**:
  - `baseline-graph`
  - `afd-eager-2a2f`, `afd-graph-2a2f`, `afd-graph-dbo-2a2f`
  - `afd-eager-2a1f`, `afd-graph-2a1f`, `afd-graph-dbo-2a1f`
- A module-wide selection subsequently hit a pre-existing GPU-only V2 guard before startup. Those six cases are outside the NPU matrix; the seven NPU cells passed without skips.
- Real three-rank CAMP2P roundtrips passed for equal 1/4/128-token blocks; unequal 4/2-token inputs reproduced corruption. After padding, 2A1F eager recovered from 0/7 to 2/7 and passed the complete basic rerun.
- Fresh-process config/scheduler probes and parent/child hash probes passed. Ruff, format, compile/import, exact-signature and diff checks passed.
- **Full 1,319-sample GSM8K was explicitly deferred by the requester for a later run.** The started evaluation was cancelled and cleaned up; no full-accuracy or performance claim is made.
- Owned serving processes, ports and NPU processes were confirmed clear.

Tested stack: Ascend 910C/A3, openEuler 24.03 LTS-SP4, Python 3.12.13, torch 2.10.0+cpu, torch-npu 2.10.0.post4, CANN 9.1.0, Triton 3.5.0 / Triton Ascend 3.2.2, memfabric-hybrid / memcache-hybrid 1.2.0. The FastAPI conflict was explicitly accepted for this experiment; inherited optional-tool conflicts remain. This does not certify a clean `pip check`.

## Documentation

Updated the runtime baseline, connector guides, recipe index, design contracts, patch inventory and contributor templates. Historical v0.19/v0.26 V3.2 and async results retain their original scope.

- [NPU v0.28 recipe and seven-cell/full-accuracy commands](https://github.com/jiangkuaixue123/afd-plugin/blob/535d5bb/recipe/npu/CAMP2pAFDConnector/deepseek_v2_lite/README.md)
- [Upgrade lessons and remaining qualification gaps](https://github.com/jiangkuaixue123/afd-plugin/blob/535d5bb/.agents/skills/upgrade-npu/references/20260916-v0.26.0-to-v0.28.0-lessons.md)

Documentation links, YAML, recipe shell syntax/test selections and final consistency checks passed. The requested synchronous upgrade scope is complete; full accuracy remains a follow-up.


lirx-pd:
**1. [P1] All Attention DP ranks need to make the same DBO decision**

[[attention_model_runner.py:1923–1936](https://github.com/vllm-project/afd-plugin/blob/535d5bbe64fd5f0703297060141491132e6c2f13/afd_plugin/v1/worker/npu/attention_model_runner.py#L1923-L1936)](https://github.com/vllm-project/afd-plugin/blob/535d5bbe64fd5f0703297060141491132e6c2f13/afd_plugin/v1/worker/npu/attention_model_runner.py#L1923-L1936)

This change lets two Attention ranks choose different numbers of stages for the same forward pass.

For example, with DP=2, TP=1 and 64 one-token requests on each rank, an initial-prefill rank uses the 512-token DBO threshold and keeps one stage. A decode rank uses the 32-token threshold and splits into two. `_sync_afd_metadata_across_dp()` synchronizes token counts and graph mode, but still uses each rank’s local `uniform_decode` to make this decision.

I reproduced `[64]` versus `[32, 32]` with real Gloo synchronization and NPU tensors collected through HCCL. Swapping the ranks gives the same mismatch; the base methods produce matching stages.

In 2A1F, FFN follows the control metadata from only one Attention rank, so it cannot match both execution layouts. I verified the stage mismatch, but did not run mismatched CAMP2P transfers to test whether they hang.

Please synchronize the final DBO decision across DP, while keeping `uniform_decode` local for attention semantics. A regression test with one prefill rank and one decode rank would cover this case.

**2. [P2] Apply `log2phy` to the generated expert IDs**

[[force_load_balance.py:253–272](https://github.com/vllm-project/afd-plugin/blob/535d5bbe64fd5f0703297060141491132e6c2f13/afd_plugin/compat/patches/npu/force_load_balance.py#L253-L272)](https://github.com/vllm-project/afd-plugin/blob/535d5bbe64fd5f0703297060141491132e6c2f13/afd_plugin/compat/patches/npu/force_load_balance.py#L253-L272)

The replacement IDs here are logical expert IDs, but dispatch now expects physical IDs.

In the pinned upstream version, `AscendRoutedExperts._select_experts()` applies `log2phy` before calling this method. Replacing `topk_ids` discards that mapping, and nothing downstream applies it again. With redundant experts, logically balanced routing can therefore send all tokens to one physical rank.

I tested this with real two-NPU MC2 dispatch/combine: EP=2, two logical experts, two redundant experts, top-k=1, and 16 tokens per source rank. The destination ranks received **32 and 0 tokens**. Applying `log2phy` to the generated IDs changed that to **16 and 16**. The test exercised transport without expert GEMMs or model inference.

Please map only the newly generated routed IDs, then append the existing shared-expert suffix. Mapping the entire incoming tensor again would map physical IDs twice. If this combination is unsupported, it should be rejected explicitly so force-load-balance benchmarks do not report a misleading workload.

**3. [P2] Keep sparse-offload request IDs when splitting a batch**

[[ubatch_utils.py:267–343](https://github.com/vllm-project/afd-plugin/blob/535d5bbe64fd5f0703297060141491132e6c2f13/afd_plugin/v1/worker/npu/ubatch_utils.py#L267-L343)](https://github.com/vllm-project/afd-plugin/blob/535d5bbe64fd5f0703297060141491132e6c2f13/afd_plugin/v1/worker/npu/ubatch_utils.py#L267-L343)

The runner now fills `req_ids_tensor` and `token_to_req`, but this constructor leaves both out. Each ubatch therefore gets `None` for both fields.

The offload metadata builder copies those values unchanged. Its decode path then raises `Sparse KV offload requires req_ids_tensor/token_to_req metadata`. This affects sparse offload when DBO actually splits the batch.

I verified the field loss using real NPU tensors and the native metadata population method. The unsplit batch retained both fields; both ubatches lost them. Preserving the fields and adjusting the indices made the control pass. Full offload inference remains untested because the available hardware is A2 and this backend requires A3.

Please slice `req_ids_tensor` by request and `token_to_req` by token. The latter also needs valid indices rebased by `request_slice.start`, since they must index the ubatch’s request array. Preserve padding values separately. If sparse offload with DBO is outside this PR’s scope, please reject the combination explicitly.

**4. [P2] The temporary DeepEP backend changes graph configuration during revalidation**

[[config_validation.py:138–143](https://github.com/vllm-project/afd-plugin/blob/535d5bbe64fd5f0703297060141491132e6c2f13/afd_plugin/compat/patches/config_validation.py#L138-L143)](https://github.com/vllm-project/afd-plugin/blob/535d5bbe64fd5f0703297060141491132e6c2f13/afd_plugin/compat/patches/config_validation.py#L138-L143), together with [[ascend_config.py:68–75](https://github.com/vllm-project/afd-plugin/blob/535d5bbe64fd5f0703297060141491132e6c2f13/afd_plugin/compat/patches/npu/ascend_config.py#L68-L75)](https://github.com/vllm-project/afd-plugin/blob/535d5bbe64fd5f0703297060141491132e6c2f13/afd_plugin/compat/patches/npu/ascend_config.py#L68-L75)

Temporarily setting `deepep_low_latency` now does more than bypass validation. The new Ascend `enable_sp()` reads the backend through `use_sequence_parallel_moe`, so this temporary value can turn on sequence-parallel graph filtering.

On a cache hit, `init_ascend_config()` returns before `derive_and_validate()` restores the actual backend. `_setup_compile_backend()` then filters capture sizes using the temporary DeepEP value.

I reproduced this with real `EngineArgs`/`VllmConfig` objects: DP=2, TP=4, EP and DBO enabled, all SP options disabled, and capture sizes `[2]`. Initial construction passed, but a second `__post_init__()` raised:

```text
cudagraph_capture_sizes [2] does not containvalues that are multiples of tp_size 4
```

Both the actual-backend control and the old SP-selector control passed. This was a configuration test; no distributed inference workers were started.

Please keep the temporary backend out of Ascend’s SP and graph configuration logic, and add a test that revalidates the same cached config. Restoring the backend in `finally` does not undo an exception or changes already made to capture sizes.

PR #350, opened 2026-09-18
test(npu): expect the NPU DSV4 registration on Ascend hosts

ksiyuan:
## Purpose

`tests/unit/package/test_package.py::test_deepseek_afd_model_registration_paths_are_lazy_strings`
pins the GPU `DeepseekV4ForCausalLM` registration path:

```
afd_plugin.model_executor.models.deepseek_v4:AFDDeepseekV4ForCausalLM
```

`afd_plugin/__init__.py` registers the NPU class instead whenever `torch_npu`
is importable, so the assertion fails on every Ascend host — the environment
`docs/npu/TESTING.md` prescribes for NPU unit tests — while passing in CI,
which has no `torch_npu` installed.

Found by running that documented command on an Ascend 950 host:
`1086 passed, 3 skipped, 2 failed`. The other failure was already fixed by #347.

## Issue

- Related issue(s): none filed; #347 fixed the sibling failure from the same run.
- Closing keyword, only if fully resolved: none.

## Scope

- In scope: the host-dependent `DeepseekV4ForCausalLM` registration assertion.
- Out of scope: the registration logic in `afd_plugin/__init__.py`, which is
  unchanged, and every other test in the file.

## Implementation Notes

The test now asserts the rule the registration applies —
`importlib.util.find_spec("torch_npu") is not None` selects
`afd_plugin.model_executor.models.npu.deepseek_v4:AFDDeepseekV4ForCausalLM` —
with a comment recording why the assertion has to stay host-aware. No
production code changes.

## Test Plan

Run the file on a host without `torch_npu`, and the same file with `torch_npu`
faked into `sys.modules` (with a real `ModuleSpec`) before `afd_plugin` is
imported. That second run is what makes the registration select the NPU path,
and it is how the failure was reproduced and fixed without Ascend hardware.

## Test Result

- Before, at the base `746427b` with a simulated Ascend host:
  `FAILED tests/unit/package/test_package.py::test_deepseek_afd_model_registration_paths_are_lazy_strings`,
  with the same assertion diff as the real A5 run
  (`- ...models.deepseek_v4:...` / `+ ...models.npu.deepseek_v4:...`).
- After: the simulated-Ascend run passes, and the plain host run passes —
  `10 passed, 15 skipped`, the skips being vLLM-gated tests.
- `ruff check` and `ruff format --check` clean; `mypy==1.11.1` for Python 3.10
  (the tests group as `tools/pre_commit/mypy.py` runs it) reports no issues.
- Both runs were made on Windows with `PYTHONUTF8=1`; no Ascend hardware was
  needed for either side of the comparison.

## Docs Impact

- Files updated: none.
- If none, reason: test-only change, no user-facing behavior.

---

<details>

- [x] Scope is bounded.
- [x] Compatibility with vLLM v0.26.0 is considered. (No runtime or API change.)
- [x] Any compat shim or monkey patch is isolated, idempotent, version-guarded, documented, and tested. (No new shim or patch.)

</details>


jiangkuaixue123:
Please fix pre commit

ksiyuan:
> Please fix pre commit

Fixed in 7765475

PR #351, opened 2026-09-18
bugfix: DBO E2E tests can pass without splitting live requests into ubatches

### LINKED ISSUE #345: [Bug]: DBO E2E tests can pass without splitting live requests into ubatches
## Summary

DBO E2E cases can pass accuracy checks without exercising microbatch splitting on real requests. The observed NPU cases create two ubatches during warmup/capture, but use only stage 0 during evaluated inference. Enabling `--enable-dbo` and passing GSM8K is therefore insufficient evidence of runtime DBO coverage.

Discovered during #341. This issue tracks test coverage and observability, not NPU ModelRunnerV2 support or async CAM.

## Current environment

- AFD: local upgrade branch `update/v0.28.0`, commit `1b4ca7b8e38ff4f201b0915a3e820ee979c1a251` (not yet published).
- vLLM: `v0.28.0`, `2cf0a6915ce544dc493a0990f2ea38d81601128a`.
- vLLM-Ascend: `bd69bad88fc19e1aeeea585416d408df8bda8fef`.
- Ascend 910 A3; Python 3.12.13; CANN 9.1; torch 2.10.0+cpu; torch-npu 2.10.0.post4.
- DeepSeek-V2-Lite, BF16, synchronous `CAMP2pAFDConnector`, NPU V1 runner.
- Attention DP2/TP1; FFN DP2/TP1/EP2 or single FFN rank; ACL Graph + DBO.
- AFD installed editable with plugin CANN operators (`AFD_BUILD_ASCEND_OPS=1 SOC_VERSION=910c`).

## Reproduction

With the target NPU environment, local checkpoint and GSM8K cache prepared:

```bash
export AFD_E2E_BACKEND=npu
export AFD_E2E_DEVICES=0,1,2,3
export AFD_NPU_E2E_MODEL=/path/to/DeepSeek-V2-Lite
export AFD_NPU_E2E_VLLM_BIN=/path/to/venv/bin/vllm
export HF_HOME=/path/to/huggingface-cache
export AFD_GSM8K_LIMIT=7
python -m pytest -s -o addopts=''   'tests/e2e/models/deepseek_v2_lite/test_deepseek_v2_lite.py::test_deepseek_v2_lite[afd-graph-dbo-2a2f]'   'tests/e2e/models/deepseek_v2_lite/test_deepseek_v2_lite.py::test_deepseek_v2_lite[afd-graph-dbo-2a1f]'
```

The scenario builder enables DBO with decode threshold 1 and prefill threshold 8. Inspect `send_dp_metadata decision` records separately for warmup, graph capture and live requests.

## Actual behavior and evidence

Both cases passed GSM8K-7 at strict/flexible match **2/7**. However, the two-stage control payload occurred only in warmup/capture:

```text
key=((0, (4, 4)), (1, (4, 4))) is_graph_capturing=False is_warmup=True is_graph_replaying=False
key=((0, (4, 4)), (1, (4, 4))) is_graph_capturing=True is_warmup=False is_graph_replaying=False
```

Each case had four such log records: two Attention ranks × one warmup and one capture record. No live-request control payload contained stage 1. Live graph replay records instead contained only stage 0, for example:

```text
key=((0, (8, 8)),) is_graph_capturing=False is_warmup=False is_graph_replaying=True
```

`_send_dp_metadata` emits separate stage entries when there are multiple ubatch slices; `AscendUBatchWrapper` selects its split path from those slices. The evidence supports warmup/capture coverage, but does not establish live split execution or communication/compute overlap.

Relevant code:

- `tests/e2e/runner.py`: DBO scenario configuration and accuracy-only success criteria.
- `afd_plugin/v1/worker/npu/attention_model_runner.py`: DP coordination and `_send_dp_metadata`.
- `afd_plugin/v1/worker/npu/ubatch_utils.py`: `check_enable_ubatch` / `is_last_ubatch_empty`.
- `afd_plugin/v1/worker/npu/npu_ubatch_wrapper.py`: runtime split/graph execution.

Lowering the token threshold alone does not guarantee a split: the eligibility checks also reject an empty second ubatch, and DP coordination considers the least-loaded rank. The workload must ensure enough real work on the participating ranks. This is an explanation to test explicitly, not a claim that every DBO workload is broken.

## Expected behavior / proposed improvement

Feature-specific E2E tests should assert that the intended execution path was actually exercised, in addition to checking outputs.

1. Add a deterministic live workload with enough concurrent requests/tokens on all required DP ranks to produce two non-empty ubatches. Cover the intended prefill/decode and graph-replay paths explicitly; a larger sample limit alone is not sufficient.
2. Expose bounded, structured evidence or counters for live split execution, including stage ID, effective token count, execution mode and warmup/capture status. Correlate Attention and FFN execution without dumping prompts or per-token log floods.
3. Require both stage 0 and stage 1 to execute for real requests. Exclude warmup and capture from the assertion. For graph tests, prove replay of the split graph rather than merely successful capture.
4. Fail with an actionable message when DBO was requested but the live split count remains zero. Report thresholds, relevant per-rank token counts and the reason splitting was ineligible.
5. Keep accuracy/output and cleanup checks. Distinguish configuration-enabled, capture-covered, live-split-covered and overlap-measured results. A stage-count assertion alone must not be described as proof of actual overlap; that requires a separate runtime trace/timing check.
6. Audit other feature-named E2E cases for the same false-positive pattern, without assuming GPU or other backends reproduce this NPU observation.

## Acceptance criteria

- [ ] NPU DBO 2A2F and 2A1F tests assert two non-empty ubatches on live requests.
- [ ] Graph cases assert live replay of the split execution path.
- [ ] A negative-control test where splitting is disabled/ineligible fails the coverage gate even when outputs remain correct.
- [ ] Warmup/capture-only stage evidence cannot satisfy the gate.
- [ ] Results include bounded coverage evidence, accuracy and cleanup status; no threshold reductions to conceal missing coverage.
- [ ] Existing DBO-disabled behavior and unrelated connector paths remain unchanged.

## Before submitting

- [x] Searched existing issues for related reports.
- [x] Reproduction environment pins the vLLM 0.28.0 and vLLM-Ascend revisions.

yujuancao07:
## Purpose

The DBO E2E scenarios (`afd-graph-dbo-2a1f`, `afd-graph-dbo-2a2f`) passed
their GSM8K accuracy gate without ever splitting a live request into
ubatches (vllm-project/afd-plugin#345). lm-eval drove one request at a
time, and with attention DP > 1 an idle attention rank always votes
against microbatching, so only engine warmup and graph capture ever
executed the split path. 

## Issue

- Related issue(s): #345 
- Closing keyword, only if fully resolved: Closes #345


hsliuustc0106 (commented):
Consolidated review of `2405942` (static review; pre-commit/DCO green). The root cause is confirmed real and the workload fix is the right direction: `--limit 7` through lm-eval's sequential-by-default client meant only one live request at a time — idle sibling rank, per-rank tokens under the split thresholds — so live inference ran single-stage while only warmup/capture exercised the split path, and 2/7 = 0.286 still cleared the 0.27 gate. `num_concurrent=12` + the 24-sample floor close the no-coverage half of that. One P1 and two P2s:

[P1] The harness still never observes whether splitting happened — tests/e2e/runner.py:757
Splitting is output-preserving, so a splitter that never splits live requests (eligibility always false at ubatch_utils.py:38, only the single-stage `{0: ...}` payload at attention_model_runner.py:1493, or 1-ubatch/partial splits) produces exactly the same PASS verdict as a healthy one — the false-positive hole in the title stays open. The base already emits the evidence: the `AFD NPU Attention send_dp_metadata decision; ... is_warmup=...` warnings, and the runner already streams attention stdout without retaining it. For `enable_dbo` runs, accumulate the attention output during the eval window and require ≥1 live two-stage decision record (not warmup/capture/replay-only), failing with the observed counts — that is also what #345's acceptance criteria ask for. Until that lands, please use "Related: #345" rather than "Closes #345" — as-is the merge would auto-close an issue whose criteria are 0-for-6, and the issue itself warns that a larger sample limit alone is not sufficient.

[P2] Post one runtime run of the new workload — tests/e2e/runner.py:770
The gated DBO workload moves from 7 sequential samples to 24 at 12-concurrent on a small-sample metric with a 0.016 margin; the body has no Test Plan/Test Result and no NPU manual evidence per docs/npu/TESTING.md for a change to the NPU-relevant gate. One run showing `n-samples effective=24`, the accuracy value, and a couple of live two-stage `send_dp_metadata decision` lines would settle both whether the gate stays green and that splits actually happen.

[P2] Update the gate docs — docs/design/module/e2e_testing.md:151
The accuracy-gate table still says "Samples: first 7" for the DBO PR/weekly gates; `max(7, 24)` now silently evaluates 24 regardless of `AFD_GSM8K_LIMIT`. One gate-table note (DBO scenarios floor at `DBO_EVAL_MIN_SAMPLES=24` with 12 concurrent) plus a Docs Impact entry closes it. The body is also missing the template's Scope/Test Plan/Docs Impact sections.

Minor: the unit tests cover only the passing path — the `AFD_GSM8K_LIMIT=all` branch (`sample_limit is not None` guard, the one path that must bypass the floor) has zero coverage.

yujuancao07:
78b59b7adc6b80f1a0d67e76777d28b4

jiangkuaixue123:
The concurrency/sample-count changes should help exercise live DBO. To verify **DBO + CUDA graph replay**, could we replace the `UBatchSlice(` log check with worker-side execution statistics exposed through `--worker-extension-cls`, following the approach in https://github.com/vllm-project/vllm/pull/56922?

That PR adds a worker extension method that reads the encoder graph manager's actual hit statistics, then queries it through `/collective_rpc` after the workload. The extension/RPC mechanism is already available in our pinned vLLM 0.26.0; we do not need that PR to merge first.

For AFD, the suggested flow is:

1. After startup/warmup, call an extension method such as `begin_dbo_graph_coverage()` to start a fresh evaluation window.
2. Run the concurrent GSM8K evaluation.
3. Query `get_dbo_graph_coverage()` and require actual live DBO graph replay evidence, alongside the existing accuracy gate.

The important part is where the evidence is produced: count the actual `cudagraph.replay()` branch in `AFDUBatchWrapper.__call__`, associated with a real request execution and two ubatches containing real tokens. Exclude dummy execution, warmup, and capture. The extension is the readout mechanism; AFD still needs the corresponding counter/probe. `UBatchSlice(` alone also occurs for eager DBO, and upstream `_dummy_run` emits the same slice log. Likewise, graph configuration or captured-graph existence does not prove replay.

Two implementation details to account for:

- vLLM 0.26.0 rejects extension attributes that conflict with existing worker methods. Add uniquely named RPC methods rather than overriding `execute_model` through the extension; have those methods read runner/wrapper instance statistics or explicitly install a test-only probe.
- With internal DP load balancing, `DPLBAsyncMPClient.call_utility_async()` broadcasts to the engines but returns only the first engine's result. Iterating over the HTTP `results` array therefore does not establish coverage of both attention DP ranks. Aggregate evidence across the appropriate DP group or collect per-rank results separately, and validate the expected rank set. If the intended assertion is simultaneous coverage across ranks, retain a shared execution/step identifier as well.

Useful negative tests: capture-only, eager DBO only, and non-DBO graph replay must all fail the coverage assertion; only live DBO graph replay should pass. This can initially establish attention-side coverage; asserting FFN graph replay would require separate FFN evidence.

References checked against this PR head (`565eb33b38c0`):
- [AFD DBO replay branch](https://github.com/vllm-project/afd-plugin/blob/565eb33b38c0570f20f19e7815402ab7b9ab6e0f/afd_plugin/v1/worker/ubatch_wrapper.py#L110)
- [vLLM 0.26.0 worker extension injection](https://github.com/vllm-project/vllm/blob/v0.26.0/vllm/v1/worker/worker_base.py#L261)
- [vLLM 0.26.0 DP utility result handling](https://github.com/vllm-project/vllm/blob/v0.26.0/vllm/v1/engine/core_client.py#L1449)

This is a code-level feasibility suggestion; I have not validated the proposed instrumentation on GPU.

Issue #353, opened 2026-09-18
[RFC]: Restructure csrc/npu into ascend_kernels / pybind / scripts around the CANN npu_op_* build system


francisleely:
### Motivation

`csrc/npu/` currently mixes three unrelated concerns in one flat namespace: the
ACLNN operator sources (`a2e/`, `e2a/`), the PyTorch extension
(`torch_extension/` + `aclnn_torch_adapter/`), and the CANN build system
(`build.sh`, `cmake/`, `build_aclnn.sh`). The flat layout has four concrete
costs:

1. **Duplicated shared headers.** `comm_args.h`, `data_copy.h`, and
   `moe_distribute_base.h` are vendored twice, once under `a2e/op_kernel/` and
   once under `e2a/op_kernel/`. The two copies are byte-identical today, so any
   fix to the MoE distribute base must be applied twice with nothing keeping
   them in sync.
2. **Operator selection is hard-coded shell control flow.** `build_aclnn.sh`
   passes `-n "a2e;e2a"` and maps SOC to operator set through `case`
   statements. Adding an operator or a SOC generation means editing shell, and
   nothing validates that a requested operator is buildable for the requested
   SOC.
3. **A vendored generic CANN build wrapper with no owner.** `build.sh` is a
   ~200-line "build open project" script carrying ccache/bisheng handling and
   `-DCUSTOM_ASCEND_CANN_PACKAGE_PATH` plumbing that AFD never varies. It is
   CANN-licensed boilerplate maintained in-tree, so every CANN bump is a
   potential merge.
4. **The A5 tiling gate is fragile.** `AFD_TILING_HAS_COMM_ENGINE` gates
   `Mc2CcTilingConfig::SetCommEngine(3)` for Ascend 950 in `a2e.cpp` and
   `e2a.cpp`. In the current tree it is wired into the code-generation compile
   options in a way that does not take effect (see Risks).

The operator half of this has an established upstream shape: the CANN
`npu_op_*` run-package build system, the same structure used by the
`umdk`/`cam` `comm_operator` `ascend_kernels` project. Aligning with it lets us
delete the vendored wrapper instead of maintaining it.

### Proposed change

Split the directory along concern boundaries, and drive operator selection from
data instead of shell logic. Diff: 56 files, +1576/−3170.

```text
csrc/npu/
  ascend_kernels/          # ACLNN operator run package (npu_op_* build system)
    CMakeLists.txt         #   npu_op_package(... TYPE RUN) driver
[...]

PR #354, opened 2026-09-20
[NPU] Restructure csrc/npu around the CANN npu_op_* build system

francisleely:
## Purpose

`csrc/npu/` mixed three unrelated concerns in one flat namespace: the ACLNN
operator sources (`a2e/`, `e2a/`), the PyTorch extension (`torch_extension/`
+ `aclnn_torch_adapter/`), and the CANN build system (`build.sh`, `cmake/`,
`build_aclnn.sh`). This restructures the directory along those boundaries and
replaces shell-encoded operator selection with a registry, so that adding an
operator or a SOC generation no longer means editing `case` branches.

Design discussion: <https://github.com/vllm-project/afd-plugin/issues/353>

## Issue

- Related issue(s): <https://github.com/vllm-project/afd-plugin/issues/353>

## Scope

- In scope: `csrc/npu/` layout and build system, `setup.py` extension source
  path, the affected unit test, and `csrc/npu/README.md`.
- Out of scope: a2e/e2a kernel behavior, the pybind surface, the connector
  data path, and `afd_plugin/compat/`.

## Implementation Notes

New layout:

```text
csrc/npu/
  ascend_kernels/     ACLNN operator run package (npu_op_* build system)
  pybind/             PyTorch extension (_C_ascend)
  scripts/            compile_ascend_proj.sh, select_ops.py, set_conf.py
  build_aclnn.sh      thin entry point
```

Behavioral changes beyond the moves:

- Operator selection becomes registry-driven. `operator_registry.json` maps a
  SOC generation to the operator directories it supports plus per-op
  metadata; `select_ops.py` validates the selection (unknown SOC, unknown
  operator, and empty entries are rejected) and applies the SHMEM filter.
- Shared kernel headers (`comm_args.h`, `data_copy.h`,
  `moe_distribute_base.h`) deduplicate into `ascend_kernels/utils/op_kernel/`.
  The a2e and e2a copies were byte-identical, so every fix previously had to be
  applied twice with nothing keeping them in sync.
- Op host sources consolidate: `a2e_def`/`a2e_proto`/`a2e_tiling` fold into
  `a2e/op_host/a2e.cpp`, `e2a_tiling` folds into `e2a/op_host/e2a.cpp`, and
  `op_api/` moves under each operator.
- Run-package naming becomes deterministic: `AFD_${SOC_VERSION}.run` instead
  of globbing `output/CANN-custom_ops*.run` and requiring exactly one match.
- `build.sh` is deleted. `ASCEND_CANN_PACKAGE_PATH` resolution moves into the
  new driver (`ASCEND_HOME_PATH`, then `ASCEND_TOOLKIT_HOME`, then the default
  toolkit path); parallelism becomes `AFD_BUILD_JOBS` (default 8) instead of
  `2 x nproc`.
- `setup.py` builds the extension from `csrc/npu/pybind`. The target name
  `afd_plugin._C_ascend` and the installed artifact layout are unchanged, as is
  the `vendors/afd-plugin` rpath contract.
- `AFD_TILING_HAS_COMM_ENGINE` now reaches the code-generation compile options as
  the file's own comment intends. The gate appends the define to
  `CODE_GEN_COMPILE_OPTS` and then initialized that variable with a plain
  `set()`, discarding the append before `npu_op_code_gen()` consumed it.
  Initializing first fixes the ordering. This is an ordering bug introduced by
  this restructure, not a pre-existing one: on `main` the define is delivered
  via `target_compile_definitions(optiling ...)` and the guarded tiling source
  is compiled into that same target.

No functional change is intended for the a2e/e2a kernels or the pybind surface,
apart from the `TWO_DIMS` magic-number extraction in `e2a/op_host/e2a.cpp` and
the A5 gate ordering fix above. Any other semantic diff in the moved sources
should be treated as a bug.

### Pre-existing lint violations fixed along the way

CI runs pre-commit only on the files a PR touches, resolved by the
`from-ref..to-ref` diff. Renaming a file puts its new path in that set, so four
files that were never linted before now have to pass. Each violation below
already existed on `main`; they are fixed here because CI would otherwise fail:

| File | Pre-existing violation |
| --- | --- |
| `pybind/pytorch_extension/NPUStorageImpl.h` | `storagImpl` typo in a comment (typos) |
| `ascend_kernels/utils/op_kernel/data_copy.h` | whole file CRLF, no trailing newline |
| `ascend_kernels/utils/op_kernel/moe_distribute_base.h` | no trailing newline |
| `.gitignore` | trailing whitespace on two pre-existing lines |

`data_copy.h` is the one worth a second look: normalizing it to LF changes every
line, so git reports the move as `R097` rather than `R100` and the diff shows a
full-file rewrite. Ignoring line endings, the content is byte-identical to the
version on `main`.

## Test Plan

CPU-only and static checks, the full pre-commit hook set on the changed files,
plus an Ascend operator build.

## Test Result

`pytest tests/unit/package/test_ascend_build_files.py` — 26 passed, covering
the new paths, the absence of duplicated shared headers, registry-to-directory
consistency for both SOC generations, and `select_ops.py` resolution (full set,
explicit subset, and rejection of unknown SOC / unknown operator / empty
entry).

`pytest tests/unit/package/` — one pre-existing failure,
`test_package.py::test_deepseek_afd_model_registration_paths_are_lazy_strings`,
which expects the non-NPU `deepseek_v4` path while this environment reports the
`npu` variant. Confirmed against a pristine `HEAD` snapshot to be unrelated to
this change.

pre-commit, run on the changed-file set with the repository configuration:
`check-merge-conflict`, `debug-statements`, `end-of-file-fixer`,
`mixed-line-ending`, `trailing-whitespace`, `ruff-check`, `ruff-format`, and
`typos` all pass. `check-yaml` skips (no YAML touched). markdownlint on
`csrc/npu/README.md` passes (markdownlint-cli2 v0.23.2 with the repository
`.markdownlint.yaml`), and `shellcheck -s bash` is clean on both shell scripts.
`actionlint` could not run locally because its Go environment needs to fetch
modules from `proxy.golang.org`, which is unreachable here; it only applies to
`.github/workflows/**` and no workflow YAML is touched, so it should skip on CI.

Ascend operator build for both registered SOC generations installs correctly
under `afd_plugin/_cann_ops_custom/vendors/afd-plugin/`.

The A5 gate was checked with `cmake -P` before and after the fix for both
`ascend950` and `ascend910_93`.

Not covered here: NPU end-to-end execution and accuracy on real hardware. The
environment used for this validation has no NPU cards, so it can compile the
operators and run static/completeness checks but cannot run the full test
suite or an E2E gate. I will run the DeepSeek-V2-Lite NPU E2E gate separately;
please tell me if you want that evidence attached before review.

## Docs Impact

- Files updated: `csrc/npu/README.md` (new layout, build flow, and the new
  `AFD_BUILD_JOBS` / `CMAKE_BUILD_TYPE` knobs) and
  `.agents/skills/review-pr/references/modules/execution-platforms.md` (module
  ownership description).

---

<details>

- [x] Scope is bounded.
- [x] Compatibility with vLLM v0.26.0 is considered. (No vLLM interface is touched; this is build-system and packaging work.)
- [x] Plugin-owned classes or explicit dotted class paths are preferred over monkey patches. (No new patch or shim; `afd_plugin._C_ascend` and `afd_ascend` are unchanged.)
- [x] Any compat shim or monkey patch is isolated, idempotent, version-guarded, documented, and tested. (n/a — none added or changed.)
- [x] Imports remain CPU-safe; CUDA-heavy work is delayed or GPU-gated. (No module-level import changes.)

</details>


jiangkuaixue123:
Fix pre-commit

